PR #32, opened 2015-08-06
Refactor to a more general framework.

jetuk:
The motivation here was to move away from water sector specific language and nuances in the core module(s). With that in mind the naming of the nodes in there is very general. There are several things that have changed with this pull request, 

- Creation of pywr.domains package to contain sector specific subclasses. A river domain has been partially implemented using similar class names from before. More are needed to finish this.
- Creation of a Domain class that is given to Nodes to identify them with a specific domain. This is an important concept because routes can not traverse different Domains. However there is a new type of Input node, InputFromOtherDomain, that can be connected to an Output from a different domain. Flow can then occur between these two nodes. A conversion_factor is used to account for any change in units or losses that occur from moving from one domain to another.
- Use of Nodes that contain other Nodes - see Storage and PiecewiseLink as examples. This is a useful addition, but should perhaps be implemented using another class, e.g. CompositeNode(Node). Connect and using these nodes is transparent from the user, but the connecting is a bit hackish due to subclass specific checking required in Node.commit.
- PiecewiseLink is a general node that contains multiple nodes. It is a useful way to have a piecewise benefit objective at a specific nodes. An example is a river gauge - which should probably be implemented as a subclass of PiecewiseLink.
- Added schematic.py (based on your code). It now has colored circles to show the different Domains, and also does not show subnodes on composite nodes.
- The GLPK solver was hacked to make all of this work.

Things to finish off,

- Tests - most are broken.
- GLPK solver is broken with timedelta timesteps. Discussion about how to handle this needs to occur.
- Confusion between the Domain class and the pywr.domains package?
- Demand, Supply and Blender classes need to be deleted or moved from core.py
- Nothing has been done with licences; they may also be broken.
- More subclasses need implementing in rivers.py. Examples that come to mind include RiverGauge(PiecewiseLink) and ReservoirWithControlCurve(Reservoir).

I am sure I have missed some stuff of these lists.

James


snorfalorpagus:
This will take me a little while to review, but in principle I think it's a nice extension to the idea.

jetuk:
I shouldn't have merged this with my master branch, should I?

snorfalorpagus:
I hadn't noticed. If you make any changes to your master branch now they will automatically be added to this pull request. Perhaps it would be better for you to create a new branch in your repo, and submit a new pull request based on that. Alternatively, create a new branch in your repo and continue your work in that.

jetuk:
OK I've reverted the commits on my master and will resubmit PR from my branch.

PR #33, opened 2015-08-07
Generalisation (on the correct branch)

jetuk:
The motivation here was to move away from water sector specific language and nuances in the core module(s). With that in mind the naming of the nodes in there is very general. There are several things that have changed with this pull request, 

- Creation of pywr.domains package to contain sector specific subclasses. A river domain has been partially implemented using similar class names from before. More are needed to finish this.
- Creation of a Domain class that is given to Nodes to identify them with a specific domain. This is an important concept because routes can not traverse different Domains. However there is a new type of Input node, InputFromOtherDomain, that can be connected to an Output from a different domain. Flow can then occur between these two nodes. A conversion_factor is used to account for any change in units or losses that occur from moving from one domain to another.
- Use of Nodes that contain other Nodes - see Storage and PiecewiseLink as examples. This is a useful addition, but should perhaps be implemented using another class, e.g. CompositeNode(Node). Connect and using these nodes is transparent from the user, but the connecting is a bit hackish due to subclass specific checking required in Node.commit.
- PiecewiseLink is a general node that contains multiple nodes. It is a useful way to have a piecewise benefit objective at a specific nodes. An example is a river gauge - which should probably be implemented as a subclass of PiecewiseLink.
- Added schematic.py (based on your code). It now has colored circles to show the different Domains, and also does not show subnodes on composite nodes.
- The GLPK solver was hacked to make all of this work.

Things to finish off,

- Tests - most are broken.
- ~~GLPK solver is broken with timedelta timesteps~~. Discussion about how to handle this needs to occur.
- Confusion between the Domain class and the pywr.domains package?
- Demand, Supply and Blender classes need to be deleted or moved from core.py
- Nothing has been done with licences; they may also be broken.
- More subclasses need implementing in rivers.py. Examples that come to mind include RiverGauge(PiecewiseLink) and ReservoirWithControlCurve(Reservoir).

I am sure I have missed some stuff of these lists.

James

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -845,7 +845,7 @@ class Output(Node):
         self.properties['benefit'] = self.pop_kwarg_parameter(kwargs, 'benefit', 1000.0)
 
 
-class Supply(Node):
+class Supply(Input):
     """A supply in the network
 
     The base supply node should be sufficient to represent simply supplies
@@ -862,7 +862,7 @@ class Supply(Node):
             A simple maximum flow constraint for the supply. For more complex
             constraints a License instance should be used.
         """
-        Node.__init__(self, *args, **kwargs)
+        Input.__init__(self, *args, **kwargs)
         self.color = '#F26C4F' # light red
 
         self.properties['max_flow'] = self.pop_kwarg_parameter(kwargs, 'max_flow', 0.0)
@@ -893,7 +893,7 @@ class Supply(Node):
             node.licenses = LicenseCollection.from_xml(licensecollection_xml)
         return node
 
-class Demand(Node):
+class Demand(Output):
     """A demand in the network"""
     def __init__(self, *args, **kwargs):
         """Initialise a new Demand node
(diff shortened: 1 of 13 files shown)


jetuk:
Actually one thing I have remembered,

- Ambiguity between 'cost' and 'benefit' properties. This needs some clarification.

jetuk:
Fixed timedelta issue with current HEAD.

snorfalorpagus:
I think that `Supply` should be moved into `rivers.py`, as the licences are fairly water-resources specific. What do you think about the `Blender` class though? It certainly has application in other domains, but I'm not sure if it should live in `core.py`.

jetuk:
I agree. However the reason they are left is because I haven't used them or
converted them to the new classes. Supply, for example, should be a
subclass of Input. I haven't looked at the licences.

jetuk:
I've just updated the RiverGauge class to show how it can use the PiecewiseLink. There is test as well. 

Also noted that I have actually commented out the licence and blender stuff in the solver.

snorfalorpagus:
Closing this in favour of #34

PR #34, opened 2015-08-26
Cython

jetuk:
Here is the initial Cython refactor I discussed. Lots of changes in here, but the main addition is `_core.pyx` to contain all of the Cython extension types. Much of the rest of the package remains in the Python space. Mostly notably the GLPK solver is not cythonized and therefore the speed improvement still has some way to go.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -1,0 +1,266 @@
+
+import numpy as np
+cimport numpy as np
+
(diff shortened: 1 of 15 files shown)


snorfalorpagus:
I'll have a go at using GLPK directly from Cython. If you could have a look at fixing some of the tests that would be useful.

PR #36, opened 2015-08-30
Cython

jetuk:
More features to include in master,
- Merged your Cython GLPK solver
- Fixed tests/code so both solvers pass in Python 2 and 3
 - Exception of a group licence for the Cython version and all the XML serialization tests.
 - Moved licence code out of solver into the `Input.get_max_flow` method
- py.test can take `--solver` command link argument to select solver
- New PiecewiseLink behaviour as we discussed last week. 
 - This works but needs refinement in terms which classes are used in the "underneath"
- New test for min_volume on Storage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
     - conda update --yes conda
 
     # dependencies
-    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy matplotlib pytest dateutil pandas statsmodels networkx xlrd cython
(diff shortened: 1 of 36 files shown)


jetuk:
Yey got the tests to pass on travis!

snorfalorpagus:
Nice work.

* I think the `--solver` option should be handled as a fixture so that we only need `request.config.getoption("--solver")` once.
* I'm not sure if licences are generic enough to be part of `BaseInput`?

PR #38, opened 2015-09-02
Max volume fix

jetuk:
Hmm not sure why this has so many commits? Is it because I rewrote  history ?

Anyway I fixed a bug with maximum volume.


snorfalorpagus:
Yes, the extra commits are due to you rewriting the history to correct your name/email.

I don't think I can fix this via GitHub. I'll have to pull your branch, and then force it onto my copy of master.

snorfalorpagus:
I've done as above. Hopefully everything should be in order now.

Issue #10, opened 2015-01-28
Reservoir control curves


snorfalorpagus:
Implement control curves for reservoirs.

 * Resource state
 * Above/Below curve take

snorfalorpagus:
Closed by #62

PR #40, opened 2015-09-02
Reservoir control curves

jetuk:
Initial attempt at addressing #10 

The control curve is implemented on the `Reservoir` by overloading the `get_cost` method. This alters the cost when the reservoir fill percentage is equal to or above a user defined threshold (`above_curve_cost`). When below the curve the usual `cost` parameter is used.

There is a test of this as well.

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -556,7 +556,7 @@ class ParameterDailyProfile(Parameter):
         self._values = values
 
     def value(self, index=None):
-        return self._values[index.dayofyear-1]
+        return self._values[index.datetime.dayofyear-1]
 
     @classmethod
     def from_xml(cls, xml):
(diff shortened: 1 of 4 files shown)


snorfalorpagus:
Why did you decide to do the control curves with the cost on the reservoir, rather than the method we worked out the other day with a variable for the deficit against a target level?

jetuk:
I guess because it was easy, that variable doesn't exist yet and I forgot
about that discussion. :(

jetuk:
Can you remind me why the additional variable is the preferred approach, and how it results in different behaviour from this implementation?

snorfalorpagus:
The issue with changing the cost based on whether the volume is above or below the curve is hysteresis. One day it's above the curve, so it uses the reservoir, then as a result the next day it's not, so it doesn't. On off on off on off. This is how Aquator works (although it uses the "resource state" metric which is maximised, rather than a cost to be minimised).

The alternate approach is to assign a cost to the deficit between the actual level and a target level. This is more complex, as you need to balance three costs: the cost of deviating from the target level, the cost of supply from the reservoir (compared to other sources), and the cost of refilling the reservoir.

I've created a gist with the prototype we wrote here: https://gist.github.com/snorfalorpagus/a600fac095928a8b8fe9

jetuk:
Doesn't this example suffer the same problem? I haven't tested. 

The `col_volume_target_deficit` can't be negative. Does that mean the reservoir can't go above its target? Or it can't cross the target storage in single timestep?

jetuk:
Not quite sure what happened with rebasing this against master. Had you incorporated these changes already? 

Anyway it's updated now. We still need to figure out how to do this.

snorfalorpagus:
Superseded by #62

PR #47, opened 2015-09-04
Modified Storage to handle multiple inputs/outputs

snorfalorpagus:
This PR modifies the `Storage` node in preparation so that it can handle multiple inputs and outputs, in preparation for reimplementation of old RiverSplit and Blender nodes.

This has only been implemented for the cyglpk solver for now (why are we maintaining two solvers now anyway?)


snorfalorpagus:
I think the Travis CI build is failing because it's only testing the PyGLPK solver, whereas I've only updated the CyGLPK solver. Why are we maintaining both solvers anyway?

jetuk:
Debugging the Python one is slightly easier? But that's about the only reason.

jetuk:
I would prefer if `StorageInput` and `StorageOutput` remain non-connectable only inheriting from `Base{Input|Output}`. This is because they are a bottleneck and ripe for Cythoning, and then we can't do inheritance from Python. `Connectable` has no reason to be in Cython.

API would look something like this,
```
storage(model, 'reservoir', num_outputs=1, num_inputs=2)
supply.connect(storage)  # defaults to_slot=0
supply.connect(storage, to_slot=1)
storage.connect(demand1, from_slot=0)
storage.connect(demand2, from_slot=1)
```

jetuk:
This looks better. How difficult is the PyGLPK solver fix? Can it be back ported from the CyGLPK version?

Can I commit to this PR?

snorfalorpagus:
It's not possible for you to commit directly to a PR opened by someone else, as you need to have write access to the branch the PR is for. If I moved the pywr repository into an organisation, give us both write access, then did development (i.e. new branches) in the pywr repo then it would be possible.

The latest commit should make this work in the python solver.

Issue #50, opened 2015-09-04
Write some documentation


jetuk:
Using Sphinx?

Should include,
- Installation
- Basic tutorial
- Advanced tutorial
- Guidance on writing a new solver
- API Reference

snorfalorpagus:
I did actually start this, but didn't get very far.

jetuk:
Also needs to include the linear programme formulation

jetuk:
I can't build the sphinx documentation due to missing 'alabaster' theme despite having installed it. Very confused.

snorfalorpagus:
JH suggested we do this as "a short communication".

jetuk:
Shall we host on www.readthedocs.org ?
They have post commit hooks to rebuild the docs.

snorfalorpagus:
I don't see why not. I've not used it before, but persumably we just need to write with sphinx / rst as suggested, then setup the webhooks.

snorfalorpagus:
The current build on readthedocs seems to be failing because of pytables dependency on libhdf5.

jetuk:
Yes, I saw that and decided something else was more interesting to fix. No
doubt this is the start of a long line of issues to get pywr installed on
RTD.

snorfalorpagus:
Documentation is hosted on GitHub Pages: https://pywr.github.io/pywr-docs/

This requires the documentation to be built manually, but that avoids the issues with some of the dependencies not being available on RTD.

To update the documentation, run `make html` in the `docs` branch/folder, then push the result to the `pywr-docs` repo `gh-pages` branch.

snorfalorpagus:
The only major bit missing from this now is the section on the problem formulation. Do you think there are issues with this and trying to publish at the same time?

snorfalorpagus:
I've added some documentation for the new JSON document format: https://pywr.github.io/pywr-docs/json.html

snorfalorpagus:
I've merged #51 as 3618c75d9e6736347b7af50a6674d7b8db373ec6.

Subsequent changes can have their own issues/PRs.

PR #51, opened 2015-09-04
Documentation

snorfalorpagus:
It's about time we had some documentation! #50

I suggest we build this up in a separate branch, then at some point squash it and merge into master.


snorfalorpagus:
Merged as 3618c75d9e6736347b7af50a6674d7b8db373ec6.

PR #56, opened 2015-09-11
Removed overloaded after on Storage. 

jetuk:
Storage should record the current volume, not the change (which is default behaviour for _core.Storage).


jetuk:
Sorry should have said. This requires we agree what/how `Storage` records thing when being used as splitter. Hence the current test failure.

jetuk:
This now merged in the Scenarios branch. Not sure if you want to do them separately.

PR #57, opened 2015-09-11
Scenarios

jetuk:
Lots of changes here. We should discuss and agree the way forward.

Generally this is what has happened,
- `_core.Node` and `_core.Storage` now contain an array of their current state (i.e. flow, volume)
- `commit` methods are given a scenario id to update the value with.
 -  'commit_all' can give all scenario results in one go.
- There is now an explicit `Model.setup()` phase that is distinct from `Model.reset()`. There are several reason for this,
  - Setup is needed to create the appropriate space for recorders and node state arrays.
  - It is also used by refactored `Solver` to setup the LP first.
  - Later on the same model can be reset and rerun, without repeating setup. This will be useful when running in a loop with the same problem size.
- A `ParameterConstantScenario` object has been created as an example of how to make use of scenarios.
- The existing solvers work with this, but just solve the same LP in a loop.

Things that need addressing,
 - Licences - these are not scenario aware at present.
 - Is the `Recorder.commit` signature correct? Does it, in general, need any more information?

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -6,7 +6,8 @@ from __future__ import print_function
 from pywr import _core
 
 # Cython objects availble in the core namespace
-from pywr._core import ParameterConstantScenario, ParameterArrayIndexed
+from pywr._core import ParameterConstantScenario, ParameterArrayIndexed, \
+    ParameterArrayIndexedScenarioMonthlyFactors
 
 import os
 from IPython.core.magic_arguments import kwds
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -104,7 +104,6 @@ class Model(object):
         self.failure = set()
         self.dirty = True
 
-
         if solver is not None:
             # use specific solver
             try:
@@ -229,6 +228,8 @@ class Model(object):
 
     def step(self):
         """Step the model forward by one day"""
+        if self.dirty:
+            self.setup()
         self.timestep = next(self.timestepper)
         return self._step()
 
@@ -257,6 +258,8 @@ class Model(object):
 
         Returns the number of last Timestep that was run.
         """
+        if self.dirty:
+            self.setup()
         for timestep in self.timestepper:
             self.timestep = timestep
             ret = self._step()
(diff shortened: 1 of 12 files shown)


pywr/core.py
```diff
@@ -118,9 +132,22 @@ def __init__(self, solver=None, parameters=None):
...
+        self.scenarios = ScenarioCollection()
 
         self.reset()
 
+    @property
+    def number_of_timesteps(self):
+        return len(self.timestepper)
+
+    @property
+    def number_of_scenarios(self):
```
snorfalorpagus: This doesn't feel very Pythonic. I wonder if we could do something more like `len(model.scenarios)` and `len(model.scenarios.combinations)`? Not sure.

tests/test_scenarios.py
```diff
@@ -0,0 +1,90 @@
+# -*- coding: utf-8 -*-
+"""
+A series of tests of the Scenario objects and associated infrastructure
+
+
+"""
+import pywr.core
+from helpers import assert_model
+# To gete simple_linear_model fixture
```
snorfalorpagus: typo

pywr/core.py
jetuk: `len(model.scenarios.combinations)` has some effort involved in computing the combinations, whereas the length is a simple product. Therefore that would require `model.scenarios.combinations` being an object itself?

snorfalorpagus: `model.scenaios.combinations` would be an iterable object with `__len__` implemented, so that you could do:

```
length = len(model.scenarios.combinations)
for combination in model.scenarios.combinations:
    pass
```

It would require a small refactor of `cartesian` so that it was iterable, but I don't think that should be too hard.

jetuk: `cartesian` is recursive, is that a problem?

snorfalorpagus: Why can't we use `itertools.product` for this anyway?

https://docs.python.org/2/library/itertools.html#itertools.product

```
>>> [x for x in itertools.product([1,2],[3,4],[5,6])]
[(1, 3, 5), (1, 3, 6), (1, 4, 5), (1, 4, 6), (2, 3, 5), (2, 3, 6), (2, 4, 5), (2, 4, 6)]
```

jetuk: I have no idea! Good point :)

PR #60, opened 2015-09-29
Added try-except clause around pyd imports. 

jetuk:
This means the imports do not fail when the solver is not built.

Tested on Windows.


jetuk:
Fix these things first,
- Do not register cython based solver classes when the import fails.
- Allow  `import pywr.solvers` before `import pywr`

PR #63, opened 2015-09-30
Fixed ParameterDailyProfile.value method

jetuk:


snorfalorpagus:
Presumably this slipped by because none of the tests use this class. Need to get some sort of coverage analysis working...

PR #67, opened 2015-10-09
More tests

jetuk:
Tests covering some of the issues. All marked as xfail at the moment as we need to decide how to deal with them.


jetuk:
Hmm I can't recreate this failure.

Issue #74, opened 2015-10-30
Configure which solvers are built with command line arguments


snorfalorpagus:
I think this is worth doing to improve the reporting from Travis. Currently, if the lpsolve solver build fails but glpk is OK, the build will succeed and we wont get any failures until it tries to run the tests. I'm thinking something like this:

```
python setup.py install --with-glpk --without-lpsolve
```

snorfalorpagus:
Closed by #75.

PR #75, opened 2015-11-03
Added support to setup.py for --with-glpk and --with-lpsolve

snorfalorpagus:
Beginning to address #74.

Need to configure Travis to use `--with-glpk` and `--with-lpsolve`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
     - sudo ln -s /usr/lib/lp_solve/liblpsolve55.so /usr/lib/liblpsolve55.so
 
 install:
-    - python setup.py install
+    - python setup.py install --with-${PYWR_SOLVER}
 
 script:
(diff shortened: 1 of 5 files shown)


snorfalorpagus:
By chance this PR came in when an update to miniconda broke the Travis install. I've fixed that and already merged to master (6db9aae60e9408e556822e36c7beeac8f1a82b2b).

These changes highlighted a few tests that were relying on the default solver, i.e. even when `--solver=lpsolve` was passed some tests were running against GLPK. I think we need a better mechanism for selecting the default solver if none is provided, rather than just assuming GLPK.

The method for accepting `--with-glpk` and `--with-lpsolve` isn't very nice, but it will do for now I think.

jetuk:
Had a quick look through the changes to `setup.py` and seems good to me - sensible default.

PR #80, opened 2015-12-07
Show build errors when no solvers are built

jetuk:
Some changes to `setup.py`:
- moved to setuptools.setup rather than distutils, and
- display traceback for the solver builds if none of them are successful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ before_install:
     - conda update --yes conda
 
     # dependencies
-    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy matplotlib pytest dateutil pandas statsmodels networkx xlrd cython six ipython pytables
+    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy matplotlib dateutil pandas statsmodels networkx xlrd cython six ipython pytables
+    # pytest from conda is too old, see https://github.com/pytest-dev/pytest/issues/995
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
It looks like the switch from `distutils` to `setuptools` is the cause of the build failures on Travis for 2.7, although that's not obvious from the tracebacks. What was the reason for the switch?

jetuk:
To make it work with MSVC on Windows. See,

https://github.com/cython/cython/wiki/CythonExtensionsOnWindows

snorfalorpagus:
I believe I've fixed this issue in my branch: https://github.com/snorfalorpagus/pywr/compare/setup_build_errors

@jetuk Please can you confirm this can be built successfully using MSVC on Windows?

jetuk:
I get a build failure for not having lpsolve setup on my Windows machine. So now it doesn't complete unless you have both GLPK and lpsolve available. Is that the intention?

snorfalorpagus:
There seems to be an issue with calling `setup()` twice when using setuptools on Python 2.x. The second call overwrites the efforts of the first call so that you end up with core module but no solvers. Now it does everything in one call, but that means it either works or it doesn't. By default it will attempt to build all solvers. If you only want GLPK, use `--with-glpk`.

jetuk:
With `--with-glpk` it works and completes the build. The fact it showed me the errors when trying to build with lpsolve was one of the reasons for this issue anyway. So I think that is solved now. I have pushed your changes to my branch.

PR #81, opened 2015-12-29
Initial attempt at adding solver statistics to the GLPK solver.

jetuk:
This method uses a dictionary on the solvers to accumulate timings on different parts of the solve. Times are summed with previous totals to accumulate the total time if the solver is used multiple times (i.e. through a series of timesteps).

Opinions please.

### DIFF
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -56,6 +56,10 @@ else:
         def solve(self, model):
             return self._cy_solver.solve(model)
 
+        @property
+        def stats(self):
+            return self._cy_solver.stats
+
 
 try:
     from .cython_lpsolve import CythonLPSolveSolver as cy_CythonLPSolveSolver
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -2,7 +2,7 @@ from libc.stdlib cimport malloc, free
 
 from pywr._core import BaseInput, BaseOutput, BaseLink
 from pywr._core cimport *
-
+import time
 include "glpk.pxi"
 
 cdef class CythonGLPKSolver:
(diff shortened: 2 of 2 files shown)


snorfalorpagus:
Seems useful. A dictionary makes sense, as other solvers might have different stats. I haven't benchmarked, but I wonder if calling functions from time.h directly and storing the results in a Cython class rather than a dictionary would make any difference? Every millisecond counts inside the tight run loop.

snorfalorpagus:
http://stackoverflow.com/a/361492/1300519

Apparently harder than I thought!

jetuk:
Also useful:

http://stackoverflow.com/questions/85451/python-time-clock-vs-time-time-accuracy

It didn't have any appreciable impact on my model. However it has some significant slow Python code. However reading `time.time()` docs it might not even be accurate enough.

snorfalorpagus:
The first answer suggests we would have to use `time.process_time` on 3.3+ and `time.clock` below that. However, both of these functions do not include GPU time so the timing would be useful for your OpenCL based solver. I'm inclined to suggest that we add in benchmarking code as needed for testing, rather than include it in the main distribution.

jetuk:
I intended these stats to be general information after a run of a model rather than super accurate for benchmarking. It would probably be useful to report both CPU and wall time.

The OpenCL solver would have to do its own thing and report the time spent in CPU, GPU, memory transfers etc. 

However we should setup some continuous benchmarking (asv?) as a separate thing.

snorfalorpagus:
OK. I guess we do something like this:

```
try:
    timer = time.process_time
except NameError:
    timer = time.clock
```

Then replace calls to `time.time()` with `timer()`.

I agree that we should have some standard benchmarks. asv doesn't work on Windows, but that doesn't bother me.

PR #83, opened 2016-01-17
Initialise AbstractNode._flow with 0D array in __init__. Fixes #82.

### LINKED ISSUE #82: Model.reset shouldn't raise exception if called before step/run
If a model hasn't been run yet calling `reset` on it raises the following exception.

```
  File "<stdin>", line 1, in <module>
  File "d:\pywr\pywr\core.py", line 322, in reset
    node.reset()
  File "pywr\_core.pyx", line 354, in pywr._core.BaseInput.reset (pywr/_core.c:9996)
  File "pywr\_core.pyx", line 357, in pywr._core.BaseInput.reset (pywr/_core.c:9946)
  File "pywr\_core.pyx", line 174, in pywr._core.AbstractNode.reset (pywr/_core.c:6260)
AttributeError: Memoryview is not initialized
```

I think this should just pas silently, as the node will already be in it's virgin state. The fix should probably be on `Node` rather than `Model`.

snorfalorpagus:
This PR fixes #82 by initialising _flow in `AbstractNode.__init__`. Otherwise, if `Node.reset` is called before `Model.run`, it attempts to access an uninitialised variable. Added a test that triggers the bug too.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -90,6 +90,8 @@ cdef class AbstractNode:
         self._domain = kwargs.pop('domain', None)
         self._recorders = []
 
+        self._flow = np.empty([0,], np.float64)
+
     property cost:
         """The cost per unit flow via the node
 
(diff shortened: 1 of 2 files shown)


jetuk:
Seems reasonable to me.

Issue #64, opened 2015-10-05
Tests for invalid network structures.


jetuk:
I assume our network has to be acyclic. The path finding algorithm only returns *simple* paths with no repeated nodes. So we need to check for this. `Node.connect` should raise an error if a cycle would be created by the connection?

snorfalorpagus:
My intent was that the model could be cyclic, but only simple paths are evaluated. Consider the following system:
```
from pywr.core import Model, Link, Input, Output
from numpy.testing import assert_allclose

model = Model()

supply1 = Input(model, 'supply1')
supply2 = Input(model, 'supply2')

j1 = Link(model, 'j1', cost=0)  # downhill
j2 = Link(model, 'j2', cost=10)   # uphill, needs to be pumped

demand1 = Output(model, 'demand1', max_flow=5, cost=-999)
demand2 = Output(model, 'demand2', max_flow=5, cost=-999)

supply1.connect(j1)
supply2.connect(j2)
j1.connect(j2)
j2.connect(j1)
j1.connect(demand1)
j2.connect(demand2)

for supply in (supply1, supply2):
    for demand in (demand1, demand2):
        for route in model.find_all_routes(supply, demand):
            print(' -> '.join([node.name for node in route]))

for supply1_flow, supply2_flow in ([0, 10],[10, 0],[5, 5]):
    supply1.max_flow = supply1_flow
    supply2.max_flow = supply2_flow
    model.step()
    assert_allclose(demand1.flow, 5, atol=1e-7)
    assert_allclose(demand2.flow, 5, atol=1e-7)
```

The duplex connection between `j1` and `j2` could represent a single physical connection, that can be used in either direction.

snorfalorpagus:
After discussion on the phone, we need to work out how to handle this:

```
supply1.connect(demand1)
demand1.connect(demand2)
```

Find out if we can replace the demand variable with a constraint, and apply the demand cost to the route instead.

jetuk:
Have started issue #66 for the inline stuff which is different to the cycles.

jetuk:
We also need to handle isolated nodes that are not on any valid route.

snorfalorpagus:
Presumably this needs to be added to `Model.check`?

Issue #31, opened 2015-06-30
Error on unknown keywords to Node


jetuk:
Should `Node.__init__` raise an error when given unknown/unused keywords?

snorfalorpagus:
I have considerred this, but couldn't see an obvious way to implement it given the way `Node` is so frequently subclassed. Must a class keep a record of every valid kwarg for it's parent nodes? Or should each `__init__` method append to a list (can this be done at instance rather than class level?)

jetuk:
Don't you usually do this with crazy metaclass magic that turns class attributes into instance attributes?

snorfalorpagus:
Closed by #85.

PR #85, opened 2016-01-20
Added check for isolated nodes in model structure, and refactored...

snorfalorpagus:
This PR has two changes bundled into 1 commit (that was written on the train). The main purpose was to add a check for nodes that weren't part of a valid route to `Model.check`. This highlighted a reservoir in one of the tests which had nothing connected to it's only output. I fixed that by making `num_outputs=0` valid, but then needed a way to pass that information from XML. The easiest way to do this was to have `Node` accept any keyword arguments. The side effect of this is that it now will accept ANY keyword argument, even if it's not valid/used. This opens the door for typographic errors. However, I'm not sure how to check for "unused" arguments because of the way the objects are subclassed. I think that's focus for another issue/PR.

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -31,6 +31,7 @@ cdef class AbstractNode:
     cdef AbstractNode _parent
     cdef object _model
     cdef object _name
+    cdef bint _allow_isolated
 
     cdef Parameter _cost_param
     cpdef get_cost(self, Timestep ts, int[:] scenario_indices=*)
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -80,6 +80,9 @@ cdef class Domain:
         self.name = name
 
 cdef class AbstractNode:
+    def __cinit__(self):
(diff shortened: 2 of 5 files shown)


snorfalorpagus:
Tag #64.

snorfalorpagus:
Tag #31.

snorfalorpagus:
I think this resolves the kwarg issue. However, it's raised another issue:

If a Node raises an exception during `__init__`, it still gets added to the `model.graph`. I think this needs to come *after* `__init__` returns. This can be done with a metaclass, but causes issues for Cython which can't have a metaclass other than `type`. I think this PR is good to merge, then create another issue for this?

jetuk:
I guess we should write a failing test for the `__init__` exception.

snorfalorpagus:
That's actually done here: https://github.com/pywr/pywr/blob/af2f28d18bad15dae4431a3d87b0ca656a2ed2e7/tests/test_core.py#L36

PR #88, opened 2016-01-23
Added debug infomation for infeasible problems in GLPK.

snorfalorpagus:
This PR adds some debugging information for infeasible problems when using the GLPK solver, using slack variables.

Can you think of a good way to provide access to the information as an argument to the exception? Is a formatted message enough? I'm thinking that it may need to be accessed from a UI in which case just `print()`ing isn't enough.

Also enabled the GLPK presolver, not that it seems to do much. I can't imagine it being a bad thing though.

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -47,6 +47,8 @@ cdef class CythonGLPKSolver:
         cdef int status
         cdef cross_domain_row
 
+        self.smcp.presolve = GLP_ON
+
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
The infeasibility API needs to be able to deal with multiple scenarios. It's possible that some scenarios will be feasible, but others not. Does it need to return an array?

jetuk:
Perhaps, but how does the run proceed if a single problem fails?

I want to think about the LP theory for this a bit. Do we need infeasible model tests for it?

Issue #48, opened 2015-09-04
Renaming a storage node doesn't update child node names


snorfalorpagus:

jetuk:
Make the name property defer to `parent` if not None?

snorfalorpagus:
This creates a bit of a problem for keeping `Model.nodes` updated.

jetuk:
`Model.nodes` could be a generator instead?

snorfalorpagus:
Like this? Seems inefficient, but it's not going to be called all that often.
```
for node in self.graph.nodes():
    if node.name == name:
        return node
raise KeyError()
```

jetuk:
Are we taking about `Model.node` or `Model.nodes` or both?

snorfalorpagus:
Sorry, `Model.node`.

jetuk:
Perhaps we should be more explicit about we treat child nodes? I.e. should they even be accessible in `Model.node`? Should a user fetch the parent and access the child through that?

snorfalorpagus:
Maybe you're right. Perhaps we don't need to add them to `Model.node` at all.

jetuk:
But they returned by `Model.nodes` which would be confusing? I.e. `Model.nodes != Model.node.values()`

snorfalorpagus:
`Model.nodes` just defers to `Model.graph.nodes`. I'm not sure we should be exposing `Model.nodes` and `Model.edges` like this. Perhaps we remove them, rename `Model.node` to `Model.nodes` and make it behave like this:

```
reservoir = model.nodes['myreservoir']
# or
for node in model.nodes:
    pass
```

This would be similar to the way Fiona iterates over features in a shapefile:
```
feature = dataset[42]
# or
for feature in dataset:
    pass
```

If the user really wants to see the graph, they can look at `Model.graph` directly.

jetuk:
And this would only contain parent level nodes? Sounds good to me.

snorfalorpagus:
This needs a flag on the node class, similar to `Node.allow_isolated` in #85.

`Node.show_graph` ?
`Node.hidden` ?
`Node.invisible` ?
???

jetuk:
Can't we use `if self.parent is not None`?

snorfalorpagus:
Closed by #89.

PR #89, opened 2016-01-25
Refactor of Model.nodes using NodeIterator class

snorfalorpagus:
This PR resolves issue #48 (renaming Storage nodes doesn't rename the children) by replacing `Model.nodes` with an instance of `NodeIterator`. The API is unchanged. It also resolves the issue identified in #85, where nodes that raise an exception in their `__init__` were still added to the graph.


jetuk:
This works on my machine, and doesn't seem to slow down the run time.

The code is quite nice.

PR #95, opened 2016-02-01
Reworked the Parameter class

jetuk:
Reworked the Parameter class to contain a reference to its AbstractNode, optional parent Parameter and provide an after() callback hook for the end of the timestep.

This code also modifies the licence module and classes to use the refactored Parameter. Licences are now no longer a special set of objects, but just subclasses of Parameter. The LicenceCollection has been converted to a general ParameterCollection with two specific subclases providing minimum and maximum of the set of Parameters.

Tests updated for AnnualLicence. However existing XML tests with licences now fail and have been marked as such. No tests currently for the ParameterCollection or its subclasses.

### DIFF
--- a/pywr/control_curves.py
+++ b/pywr/control_curves.py
@@ -1,0 +1,33 @@
+"""
+This module contains a set of pywr._core.Parameter subclasses for defining control curve based parameters.
+"""
+
+from ._parameters import Parameter as BaseParameter
+import numpy as np
+
+
+class BaseParameterControlCurve(BaseParameter):
+    """ Base class for all Parameters that rely on a the attached Node containing a control_curve Parameter
+
+    """
+    pass
+
+
+class ParameterControlCurvePiecewise(BaseParameterControlCurve):
+    """ A control curve Parameter that returns one of two values depending on whether the current
+    volume is above or below the control curve.
+
+    """
+    def __init__(self, above_curve_value, below_curve_value):
+        super(ParameterControlCurvePiecewise, self).__init__()
+
+        self.above_curve_value = above_curve_value
+        self.below_curve_value = below_curve_value
+
+    def value(self, ts, scenario_indices=[0]):
+        i = self.node.model.scenarios.ravel_indices(scenario_indices)
+        control_curve = self.node.control_curve.value(ts, scenario_indices)
+        # If level above control curve then return above_curve_cost
+        if self.node.current_pc[i] >= control_curve:
+            return self.above_curve_value
+        return self.below_curve_value
--- a/pywr/control_curves.py
+++ b/pywr/control_curves.py
@@ -19,6 +19,8 @@ class BaseParameterControlCurve(BaseParameter):
             The Parameter object to use as a control_curve. It should not be shared with other
             Nodes and Parameters because this object becomes control_curve.parent
         """
+        if control_curve.parent is not None:
+            raise RuntimeError('control_curve already has a parent.')
         control_curve.parent = self
         self.control_curve = control_curve
 
--- a/pywr/control_curves.py
+++ b/pywr/control_curves.py
@@ -10,7 +10,33 @@ class BaseParameterControlCurve(BaseParameter):
     """ Base class for all Parameters that rely on a the attached Node containing a control_curve Parameter
 
     """
-    pass
+    def __init__(self, control_curve):
+        """
+
+        Parameters
+        ----------
+        control_curve : Parameter
+            The Parameter object to use as a control_curve. It should not be shared with other
+            Nodes and Parameters because this object becomes control_curve.parent
+        """
+        control_curve.parent = self
+        self.control_curve = control_curve
+
+    def setup(self, model):
+        self.control_curve.setup(model)
+        super(BaseParameterControlCurve, self).setup(model)
+
+    def reset(self):
+        self.control_curve.reset()
+        super(BaseParameterControlCurve, self).reset()
+
+    def before(self, ts):
+        self.control_curve.before(ts)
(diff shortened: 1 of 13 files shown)


pywr/parameters.py
```diff
@@ -27,6 +27,62 @@ def from_xml(cls, model, xml):
...
 
+class ParameterCollection(Parameter):
+    """A collection of Parameters
+
+    This object behaves like a set. Licenses can be added to or removed from it.
+
+    """
+    def __init__(self, parameters=None):
+        if parameters is None:
+            self._parameters = []
```
snorfalorpagus: This should be a set, not a list.

```
self._parameters = set()
```

tests/test_license.py
```diff
@@ -14,48 +16,47 @@ def test_daily_license():
...
-    assert(len(collection) == 2)
-
-    assert(collection.available(Timestep(datetime(2015, 1, 1), 0, 0)) == 42.0)
-    assert(collection.resource_state(Timestep(datetime(2015, 1, 1), 0, 0)) == 1.0)
-    assert(collection.resource_state(Timestep(datetime(2015, 2, 1), 0, 0)) > 1.0)
-
-    collection.commit(0, 360.0)
-    assert(collection.available(Timestep(datetime(2015, 12, 1), 0, 0)) == 5.0)
-    assert(collection.resource_state(Timestep(datetime(2015, 12, 1), 0, 0)) < 1.0)
+def test_simple_model_with_annual_licence(simple_linear_model):
```
snorfalorpagus: Need also to test that annual licence resets to zero at the start of the year, or if `Model.reset()` is called.

jetuk:
Should `Parameter.setup` take a `model` argument now the objects have access to the `Node` they are attached to?

snorfalorpagus:
Can you just use Node.model?

jetuk:
Exactly ...

jetuk:
There's a problem with the `AnnualLicense` reset at year end. I'd appreciate if you could look at the failing test and see what you think. 

It fails because the timestep does land on 1st January and therefore there is more licence available than there should be. I.e. it is assuming no licence is consumed between the 1st Jan and the first timestep in each year. I think we should assume the rate of consumption is the same as the previous timestep before the reset, but not sure how best to deal with that.

tests/test_license.py
```diff
@@ -14,48 +16,71 @@ def test_daily_license():
...
 
-    assert(len(collection) == 2)
+def test_simple_model_with_annual_licence_multi_year(simple_linear_model):
+    """ Test the AnnualLicense over multiple years
+    """
+    import pandas as pd
+    import datetime, calendar
+    m = simple_linear_model
+    # Modify model to run for 5 years on 30 days timestep
+    m.timestepper.end = pd.to_datetime('2020-12-31')
```
snorfalorpagus: Should probably be explicit about start year here too.

snorfalorpagus:
I understand the issue you've hit with this. I agree that it needs some discussion.

The issue is that the timestep might cross a (somewhat artificial) boundary not cleanly (i.e. part of the timestep is in one year, the other in the next). I think the correct thing to do in this instance is to return the volume expected to be accessed in that timestep. That will involve adding what's remaining in this year to what's acceptable next year. Unfotunately this adds a fair amount of code to the function but I don't think it can be avoided.

The other approach would be to ensure the timesteps are divisible in an even, sensible manner. e.g. 52 weeks in a year exactly, or 12 months exactly.

jetuk:
As we discussed this PR now contains `Reservoir` without a `control_curve` attribute, but provides that functionality in a new set of classes.

Also have a look at the new licence subclasses. Perhaps they should be renamed to remove "Cost"?

pywr/licenses.py
```diff
@@ -109,54 +135,33 @@ def xml(self):
...
-            resource_states.append(license.resource_state(timestep))
-        resource_states = [r for r in resource_states if r is not None]
-        return min(resource_states)
-    def commit(self, scenario_index, value):
-        for license in self._licenses:
-            license.commit(scenario_index, value)
-    def reset(self):
-        for license in self._licenses:
-            license.reset()
+class AnnualExponentialCostLicense(AnnualLicense):
```
snorfalorpagus: Needs test.

pywr/licenses.py
```diff
@@ -109,54 +135,33 @@ def xml(self):
...
-            licenses.add(license)
-        return LicenseCollection(licenses)
+    def value(self, timestep, scenario_indices=np.array([0], dtype=np.int32)):
+
+        remaining = super(AnnualExponentialCostLicense, self).value(timestep, scenario_indices)
+        expected = self._amount / (365 + int(calendar.isleap(timestep.datetime.year)))
+        x = remaining / expected
+        return self._max_cost * np.exp(-x/self._k)
+
+class AnnualHyperbolaCostLicense(AnnualLicense):
```
snorfalorpagus: Needs test.

pywr/control_curves.py
```diff
@@ -0,0 +1,76 @@
...
+    """ Base class for all Parameters that rely on a the attached Node containing a control_curve Parameter
+
+    """
+    def __init__(self, control_curve):
+        """
+
+        Parameters
+        ----------
+        control_curve : Parameter
+            The Parameter object to use as a control_curve. It should not be shared with other
```
snorfalorpagus: This is a potential pitfall. I imagine it not uncommon to want to share the same control curve between the cost and max_flow parameters. Perhaps this function should first check if `control_curve.parent` is already set and raise an exception if it is?

jetuk:
Addressed comments.

snorfalorpagus:
Merged.

PR #104, opened 2016-02-19
Nodes can now access current model implicitly

snorfalorpagus:
Globals are bad, but I'm stick of having to type `model` every time I create a `Node`. Now you can do this:

```
model = Model(solver='glpk')
supply = Input('supply1', max_flow=20)
```

I think this makes the model API generally look cleaner. Most of the time we'll be working with just one `Model` object anyway.

The current model is tracked by a global variable, accessable via `pywr.core.gcm()`. You can also clear the current model with `clm()`.

I've had to update all the tests to work with this, as most passed model as a sequential argument not a keyword argument. However, I've also added some new tests - have a look at the end of `test_core.py`:

https://github.com/pywr/pywr/blob/implicit_model/tests/test_core.py#L277-L324

If you try to create a `Node` before `Model` it will raise an exception. If you try to connect a node to another node in a different model it will also complain.

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -160,6 +160,14 @@ class NodeIterator(object):
         # support for old API
         return self
 
+model = None
+def gcm():
+    global model
+    return model
+def clm():
+    global model
+    model = None
+
 class Model(object):
     """Model of a water supply network"""
     def __init__(self, **kwargs):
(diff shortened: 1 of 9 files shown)


pywr/core.py
```diff
@@ -160,6 +160,14 @@ def __call__(self):
         # support for old API
         return self
 
+model = None
```
jetuk: should this be `__model` to prevent accident importing ?

jetuk:
This should also be extended to the `Recorder` and its subclasses.

Issue #103, opened 2016-02-18
Most (all?) Parameter subclasses don't call super().__init__


jetuk:
They probably ought to do this in case we had some common `__init__` code later.

snorfalorpagus:
Agreed. We can fix this at the same time as #101.

Issue #101, opened 2016-02-15
Naming convention


jetuk:
This is probably my fault but, `Parameter...` vs `...Recorder` is inconsistent.

snorfalorpagus:
I hadn't even noticed. Which do you prefer?

jetuk:
The second.

Issue #97, opened 2016-02-02
Add harmonic series Parameter


jetuk:
Add an implementation of an harmonic series for Parameters. Inspired by Richardson (1981) paper on stochastic rainfall modelling. Here the parameters of the Markov Chain vary throughout the year based on an harmonic series. A similar approach could be useful for `Reservoir` control curves amongst other things.

The nice thing would be that series with a single harmonic is only defined by three variables which makes it nice to opimise.

Pseduo code,
```
T = 365 if not leap_year else 366
x(t) = C_0 + sum(C_i * cos(2*pi*i*doy/T + theta_i) for i in range(1, N+1))
where,
     N is the number of harmonics in the series
```

jetuk:
Fixed with #106 merge.

Issue #16, opened 2015-02-02
Standardise approach to Genetic Algorithms (GA)


snorfalorpagus:
Example of using a GA to generate a reservoir control curve:

https://github.com/snorfalorpagus/pywr/blob/ga/examples/example_ga1.py

Investigate adding some standard structures for this kind of thing:
- [ ] Reservoir control curves
- [ ] Individual parameters

jetuk:
I propose we do this in a `pywr.optimisation`  package. 

I currently have a separate project / repo to do interaction with Borg, but it is not the only MOEA solution. The interaction between `Parameters` and `Recorders` should be agnostic to the MOEA being used and therefore can be implemented in the main `pywr` repository. 

I envisage the same type of approach as with the LP solvers. We provide a generic way to provide multiple MOEA solvers using the same set of core objects.

I will keep the Borg solver outside of MOEA because of the licence problem, but we can add one for say [inspyred](http://aarongarrett.github.io/inspyred/) in the main repo.

snorfalorpagus:
```
class BorgModel(Model):
    """
    Extension to the pywr.core.Model to add features for using Borg MOEA.
    """
    def __init__(self, *args, **kwargs):
        super(BorgModel, self).__init__(*args, **kwargs)
        self.problem = None
        # Container for BorgVariable objects
        self.variables = []
        self._variable_map = None
        # Container for BorgObjective objects
        self.objectives = []

    def setup(self, ):
        """
        Setup the model and Borg problem before solving.
        """
        super(BorgModel, self).setup()
        # Create index map from variable array to BorgVariable objects
        variable_map = [0]
        for var in self.variables:
            variable_map.append(var.size)

        # Create the BorgProblem object with correct number of variables
        print(variable_map[-1], len(self.objectives))
        self.problem = PywrProblem(self, variable_map[-1], len(self.objectives), 0)

        # Apply bounds
        for i, var in enumerate(self.variables):
            var.set_bounds(variable_map[i])

        for i, obj in enumerate(self.objectives):
            obj.set_epsilon(i)

        self._variable_map = variable_map

    def evaluate(self, vars, objs, cons):
        """
        Evalute the current variables
        """
        # Update BorgVariable objects with new values
        for i, j, var in zip(self._variable_map[:-1], self._variable_map[1:], self.variables):
            var.update(vars[i:j])
```
Note: BORG is not GPL-compatible.

I think we can integrate two parts of this into the vanilla pywr code:

[...]

PR #106, opened 2016-02-21
Optimisation

jetuk:
This branch is getting a bit out of hand. 

I have addressed #16, #97, #101 and #103 here so far. There is also an MOEA example included.

Still to complete:
- [x] Add `Recorder` subclasses in MOEA example to `pywr.recorders`
- [x] Decide whether the `OptimisationModel` subclass should be left in the example or also added to `pywr` proper.


jetuk:
Possibly the only outstanding thing is to add `Recorder.value` as a not implemented default method on the `Recorder` base class.

snorfalorpagus:
The build failures are rounding errors. Should be easy to fix.

jetuk:
Anything else to add to this now?

pywr/_core.pyx
```diff
@@ -57,9 +57,11 @@ cdef class ScenarioCollection:
         def __get__(self):
             if len(self._scenarios) == 0:
                 return (1, )
-            return (sc._size for sc in self._scenarios)
+            return tuple(sc.size for sc in self._scenarios)
 
-    cpdef int ravel_indices(self, int[:] scenario_indices):
+    cpdef int ravel_indices(self, int[:] scenario_indices) except? -1:
```
snorfalorpagus: Is `-1` a valid return from this function if an exception hasn't occured? If not the `?` isn't required.

snorfalorpagus:
This looks good to me. Do you want to merge it?

pywr/_core.pyx
jetuk: You make a good point, but #109 rewrites this function anyway.

PR #115, opened 2016-03-20
Fix105

jetuk:
This commit allows `Storage.current_pc` to be calculated when maximum volume is a `Parameter`.


snorfalorpagus:
Merged.

PR #124, opened 2016-04-16
Add a name to Recorder and use RecorderIterator for access in Model.

jetuk:
This approach is similar to NodeIterator but is more simple. RecorderIterator is a wrapper around a list of recorders but provides access through keys and also iteration. The used API for Model.recorders is not significantly changed, but now provides access via name.

Recorders have a default name that is the the lowercase of the class name. NodeRecorder and StorageRecorder provide a default naming include the node's name.

### DIFF
--- a/pywr/_recorders.pxd
+++ b/pywr/_recorders.pxd
@@ -4,6 +4,7 @@ from _core cimport Timestep, AbstractNode, Storage
 
 cdef class Recorder:
     cdef bint _is_objective
+    cdef object _name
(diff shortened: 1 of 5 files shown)


snorfalorpagus:
Looks good to me.

snorfalorpagus:
(this isn't finished yet)

jetuk:
I think it's finished now. I thought there were more subclasses to fix than there was!

PR #131, opened 2016-06-09
Improved exception when running a model with no valid routes.

### LINKED ISSUE #129: Better exception for model with no routes
This isn't enough:
```
>   assert(routes)
E   AssertionError
```

snorfalorpagus:
As title: Improved exception when running a model with no valid routes.

Closes #129.


jetuk:
Looks fine to me.

PR #134, opened 2016-06-13
Fix #130; Unhelpful exception when Storage.max_volume == 0

### LINKED ISSUE #130: Unhelpful exception when Storage.max_volume == 0
The exception raised when max_volume == 0 is not helpful.

Also, should check that max_volume > initial_volume > 0.

```
    model = Model(
        solver=solver,
        start=pandas.to_datetime('2016-01-01'),
        end=pandas.to_datetime('2016-01-01')
    )

    storage = Storage(model, 'storage', num_inputs=1, num_outputs=0)
    otpt = Output(model, 'output', max_flow=99999, cost=-99999)
    storage.connect(otpt)

    storage.max_volume = 0

    model.run()
```

```
pywr\core.py:377: in run
    self.setup()
pywr\core.py:406: in setup
    self.reset()
pywr\core.py:413: in reset
    node.reset()
pywr\_core.pyx:574: in pywr._core.Storage.reset (pywr/_core.c:15273)
    cpdef reset(self):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   yield np.array(comb, dtype=np.int32)
E   ZeroDivisionError: float division

pywr\_core.pyx:12: ZeroDivisionError
```

jetuk:
As per subject. Fixed by raising a `ValueError` instead.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -589,8 +589,10 @@ cdef class Storage(AbstractNode):
             # Ensure variable maximum volume is taken in to account
             if self._max_volume_param is not None:
                 mxv = self._max_volume_param.value(self.model.timestepper.current, si)
-
-            self._current_pc[i] = self._volume[i] / mxv
+            try:
+                self._current_pc[i] = self._volume[i] / mxv
+            except ZeroDivisionError:
+                raise ValueError("Zero maximum volume is invalid on Storage object: {}".format(self))
 
     cpdef before(self, Timestep ts):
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
I'm not sure if this is the correct approach. If we're managing options using scenarios (i.e. same LP structure) it might make sense to have a reservoir volume equal to zero.

PR #136, opened 2016-06-14
Add DataFrameParameter

### LINKED ISSUE #135: Automate dataframe creation from recorder results
It would be useful to have an automated way to create a pandas dataframe from recorders after the model has been run. This has two parts:

- Create a pandas date index from the timestepper
- Create the dataframe from a list of recorder names (needs to be scenario-aware)

The former will be useful in #127.

```
index = pandas.date_range(model.timestepper.start, model.timestepper.end, freq="D")

df = pandas.DataFrame({
        "A": model.recorders["A"].data[:,0],
        "B": model.recorders["B"].data[:,0],
        "C": model.recorders["demand_supplied"].data[:,0]
    }, index=index)
```

jetuk:
Fixes #135

Includes
- `Timestepper.datetime_index` property to return a `pandas.DatetimeIndex`
- `ScenarioCollection.combination_names` to return a generator of scenario combination names
- Added `to_dataframe` method to the three numpy array based recorders
- Added `Model.to_dataframe` method that concatenates any recorders that can produce a dataframe
- Some extra tests of the above


jetuk:
Make the `MultiIndex` have a level for each `Scenario`, and use integer based indexes for each. Removes the need for horrible scenario name combinations.

snorfalorpagus:
The current build is failing because `range` returns a `list` in Python 2.x, but a generator in Python 3.x. Just need to wrap it: `list(range(365))`.

jetuk:
I fixed that. Appveyor is failing because the path is different for running the tests. How do I get the path in the json file to work regardless of the location the tests are run from?

jetuk:
I think this is ready to go now; can you review?

pywr/core.py
```diff
@@ -358,7 +367,7 @@ def load(cls, data, model=None, path=None, solver=None):
                 timestep=timestep,
                 path=path,
             )
-
+        print(path)
```
snorfalorpagus: Remove this.

pywr/parameters/_parameters.pyx
```diff
@@ -249,45 +249,53 @@ def load_parameter(model, data):
...
         try:
-            values = df[column].values
+            df = df[column]
         except KeyError:
             raise KeyError('Column "{}" not found in dataset "{}"'.format(column, url))
-    
-    return values
+
+    # Convert to regular frequency
+    df = df.asfreq(df.index.inferred_freq)
```
snorfalorpagus: `index.inferred_freq` will return `None` if there is an issue with the data (e.g. missing values). We should capture this and raise a useful exception.

tests/test_core.py
```diff
@@ -158,6 +158,7 @@ def test_slots_from(solver):
     assert(len(model.edges()) == 0)
 
 
+@pytest.mark.xfail
```
snorfalorpagus: These tests should be fixed not skipped.

snorfalorpagus:
Unless I am mistaken I can't see a test/example of accessing the dataframe produced from recorders to access a specific value (e.g. node X, scenario A1+B2, timestep 5).

snorfalorpagus:
Merged as of ca4d1019651db320c959fcf0f8b4bdecdbf0e87c.

PR #138, opened 2016-06-14
Unhelpful exception when Storage.max_volume == 0

jetuk:
Fiox #130. As per subject. Fixed by returning np.nan


snorfalorpagus:
Merged as 004f21ed4007c1e4956d048e5d4e11dd57df6462.

PR #146, opened 2016-06-16
Improvements to RiverGauge

### LINKED ISSUE #141: Improvements to RiverGauge class
We need to make two improvements to `RiverGauge`.

1. Add `mrf_cost` property to the class to allow modification of the weight on the sublink without accessing it directly.
2. Add support for setting the MRF flow and cost to non-constant values in JSON.

See Reservoir.load for hints on 2.

snorfalorpagus:
- Added RiverGauge.mrf_cost property
 - Added JSON support for non-constant MRF values and costs

Closes #141.


snorfalorpagus:
Should be using a recorder in the tests, rather than accessing `.flow` directly...

jetuk:
This just needs `assert_allclose` for the float assertions instead.

Issue #140, opened 2016-06-16
Circular route crashes GLPK solver


snorfalorpagus:
Circular routes seem to crash the GLPK solver! I thought we had a test for this, but apparently not.

The error raised is below:
```
glp_set_mat_row: i = 16; ind[5] = 4; duplicate column indices not allowed
Error detected in file ..\src\glpapi01.c at line 776
```

This model triggers the issue.

```
from pywr.core import Model, Input, Output, Storage, Link, Catchment, RiverGauge, River

model = Model(solver="glpk")

catchment = Catchment(model, "catchment", flow=100)
reservoir = Storage(model, "reservoir", max_volume=10000, initial_volume=5000)
gauge = RiverGauge(model, "mrf", mrf=22, mrf_cost=-1000)
demand = Output(model, "demand", max_flow=50, cost=-100)
pumping_station = Link(model, "pumping station", max_flow=100)
river1 = River(model, "river1")
river2 = River(model, "river2")
waste = Output(model, "waste")

catchment.connect(river1)
river1.connect(river2)
river2.connect(gauge)
gauge.connect(waste)
river2.connect(pumping_station)
pumping_station.connect(reservoir)
reservoir.connect(river1)
reservoir.connect(demand)

model.check()
model.run()
```

snorfalorpagus:
Closed by #148.

PR #148, opened 2016-06-17
Fixed issue with circular route around Storage in GLPK solver.

snorfalorpagus:
Fixes issue #140.

The problem was when a reservoir has a route that goes from one of it's Inputs to another of it's Outputs.

In the GLPK solver this resulted in the algorithm trying to set the value for the row that constrains changes in the volume of the reservoir twice: +1 for one, -1 for the other.

The fix is to detect this condition and not modify the matrix (i.e. default to 0). I *think* this is the correct thing to do, although if the +1 and -1 didn't cancel out, perhaps due to some loss in the route (which we don't currently implement) this may need more though.

Please you you review this @jetuk. It's the kind of thing that's easy to screw up.

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -140,8 +140,8 @@ cdef class CythonGLPKSolver:
         if len(storages):
             self.idx_row_storages = glp_add_rows(self.prob, len(storages))
         for col, storage in enumerate(storages):
-            cols_output = [n for n, route in enumerate(routes) if route[-1] in storage.outputs]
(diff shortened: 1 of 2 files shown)


jetuk:
I think this is correct. It's a limitation (perhaps not) of the GLPK API to not allow two entries for the same column when passing a row. It would appear however the lpsolve does allow this because it passes the test with no modification.

I don't think there is a problem with potential losses because we don't allow conversion along a route (i.e. within the same domain).

snorfalorpagus:
It's not clear from the lpsolve documentation what happens in this case, but it seems to work.

Issue #143, opened 2016-06-16
Demand saving controller


snorfalorpagus:
We need a generic approach to demand saving. Previously this has been done in a "controller" object.

- The demand level should be calculated only once
- This is usually based on reservoir volume(s) in relation to a control curve(s)
- It would be useful to be able to have an aggregate demand saving level, based on the worst-case demand saving level of two or more reservoirs.
- We can then wrap the demand max_flow parameter in a demand saving adjustment parameter.

Need to be able to define this using JSON.

snorfalorpagus:
The same controller could also be used to trigger changes to an MRF value in drought conditions.

I think we probably need to have a new `"controllers"` section in the JSON.

jetuk:
We could add a new property to `Reservoir`, similar to `level` that can take a parameter but is intended to return the demand saving level? Or is this specialising the class too much?

Why can't we have a single controller `Parameter` in the `"parameters"` section that is reused by many nodes? It would have to do its calculations in `before` rather than `value` to be most efficient.

snorfalorpagus:
I don't think calculating the demand saving level directly on a reservoir is the right way to go. Often the demand saving will be based on multiple reservoirs. Sometimes it's not based on a reservoir at all (e.g. for WRZs that have no storage, demand saving might be based on river flows).

Putting it in the `"parameters"` section is a good idea. Same for `before`.

We also need to be able to record the demand saving level. Do we need a recorder that takes a parameter directly, rather than a node?

jetuk:
Here's how I would do this

A new subclass of `BaseControlCurveParameter` something like this. It assumes that self.control_curve is a `Parameter` that returns an iterable of multiple curves and then returns the first failure.
```
class MultiControlCurvePosition(BaseControlCurveParameter):

    def value(self, ts, scenario_index):
        i = scenario_index.global_id
        control_curves = self.control_curve.value(ts, scenario_index)
        node = self.node if self.storage_node is None else self.storage_node
        # Assumes control_curves is sorted highest to lowest

        for j, cc in enumerate(control_curves):
            if node.current_pc[i] >= cc:
                return j
        return len(control_curves)
```
You would then use this with `MaximumParameterCollection` to pass a number of them for different reservoirs and return the worst. 

[...]

PR #149, opened 2016-06-19
Big refactor of control curve classes.

jetuk:
The following is included here:
- Allow `BaseControlCurveParameter` to accept multiple control curves
- Initialisation of control curves from floats, ints or `Parameter` objects
- Private classmethods `_load_control_curves` and `_load_storage_node` to load their respective items from a dict of data - intended for JSON support.
- Refactored `ControlCurvePiecewiseParameter` to `ControlCurveParameter`.
  - This is now the basic generic control curve class.
  - It supports returning either values from an array or values from other `Parameters`.
  - Can be loaded from a dict.
  - Test of this are included.
- Refactored `ParameterCollection` to `AggregatedParameter`
  - Similar approach `AggregatedRecorder`.
  - No separate subclasses. Instead the user specifies the `agg_func` to use.
  - Includes dict loading.

I think this fixes 143. Though it probably needs an example of how to do that.

### DIFF
--- a/pywr/parameters/_control_curves.pyx
+++ b/pywr/parameters/_control_curves.pyx
@@ -1,49 +1,69 @@
 import numpy as np
 cimport numpy as np
-from .parameters import parameter_registry
-from ._parameters import load_parameter
+from .parameters import parameter_registry, ConstantParameter
+from ._parameters import load_parameter, load_parameter_values
+
 
 cdef class BaseControlCurveParameter(Parameter):
     """ Base class for all Parameters that rely on a the attached Node containing a control_curve Parameter
 
     """
-    def __init__(self, control_curve, Storage storage_node=None):
+    def __init__(self, control_curves, Storage storage_node=None):
         """
 
         Parameters
         ----------
-        control_curve : Parameter
-            The Parameter object to use as a control_curve. It should not be shared with other
+        control_curve : iterable of Parameter objects or single Parameter
+            The Parameter objects to use as a control curve(s). These should not be shared with other
             Nodes and Parameters because this object becomes control_curve.parent
         """
         super(BaseControlCurveParameter, self).__init__()
-        if control_curve.parent is not None:
-            raise RuntimeError('control_curve already has a parent.')
-        control_curve.parent = self
-        self._control_curve = control_curve
+        self.control_curves = control_curves
         self._storage_node = storage_node
 
     cpdef setup(self, model):
-        self.control_curve.setup(model)
+        cdef Parameter cc
+        for cc in self.control_curves:
+            cc.setup(model)
         super(BaseControlCurveParameter, self).setup(model)
 
     cpdef reset(self):
-        self.control_curve.reset()
+        cdef Parameter cc
+        for cc in self.control_curves:
+            cc.reset()
         super(BaseControlCurveParameter, self).reset()
 
     cpdef before(self, Timestep ts):
-        self.control_curve.before(ts)
+        cdef Parameter cc
+        for cc in self.control_curves:
+            cc.before(ts)
         super(BaseControlCurveParameter, self).before(ts)
 
     cpdef after(self, Timestep ts):
-        self.control_curve.after(ts)
+        cdef Parameter cc
+        for cc in self.control_curves:
+            cc.after(ts)
         super(BaseControlCurveParameter, self).after(ts)
 
-    property control_curve:
+    property control_curves:
         def __get__(self):
-            return self._control_curve
-        def __set__(self, value):
-            self._control_curve = value
+            return self._control_curves
+        def __set__(self, control_curves):
+            # Accept a single Parameter and convert to a list internally
+            if isinstance(control_curves, Parameter):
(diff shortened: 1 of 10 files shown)


pywr/parameters/_control_curves.pyx
```diff
@@ -1,94 +1,165 @@
...
+from .parameters import parameter_registry, ConstantParameter
+from ._parameters import load_parameter, load_parameter_values
+
 
 cdef class BaseControlCurveParameter(Parameter):
     """ Base class for all Parameters that rely on a the attached Node containing a control_curve Parameter
 
     """
-    def __init__(self, control_curve, Storage storage_node=None):
+    def __init__(self, control_curves, Storage storage_node=None):
```
snorfalorpagus: Add storage_node to docstring

pywr/parameters/_control_curves.pyx
```diff
@@ -1,94 +1,165 @@
...
+        if 'storage_node' in data:
+            return model.node[data['storage_node']]
+        return None
+
+
 parameter_registry.add(BaseControlCurveParameter)
 
 
 cdef class ControlCurveInterpolatedParameter(BaseControlCurveParameter):
     """ A control curve Parameter that interpolates between three values.
```
snorfalorpagus: I think this needs to be upgrade to 4 values to allow for a discontinuity.

pywr/parameters/control_curves.py
```diff
@@ -3,79 +3,142 @@
...
+        The values to return if the `Storage` object is above the correspond control curve.
+        I.e. the first value is returned if the current volume is above the first control curve,
+        and second value if above the second control curve, and so on. The length of `values`
+        must be one more than than the length of `control_curves`. I
+    parameters : iterable `Parameter` objects or `None`, optional
+        If `values` is `None` then `parameters` can specify a `Parameter` object to use at level
+        of the control curves. In the same way as `values` the first `Parameter` is used if
+        `Storage` is above the first control curve, and second `Parameter` if above the
+        second control curve, and so on.
+    storage_node : `Storage` or `None`, optional
```
snorfalorpagus: We need to provide the facility to use the combined storage of multiple reservoirs. There are two approaches to this:

1. The control curve needs a `storage_nodes` instead of `storage_node` and the calculations should be done assuming a list of nodes rather than a single node.
2. Add a new "group node" which would behave as a storage node, but underneath would aggregate the percentage calculation.

Both approaches have merit. Option 2 makes the control curve code simpler, and also opens the door for things like group constraints (on non-storage nodes) later (e.g. `group.max_flow = x`). Option 1 can be done within the existing framework. Thoughts?

pywr/parameters/control_curves.py
```diff
@@ -3,79 +3,142 @@
...
+
+     This parameter can be used to return different values when a `Storage` node's current
+      volumes is at different percentage of `max_volume` relative to predefined control curves.
+      Control curves must be defined in the range [0, 1] corresponding to 0% and 100% volume.
+
+     By default this parameter returns an integer sequence from zero if the first control curve
+      is passed, and incrementing by one for each control curve (or "level") the `Storage` node
+      is below.
+
+    Parameters
```
snorfalorpagus: This belongs in the `__init__` docstring. The docstring for the class should include "Attributes" (with some duplication of names in this case) and "Methods" (which I don't think we need atm). https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-classes

snorfalorpagus:
This looks good to me. I'd like to see support for multiple reservoirs and a single control curve (see comment above) before merging.

I think this does address #143, but the additional changes required could be in another PR.

pywr/parameters/control_curves.py
jetuk: That page says the `__init__`  documentation belongs in the class docstring?

snorfalorpagus: Ooops. I read that wrong...

pywr/parameters/control_curves.py
```diff
@@ -3,79 +3,295 @@
...
+            parameters = []
+            for pdata in parameters_data:
+                parameters.append(load_parameter(model, pdata))
+
+        return cls(control_curves, values=values, parameters=parameters, storage_node=storage_node)
 
     def value(self, ts, scenario_index):
         i = scenario_index.global_id
-        control_curve = self.control_curve.value(ts, scenario_index)
         node = self.node if self.storage_node is None else self.storage_node
```
snorfalorpagus: This could be moved to `__init__`, i.e. if `storage_node` is not specified, set it to self.node.

snorfalorpagus:
Can you talk to me before pushing anything more on this branch?

pywr/parameters/control_curves.py
jetuk: `self.node` is the `Node` the parameter is attached to, which could be different  to `storage_node` so I don't think that's a good idea. It could be set to a different private attribute e.g. `self._storage_node`

Issue #122, opened 2016-04-11
Cache for Parameters


snorfalorpagus:
Some `Parameters` will benefit from being cached.

```
from pywr.core import CacheParameter
cache = CacheParameter(parameter)
node.max_flow = cache
```

The cache only needs to retain values for the current timestep - if a query is made for another timestep the cache can be flushed.

It will be up to the user to determine when it's appropriate to cache a parameter. It should be useful for any parameter that's being called more than once in a timestep, where the result is more than a simple lookup.

snorfalorpagus:
We'll also need to refresh the cache if the scenario switches. Do you think something like this will be enough:

```
class CachedParameter(Parameter):
    def __init__(self, parameter):
        self._parameter = parameter
        self.timestep = None
        self.scenario_index = None
        self.cached_value = None
    def value(self, timestep, scenario_index):
        if not timestep is self.timestep or not scenario_index is scenario_index:
            # refresh the cache
            self.cached_value = self._parameter.value(timestep, scenario_index)
            self.timestep = timestep
            self.scenario_index = scenario_index
        return self.cached_value
```

The JSON syntax would look like this:

```
{
    "name": "catchment",
    "type": "catchment",
    "flow": {
        "type": "cached",
        "parameter": {
            "type": "arrayindexed",
            ...
        }
    }
}
```

snorfalorpagus:
I might even be tempted to make a special case for this:

```
{
    "name": "catchment",
    "type": "catchment",
    "flow": {
        "cached": true,
        "type": "arrayindexed",
        ...
    }
}
```

PR #156, opened 2016-06-24
Added CachedParameter and fixed bug when loading shared params.

snorfalorpagus:
This PR addresses #122. Parameters can now be cached easily using the `CachedParameter` class. In JSON this can be written as:

```
{
    "cached": true,
    "type": ...
    "values": ...
}
```

The cache is refreshed if the timestep or scenario index is changed. A future change might be to create a subclass which only refreshes if the timestep changes, for parameters that are time-varying but constant between scenarios and expensive to calculate, e.g. a catchmod boundary condition.

As a benchmark, caching the demand saving level in the London model resulted in a speed increase of 10%.

The PR also fixes an issue with the revised loading of parameters of JSON, whereby parameters that were used more than once were loaded multiple times.

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -382,6 +382,10 @@ class Model(object):
             parameters_to_load = data["parameters"]
         except KeyError:
             parameters_to_load = {}
+        else:
+            for key, value in parameters_to_load.items():
+                if isinstance(value, dict):
+                    parameters_to_load[key]["name"] = key
         model._parameters_to_load = parameters_to_load
 
         # load the remaining nodes
--- a/pywr/parameters/_parameters.pxd
+++ b/pywr/parameters/_parameters.pxd
@@ -21,6 +21,7 @@ cdef class Parameter:
     cpdef update(self, double[:] values)
     cpdef double[:] lower_bounds(self)
     cpdef double[:] upper_bounds(self)
+    cdef public str name
 
 cdef class ArrayIndexedParameter(Parameter):
     cdef double[:] values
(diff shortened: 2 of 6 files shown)


jetuk:
I guess you could use "both", "time-only", or "scenario-only" for the different cases? Caching in time is obviously easier. So the first two cases could be easily accommodated here, but the caching by scenario only is more difficult and less useful?

```
{
    "cached": "both",
    "type": ...
    "values": ...
}
```

```
{
    "cached": "time-only",
    "type": ...
    "values": ...
}
```

jetuk:
Happy to look at merging and fixing the conflicts in this if you would like? Let me know if the suggestion above is useful now.

snorfalorpagus:
Should we have subclasses for the different types of caching? Or an if-statement in the `index`/`value` methods? Would the former be quicker?

jetuk:
I think subclasses is the better method.

snorfalorpagus:
The only downside to subclasses that I can think of it that it's not possible to change the method after the instance is created. That's not a big deal though. I'll sort this.

snorfalorpagus:
I've made a start towards caching in time or scenario. Need to get our names sorted out. Currently `CachedTimeParameter` is a cache that is constant between scenarios and varies with time. Perhaps that is the wrong way around? Anyway, the inverse (varying in scenario but not time) is harder to implement.

jetuk:
I would not implement the inverse (varying in scenario but not time) at the moment.

snorfalorpagus:
I've merged the basic CachedParameter object. Let's save the more complicated parameters for another day.

PR #161, opened 2016-06-28
Added some standard figures for calibration

snorfalorpagus:
This PR adds some standard matplotlib figures for use when calibrating a model.

It includes a limited function for aligning timeseries (`align_series`) which could be improved at some point.


jetuk:
Looks good. I've not tested it. Only one comment on the use of `pandas.concat` for the aligning instead.

Issue #154, opened 2016-06-23
Nodes coordinates


khaledk2:
Nodes coordinates are needed mainly for network visualization in PywrApp, as JSON does not contains this information, I set random numbers for the nodes coordinates. 
If Pywr doesn't need nodes coordinates to function, I would suggest having a separate csv file which contains the coordinates, what do you think?

snorfalorpagus:
Coordinates aren't required to run the model, but they are useful for visualisation.

Pywr does currently support storing an x,y coordinate pair on a node:

https://github.com/pywr/pywr/blob/master/pywr/core.py#L724-L729

To set these properties from JSON use the "position" keyword. e.g.

```
        {
            "name": "supply1",
            "type": "Input",
            "max_flow": 15,
            "cost": 1,
            "position": [5, -20]
        },
```

I've also been considering allowing for multiple pairs of coordinates. This would allow a "schematic" view as well as a geographic coordinate for overlaying on a map. I'm not sure what this would look like in JSON. Maybe: `{"schematic_position": [5, -20], "geographic_position": [-1.509445, 51.30284]}`.

khaledk2:
Not all the nodes have position in this example:

https://github.com/pywr/pywr/blob/master/model.json

only catchment1 has a position and the other nodes do not
It is needed that all nodes have coordinates to visualize the network.

khaledk2:
I think it is a good idea to have multiple pairs of coordinates.

jetuk:
With geographic position would we need to define the coordinate system somewhere? Should this be a the top or with each position pair?

snorfalorpagus:
My preference would be to do the same as GeoJSON - require coordinates to be given in WGS84 (i.e. latitude and longitude). Most of the JavaScript libraries are set up to work in with LatLngs.

khaledk2:
GeoJSON format is a good option, this would be my preference also

jetuk:
That's agreed then, and it's simple! So the remaining question is do we set it up like,

```
{"schematic_position": [5, -20], "geographic_position": [-1.509445, 51.30284]}
```

or

```
{"position": {
    "schematic": [5, -20],
    "geographic": [-1.509445, 51.30284]
} }
```

khaledk2:
I would prefer the second format, i.e. 
{"position": {
    "schematic": [5, -20],
    "geographic": [-1.509445, 51.30284]
} }

jetuk:
:+1:

[...]

PR #163, opened 2016-06-28
Added schematic and geographic position attributes to Node

snorfalorpagus:
This commit implements the schematic and geographic position attributes discussed in #154.

Note that it's still valid for these properties to be `None` (or `null` in JSON). I don't want to enforce providing them as it's unnecessary for simple models.

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -750,12 +750,7 @@ class Node(with_metaclass(NodeMeta, Drawable, Connectable, BaseNode)):
             A unique name for the node
         """
 
-        schematic_position = kwargs.pop("schematic_position", None)
-        geographic_position = kwargs.pop("geographic_position", None)
-        if schematic_position is not None:
-            schematic_position = tuple(schematic_position)
-        if geographic_position is not None:
-            geographic_position = tuple(geographic_position)
+        position = kwargs.pop("position", {})
 
         color = kwargs.pop('color', 'black')
         min_flow = pop_kwarg_parameter(kwargs, 'min_flow', 0.0)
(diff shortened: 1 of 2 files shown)


pywr/core.py
```diff
@@ -790,21 +791,15 @@ def load(cls, data, model):
...
-        except KeyError:
-            try:
-                position = data.pop('position')
-                x, y = position
-                x = float(x)
-                y = float(y)
-            except KeyError:
-                x = None
-                y = None
+        schematic_position = data.pop("schematic_position", None)
```
jetuk: Shouldn't this be the following to agree with discussion in #154 ?

```
position_data = data.pop("position", {})
schematic_position = position_data.pop("schematic", None)
geographic_position = position_data.pop("geographic", None)
```

tests/test_core.py
```diff
@@ -136,6 +136,27 @@ def test_slots_connect_disconnect(solver):
...
 
+def test_node_position(solver):
+    model = Model(solver=solver)
+
+    node1 = Input(model, "input", schematic_position=(10, 20), geographic_position=(-1, 52))
+
+    data = {
+        "name": "output",
+        "type": "output",
+        "schematic_position": [30, 40],
```
jetuk: See above. I thought in #154 agreed the following JSON format,
```
{"position": {
    "schematic": [5, -20],
    "geographic": [-1.509445, 51.30284]
} }
```

jetuk:
Not sure this agrees with #154 discussion.

snorfalorpagus:
I think you're probably right. I'll make the changes.

snorfalorpagus:
I think this implements the position correctly. It's not very strict - you can pass whatever you want as the position dictionary.

PR #169, opened 2016-07-06
Speedup for GLPK solver

snorfalorpagus:
This PR includes a few speedups in the GLPK solver. Changes include:

- When calculating the cost of a route, all of the node costs are calculated and cached in advance.
- Added a placeholder on `AbstractNode` to store solver-specific data.
- Where feasible moved to `cvarray` instead of `list`.

I suspect additional speedups might be possible if we manage to translate the `routes` list into something more C-ish. However this is not especially easy as it's a list of list of Nodes, with heterogenous length.

This was benchmarked using a model with 143 nodes and 239 routes (the most complex to date!). The changes result in a 2x increase in speed!

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -2,11 +2,17 @@ from libc.stdlib cimport malloc, free
 
 cimport cython
 
+import numpy as np
+cimport numpy as np
+
 from pywr._core import BaseInput, BaseOutput, BaseLink
 from pywr._core cimport *
 from pywr.core import ModelStructureError
 import time
 include "glpk.pxi"
+
+cdef class AbstractNodeData:
+    cdef public int id
 
 cdef class CythonGLPKSolver:
     cdef glp_prob* prob
@@ -19,6 +25,9 @@ cdef class CythonGLPKSolver:
     cdef list routes
     cdef list non_storages
     cdef list storages
+    cdef list routes_cost
+    cdef list nodes_with_cost
+    cdef int num_nodes
     cdef public object stats
 
     def __cinit__(self):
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
The appveyor failure is just appveyor being useless again.

pywr/solvers/cython_glpk.pyx
```diff
@@ -186,28 +250,43 @@ cdef class CythonGLPKSolver:
...
 
-        # update route properties
+        # update route cost
+
+        t0 = time.clock()
+
+        # update the cost of each node in the model
+        cdef double[:] node_costs = self.node_costs_arr
+        for _node in self.nodes_with_cost:
+            data = _node.__data
```
jetuk: Isn't this slow because `_node.__data` is of type `object` and not `AbstractNodeData`?

pywr/solvers/cython_glpk.pyx
```diff
@@ -51,6 +66,24 @@ cdef class CythonGLPKSolver:
...
+        cdef int n
+
+        self.all_nodes = list(model.graph.nodes())
+        self.nodes_with_cost = []
+
+        n = 0
+        for _node in self.all_nodes:
+            _node.__data = AbstractNodeData()
+            _node.__data.id = n
+            if hasattr(_node, "get_cost"):
```
jetuk: Is this ever false?

jetuk:
I guess this is hard to do without caching the data in the nodes themselves? I think that because `__data` is `cdef object` it'll be slow to access. But that would require the `AbstractNodeData` being moved to `_core`. Is that worth doing? The lpsolve solver would presumably use the same approach?

pywr/solvers/cython_glpk.pyx
snorfalorpagus: Aggregated nodes don't have a cost.

pywr/solvers/cython_glpk.pyx
snorfalorpagus: I wanted to keep this agnostic to the solver. Try changing it and see if it makes a difference?

snorfalorpagus: I've done some benchmarking and it seems this doesn't make a difference.

Issue #158, opened 2016-06-27
Additional LP constraints


snorfalorpagus:
It would be useful to be able to constrain a flow in a node (or group of nodes) as either GT/LT/EQ to another node (or group of nodes) multipled by a factor. This is easy to do in the linear program. The question is how do we expose it in the API?

We talked about using the new aggregated node class to apply a constant constraint to a group of nodes. Psuedocode:

```
agg_node = AggregatedNode([node1, node2])
agg_node.max_flow = 10.0
```

I think this is a nice way of doing things. The GT/LT/EQ constraints can be done using `min_flow`, `max_flow` and `flow` as above. However I'm not sure how this extends to a non-constant constraint as described in the first paragraph.

I can think of two immediate uses for this. One is to re-implement the RiverSplit functionality we used to have. The other is for discharges that are proportional to an abstraction (multiplied by some factor). A compound node could be created for the RiverSplit which would allow a constant for the equation `y=mx+c`.

Thoughts @jetuk ?

snorfalorpagus:
We talked about implementing the river split like this:

![split](https://cloud.githubusercontent.com/assets/3883947/16528095/8ba1653e-3fb2-11e6-9833-e15f8d418745.png)

PR #174, opened 2016-07-08
Constraints for AggregatedNode

snorfalorpagus:
This PR adds support for constraints on AggregatedNode, specifically max_flow, min_flow and a ratio between the member nodes.

This partly addresses #158.

Outstanding issues are that these parameters are not time varying, and that there isn't a default component that creates the structure in #158.

I think it's good enough for now though!

It also fixes a minor memory leak, and tweaks the way tests are run in the conda recipe.


snorfalorpagus:
Still need to implement JSON load for AggregatedStorage...

snorfalorpagus:
The most recent commit adds support for AggregatedStorage nodes in JSON, and also allows them to be used by control curves. This needs a test.

I'm done for the weekend now.

snorfalorpagus:
The most recent commit creates the system illustrated in #158, using an `AggregatedNode` and a `PiecewiseLink`. The final thing to do on this is to simplify the creation of this setup with a class that does it all for you.

tests/test_agg_constraints.py
```diff
@@ -145,11 +145,49 @@ def test_aggregated_node_max_flow_same_route(model):
...
+    y = mx + c, where y is the MRF, x is the upstream flow and m and c are
+    constants.
+
+    TODO: build a standard class which simplifies the construction of this setup
+
+             / -->-- X0 -->-- \
+    A -->-- Xo -->-- X1 -->-- Xi -->-- C
+             \ -->-- X2 -->-- /
+                     |
+                     Bo -->-- Bi --> D
```
jetuk: Why can't X2 connect straight to D?

snorfalorpagus: Domains.

jetuk: Who's idea were they? Jeesh

snorfalorpagus:
I think this is ready to merge. The application of the aggregated node to do the Y=MX+C MRF and RiverSplit can come in a later PR.

snorfalorpagus:
Merged.

PR #180, opened 2016-07-15
Add convenience classes for splits

### LINKED ISSUE #158: Additional LP constraints

jetuk:
Fixes #158

- Adds MultiSplitLink as the most general way of extending PiecewiseLink with additional downstream slots.
- Adds RiverSplit and RiverSplitWithGauge as convenience initialisers for the most common cases.


jetuk:
This now also implements the `AggregatedNode` constraints in lpsolve.

snorfalorpagus:
Need to remove the old test.

https://github.com/pywr/pywr/blob/e4fe517b1796bd4d225e275a9c9a361490f7ba41/tests/test_run.py#L94-L102

pywr/core.py
```diff
@@ -1155,6 +1155,124 @@ def after(self, timestep):
...
+            nodes = []
+            valid_ratios = []
+            for r, nd in zip(ratios, self.sublinks[n:]):
+                if r is not None:
+                    nodes.append(nd)
+                    valid_ratios.append(r)
+
+            remaining_ratio = 1.0 - sum(valid_ratios)
+            if remaining_ratio < 0.0:
+                raise ValueError('Sum of ratios is more than 1.0.')
```
snorfalorpagus: Could we normalise the ratios instead? Some ratios aren't very nice to specify like this, e.g. 1/3rd would have to be `(0.3333333333, 0.666666666)`. Obviously in Python you could just pass `(1/3, 2/3)`, but this isn't possible in JSON. Instead it would be nicer to pass `(1, 2)`.

tests/test_agg_constraints.py
```diff
@@ -145,21 +141,25 @@ def test_aggregated_node_max_flow_same_route(model):
...
+def test_piecewise_constraint(model, flow):
     """Test using an aggregated node constraint in combination with a
     piecewise link in order to create a minimum flow constraint of the form
     y = mx + c, where y is the MRF, x is the upstream flow and m and c are
     constants.
 
+    Flows are tested at 100, 200 and 300 to ensure the aggregated ratio works
+    when there is too much to route entirely through to node 'D'.
+
     TODO: build a standard class which simplifies the construction of this setup
```
snorfalorpagus: Done! Remove this TODO.

snorfalorpagus:
Needs JSON support.

pywr/solvers/cython_lpsolve.pyx
```diff
@@ -274,11 +286,80 @@ cdef class CythonLPSolveSolver:
...
+                    if node in route:
+                        matrix.add(n)
+            length = len(matrix)
+            ind = <int*>malloc(length * sizeof(int))
+            val = <double*>malloc(length * sizeof(double))
+            for i, col in enumerate(matrix):
+                ind[i] = 1+col
+                val[i] = 1.0
+            #glp_set_mat_row(self.prob, row, length, ind, val)
+            #glp_set_row_bnds(self.prob, row, constraint_type(min_flow, max_flow), min_flow, max_flow)
```
snorfalorpagus: Remove these.

snorfalorpagus:
Fix that last issue (the comments in lpsolve) and I'm happy for this to be merged.

jetuk:
Done and merged. Thank you.

PR #181, opened 2016-07-15
Add ParameterRecorder and IndexParameterRecorder.

jetuk:
This commit adds Recorder subclasses for Parameter and IndexParameter. Currently there are NumpyArray... versions of these base classes implemented and tested. Parameters now have access to the _recorders list to find out which Recorder objects are looking at them.

Spotted a couple of bugs on ControlCurveIndexParameter as well that are fixed.

TODO
- Currently there is no way to give a Parameter a name via keyword arguments to __init__.
- More subclasses to use the new Recorders.

### DIFF
--- a/pywr/_recorders.pyx
+++ b/pywr/_recorders.pyx
@@ -95,12 +95,56 @@ cdef class StorageRecorder(Recorder):
         self._node = node
         node._recorders.append(self)
 
+    property node:
+        def __get__(self):
+            return self._node
+
     def __repr__(self):
         return '<{} on {} "{}" ({})>'.format(self.__class__.__name__, repr(self.node), self.name, hex(id(self)))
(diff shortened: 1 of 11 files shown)


pywr/_recorders.pyx
```diff
@@ -103,6 +107,46 @@ cdef class StorageRecorder(Recorder):
...
+        self._param = param
+        param._recorders.append(self)
+
+    property parameter:
+        def __get__(self):
+            return self._param
+
+    def __repr__(self):
+        return '<{} on {} "{}" ({})>'.format(self.__class__.__name__, repr(self.parameter), self.name, hex(id(self)))
+
```
snorfalorpagus: Why is `__repr__` defined twice?

Issue #155, opened 2016-06-24
Additional constraint parameters


snorfalorpagus:
Following on from #143, there are a couple of constraints/patterns we don't have a standard parameter for.

1. Parameters with a timestep-counting constraint. e.g. `max_flow` is dependent on the number of days since a node had max_flow > 0, or number of days since a demand saving level was breached.
2. Parameter dependent on the average flow over the last x timesteps, e.g. an option is only enabled if average flow for the last week is < X.

snorfalorpagus:
@jetuk What do you think of this approach?

https://gist.github.com/snorfalorpagus/97cf8877c3094b124aa6a3b07b0ca9ce

jetuk:
I like this. It's blurring the lines between Recorder and Parameter. Does it imply that the license classes are implemented incorrectly? I.e. the Parameters track the state there rather than Recorders

snorfalorpagus:
I think the first parameter can be implemented in a similar fashion: a counting recorder (+1 each timestep a condition is met) and a parameter which divides that count by a constant and uses it as an index. The easiest way to implement the counting recorder is to have it as a meta-recorder (it's value is based on another recorder). However, I think this might require the dependency tree we talked about a little while ago for updating parameters/recorders.

snorfalorpagus:
The `MeanFlowRecorder` implements the second feature.

The first feature turned out to be more involved / too specific to the model.

PR #183, opened 2016-07-18
Added MeanFlowRecorder and RecorderThresholdParameter.

snorfalorpagus:
This PR adds a `MeanFlowRecorder` which records the mean flow of a node for the past N timesteps, and `RecorderThresholdParameter` which returns one of two values depending on the value of a recorder vs a threshold.

This required a refactor of `load_recorder` to be more in line with `load_parameter`, as recorders can now be a dependency of parameters.

This implements the second point in #155.


jetuk:
Is this ready to review now?

snorfalorpagus:
Yes.

PR #184, opened 2016-07-18
Initial work on a VirtualStorage class.

jetuk:
Currently enforces the volume constraint in the solver.

Still needs to influence the costs of routes that are connected to it.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -273,7 +273,7 @@ cdef class AbstractNode:
             self._flow[i] += value[i]
 
     cpdef after(self, Timestep ts):
-        pass
+        self._prev_flow[:] = self._flow[:]
 
     cpdef check(self,):
         pass
@@ -459,7 +459,7 @@ cdef class Node(AbstractNode):
     cpdef after(self, Timestep ts):
         """Called at the end of the timestep"""
         AbstractNode.after(self, ts)
-        self._prev_flow[:] = self._flow[:]
+
         # Complete any parameter calculations
         if self._cost_param is not None:
             self._cost_param.after(ts)
@@ -501,7 +501,7 @@ cdef class AggregatedNode(AbstractNode):
             self.model.dirty = True
 
     cpdef after(self, Timestep ts):
-        AbstractStorage.after(self, ts)
+        AbstractNode.after(self, ts)
         cdef int i
         cdef Node n
 
@@ -857,3 +857,25 @@ cdef class AggregatedStorage(AbstractStorage):
         nodes = [model._get_node_from_ref(model, node_name) for node_name in data["storage_nodes"]]
         agg = cls(model, name, nodes)
         return agg
+
+
+cdef class VirtualStorage(Storage):
+
+    property nodes:
+        def __get__(self):
+            return self._nodes
+
+        def __set__(self, value):
+            self._nodes = list(value)
+            self.model.dirty = True
+
+    cpdef after(self, Timestep ts):
+        cdef int i
+        cdef ScenarioIndex si
+        cdef AbstractNode n
+
+        for i, si in enumerate(self.model.scenarios.combinations):
+            self._flow[i] = 0.0
+            for n in self._nodes:
+                self._flow[i] -= n._flow[i]
+        Storage.after(self, ts)
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -860,6 +860,8 @@ cdef class AggregatedStorage(AbstractStorage):
 
 
 cdef class VirtualStorage(Storage):
+    def __cinit__(self, ):
+        self.virtual = True
 
     property nodes:
         def __get__(self):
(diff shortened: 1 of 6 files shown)


jetuk:
I should say, the reason for doing this is to implement `Parameters` that have some sort of storage element to them. We have the current `Licence` classes but they effectively emulate a crude storage in Python space, so we can acknowledge this should be a core part of the LP.

It raised a different question about licences also. Should they be a hard or soft constraint on flow?

pywr/solvers/cython_glpk.pyx
```diff
@@ -211,8 +218,27 @@ cdef class CythonGLPKSolver:
...
+                ind[1+n] = self.idx_col_routes+c
+                val[1+n] = -num
+
+            glp_set_mat_row(self.prob, self.idx_row_virtual_storages+col, len(cntr), ind, val)
+            free(ind)
+            free(val)
+
         # aggregated node flow ratio constraints
-        self.idx_row_aggregated = self.idx_row_storages + len(storages)
+        self.idx_row_aggregated = self.idx_row_storages + len(virtual_storages)
```
snorfalorpagus: This should be:

```
self.idx_row_aggregated = self.idx_row_virtual_storages + len(virtual_storages)
```

Alternatively, we could set `self.idx_row_aggregated = None`, then in the loop when the first row is created (`if ... is None`) assign it the value returned by `glp_add_rows`.

snorfalorpagus:
Licences should be a hard constraint. A node should not be allowed to breach it's licence.

The complexity arrises from annual (or seasonal/rolling) licences where the equivalent daily average (= volume / 365) is less than the daily licence. At this point you need to manage it as a finite resource. The way this is done will depend on the specifics of the system. Some sources will be operated at their daily equivalent rate all year long, while others are really just operate during the peak period.

I think in principle I'm not opposed to moving the constraint from Python-space to the LP.

jetuk:
If we implement licences with a virtual storage then they won't be able to breach the conditions. However we will also want their cost to increase if they are relatively overused. Is the intention then that a licence parameter creates its own `VirtualStorage` to limit the total flow, but is designed to be assigned to a cost attribute?

jetuk:
I'm going to rebase this PR since #190 merged.

Seems like all the licence parameters will need rewriting to use it.

jetuk:
So this is working a bit better now. The tests fail only on lpsolve. I have a half written `RecessionCurve` class to complete.

pywr/_core.pyx
```diff
@@ -500,7 +500,7 @@ cdef class AggregatedNode(AbstractNode):
             self.model.dirty = True
 
     cpdef after(self, Timestep ts):
-        AbstractStorage.after(self, ts)
+        AbstractNode.after(self, ts)
```
snorfalorpagus: Ouch. Nasty bug. I wonder how this went unnoticed?

jetuk: I think all the work was in `Storage.after` rather than `AbstractStorage.after`.

pywr/solvers/cython_glpk.pyx
```diff
@@ -211,8 +216,37 @@ cdef class CythonGLPKSolver:
             free(ind)
             free(val)
 
+        # virtual storage
+        if len(virtual_storages):
+            self.idx_row_virtual_storages = glp_add_rows(self.prob, len(virtual_storages))
+        for col, storage in enumerate(virtual_storages):
+            # We need to handle the same route appearing twice here.
```
snorfalorpagus: Is this a TODO?

pywr/solvers/cython_glpk.pyx
```diff
@@ -211,8 +216,37 @@ cdef class CythonGLPKSolver:
...
+            cols = {}
```
snorfalorpagus: I've avoided dictionaries elsewhere in the solver. Is this really needed?

pywr/solvers/cython_glpk.pyx
jetuk: No, it works with that.

pywr/solvers/cython_glpk.pyx
jetuk: It is needed to handle the multiple routes I think.

jetuk:
What's left to add to this before merge? The licence refactor?

snorfalorpagus:
Yes. Will this work OK with rolling licences? I guess the volume would need to be reset each timestep?

snorfalorpagus:
Support for lpsolve is done. Need to add JSON support, and some documentation (esp. how the factors work).

snorfalorpagus:
The virtual storage class has a "cost" parameter that I don't think is respected. I think this is just a carry-over from storage and should be removed? @jetuk

jetuk:
That's true about cost. TBH the  hierarchy for the `Storage` classes probably needs a more fundamental think through. I.e. which should have which parameter.

snorfalorpagus:
Let's talk about this tomorrow. I can see some sense in allowing a virtual storage to have a cost.

PR #185, opened 2016-07-19
Faster calculation of day of year in _core.Timestep

snorfalorpagus:
This change has a significant impact on the speed of my large (150 node, 250 edges) model. It goes from 1250 timesteps/second to 1900 timesteps/second - a 50% speedup for 3 lines of code!


snorfalorpagus:
The second commit optimises `AggregateParameter` and `AggregateIndexParameter` which were the next slowest parameters on the list. My model is now at 2200 timesteps/second.

PR #186, opened 2016-07-19
Added comment attribute to Nodes, Parameters and Recorders

### LINKED ISSUE #144: Comments in JSON
Add support for adding comments in JSON. These need to be preserved between save/load, so we'll need to add a "comments" property to nodes/parameters/etc.

snorfalorpagus:
This can be set directly (node.comment = "hello"), or using
a keyword argument (Node(model, name, comment="hello"), or via
JSON.

Closes #144.

### DIFF
--- a/pywr/_recorders.pyx
+++ b/pywr/_recorders.pyx
@@ -8,11 +8,12 @@ import pandas as pd
 recorder_registry = set()
 
 cdef class Recorder:
-    def __init__(self, model, name=None):
+    def __init__(self, model, name=None, comment=None):
         self._model = model
         if name is None:
             name = self.__class__.__name__.lower()
         self.name = name
+        self.comment = comment
         model.recorders.append(self)
 
     property name:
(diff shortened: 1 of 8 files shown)


pywr/_recorders.pyx
```diff
@@ -122,10 +123,10 @@ recorder_registry.add(ParameterRecorder)
 
 
 cdef class IndexParameterRecorder(Recorder):
-    def __init__(self, model, IndexParameter param, name=None):
+    def __init__(self, model, IndexParameter param, name=None, **kwargs):
         if name is None:
             name = "{}.{}".format(self.__class__.__name__.lower(), param.name)
-        Recorder.__init__(self, model, name=name)
+        Recorder.__init__(self, model, name=name, **kwargs)
```
jetuk: `super`?

jetuk:
Looks good except for that one change to using `super`. Not sure it matters but looks like we're converging to `super` everywhere.

PR #209, opened 2016-08-12
Recorder value refactor

jetuk:
Here's the work I've done to `Recorders` recently. 

This currently hangs off the virtual storage branch due to my disorganisation.


pywr/_recorders.pyx
```diff
@@ -1,20 +1,44 @@
...
 cimport numpy as np
-
 from pywr._core cimport Timestep
-
 import pandas as pd
+from past.builtins import basestring
 
 recorder_registry = set()
 
+cdef enum AggFuncs:
```
snorfalorpagus: We should move this (and the one in `_parameters`) into `_core`.

jetuk:
Should I rebase this to master without the virtual storage stuff?

snorfalorpagus:
Yes please. Probably easiest to cherry pick the commits:

```
git checkout master
git checkout -b new_branch
git cherry-pick xxxxxxx
git cherry-pick xxxxxxx
git cherry-pick xxxxxxx
```

Where xxx are the commit IDs you want.

pywr/_recorders.pyx
jetuk: I think I hit this trying to do this ...

https://github.com/cython/cython/issues/531

https://groups.google.com/forum/#!msg/cython-users/ZoLsLHwnUY4/74_JZaZvAAAJ

snorfalorpagus: I don't think the enums need to be cpdef. They're just used so the compiler can optimise the switch/case statement. The end-user is expected to use strings (e.g. "sum").

snorfalorpagus:
I think for now I'd be OK with you just keeping `AggFuncs` in `_recorders` (with a duplicate in `_parameter`) and renaming the constants to use upper case to avoid collision with builtin names.

jetuk:
I made the constant changes and merged this.

Issue #86, opened 2016-01-21
Code coverage


snorfalorpagus:
Configure code coverage using (something like) Coveralls: https://coveralls.io/

snorfalorpagus:
Progress in this branch: https://github.com/snorfalorpagus/pywr/commits/coveralls

SO question: http://stackoverflow.com/questions/34935796/how-to-use-coverage-analysis-with-cython

snorfalorpagus:
It looks like this may not be possible. From what I can tell the support for cdef classes (which is most of our cython code...) is lacking in coverage.py.

snorfalorpagus:
Cython.org is down at the momement so I've posted something on cython-users:

https://groups.google.com/forum/#!topic/cython-users/e80vdI9G_3w

snorfalorpagus:
I have decided that some coverage analysis (even if partially broken for cython) is better than nothing. I'll try and address this soon.

PR #214, opened 2016-08-17
Added coverage support

snorfalorpagus:
Some support for coverage is better than none, even if it's not working 100% correctly for Cython.

Build cython modules with tracing enabled:

```
python setup.py --with-glpk --with-lpsolve --enable-tracing develop
```

Then run coverage on tests
```
coverage run --source pywr py.test tests
coverage report
coverage html
```

#86

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+plugins = Cython.Coverage
+omit = *.pxd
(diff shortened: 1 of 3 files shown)


jetuk:
Does this enable CI based coverage too?

snorfalorpagus:
Not yet.

jetuk:
Merge without that?

Issue #237, opened 2016-08-26
Load all parameters, even if they aren't referenced


snorfalorpagus:
Parameters are only loaded from the parameters section of the JSON document if they are referenced somewhere else (e.g. by a node property, or a recorder).

https://github.com/pywr/pywr/blob/master/pywr/core.py#L437-L460

This seemed sensible to me when I was writing it - why load something you aren't going to use? I overlooked the use-case where you are defining a parameter in JSON for later use by something from Python space (e.g. a control curve which is easy to define in JSON, but an operating rule which extends the basic Pywr functionality).

PR #238, opened 2016-08-26
WIP: Load all parameters

snorfalorpagus:
This PR addresses #237. Parameters are loaded, even if they aren't referenced anywhere in the JSON.

Also fixes a bug from the switch from recorders being a list to a dictionary.

I need to add a test for this that involves something a little more convoluted: a parameter that references a recorder that references a parameter that references a recorder... or something like that. The fact the current tests pass isn't enough.

### DIFF
--- a/pywr/core.py
+++ b/pywr/core.py
@@ -435,9 +435,6 @@ class Model(object):
         model._nodes_to_load = nodes_to_load
 
         # collect parameters to load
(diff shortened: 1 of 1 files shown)


jetuk:
OK, should I wait for that last bit before merge?

snorfalorpagus:
I don't think there is a way to create this setup currently. #240.

PR #250, opened 2016-09-01
Renamed ConstantParameter `values` to `value` in JSON

snorfalorpagus:
This PR renames the "values" argument to `ConstantParameter` in JSON to "value". This makes more sense to me, given that the value is singular. Previously the name was inherited from `Parameter`.

This will break a lot of existing models, although the fix is trivial. I don't know if we need to start thinking about version numbers...

An altenative approach is to support both forms ("values" AND "value").

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -153,6 +153,13 @@ cdef class ConstantParameter(Parameter):
 
     cpdef double[:] upper_bounds(self):
         return self._upper_bounds
+
+    @classmethod
+    def load(cls, model, data):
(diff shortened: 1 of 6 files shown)


jetuk:
Is there a problem supporting both "values" and "value"?

snorfalorpagus:
> Is there a problem supporting both "values" and "value"?

Not a technical one, no.

snorfalorpagus:
#250.

jetuk:
Should we support both then, or not?

snorfalorpagus:
Let's go with both. I'll make the change.

PR #255, opened 2016-09-16
Added minimum_version to JSON document metadata.

snorfalorpagus:
This PR adds a "minimum_version" property to JSON documents. If the version specified is greater than the installed version of Pywr, a warning is raised.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
     - conda update --yes conda
 
     # dependencies
-    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy matplotlib dateutil pandas statsmodels networkx xlrd cython six ipython pytables future
+    - conda install --yes python=$TRAVIS_PYTHON_VERSION atlas numpy scipy matplotlib dateutil pandas statsmodels networkx xlrd cython six ipython pytables future packaging
     # pytest from conda is too old, see https://github.com/pytest-dev/pytest/issues/995
     - pip install pytest==2.8.5
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
     matrix:
         - CONDA_NPY=111  CONDA_PY=27
         - CONDA_NPY=111  CONDA_PY=34
+        - CONDA_NPY=111  CONDA_PY=35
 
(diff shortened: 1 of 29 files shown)


snorfalorpagus:
In addition to the initial purpose of this branch, I've transitioned to using Travis' "container-based" builds. The Travis build is now done using the same system as Appveyor, i.e. using the conda build system. The result is a more standardised setup - you only have to specify new dependencies in `conda-recipe/meta.yaml` instead of in `.travis.yml` manually. Also, the build times on Travis are significantly reduced (15 minutes down to 3 minutes).

Need to squash some of these commits before merging...

jetuk:
Is this ready now?

snorfalorpagus:
Yes.

.travis.yml
```diff
@@ -1,40 +1,30 @@
...
-    - "2.7"
-    - "3.4"
+sudo: false
 
 env:
-    - "PYWR_SOLVER=glpk"
-    - "PYWR_SOLVER=lpsolve"
+    matrix:
+        - CONDA_NPY=111  CONDA_PY=27
+        - CONDA_NPY=111  CONDA_PY=34
```
jetuk: Should we add 3.5?

pywr/core.py
```diff
@@ -428,6 +430,17 @@ def load(cls, data, model=None, path=None, solver=None):
         # data is a dictionary, make a copy to avoid modify the input
         data = copy.deepcopy(data)
 
+        # check minimum version
+        try:
+            minimum_version = data["metadata"]["minimum_version"]
+        except KeyError:
+            pass
```
jetuk: Warning for no minimum version?

.travis.yml
snorfalorpagus: I don't see why not. I don't see any benefit in testing other numpy versions though.

pywr/core.py
snorfalorpagus: That should encourage users to add it. We'll have to modify all the test models.

.travis.yml
jetuk: Is this something we'll forget to update? How do you always use the latest version?

jetuk:
I'll merge once it passes CI

PR #261, opened 2016-09-26
Position of nodes in d3 graph will be fixed after dragging.

snorfalorpagus:
This PR tweaks the graph used in jupyter so that if a node is dragged to a position, it becomes "sticky" (i.e. it will stay where it's dragged to). Double-clicking the node will make the node behave normally again.

### DIFF
--- a/pywr/notebook/__init__.py
+++ b/pywr/notebook/__init__.py
@@ -43,7 +43,8 @@ def nx_to_json(model):
     for n, name in enumerate(nodes):
         node = {"name": name}
         classes = []
-        cls = model.node[name].__class__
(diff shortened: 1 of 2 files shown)


jetuk:
Is it possible to seed the positions from the beginning? Or is that a nightmare with coordinate transforms?

snorfalorpagus:
> Is it possible to seed the positions from the beginning? Or is that a nightmare with coordinate transforms?

If a node has `.position["schematic"]` set then it's used and the node is sticky from the beginning.

The geographic position isn't used - I think it will be quite a bit of work to get it displayed over a map.

PR #262, opened 2016-09-26
Added an index argument when loading a DataFrame

jetuk:
Useful for providing constants from a table by looking up column and row.


snorfalorpagus:
Nice idea. Can you:

- Test the values are loaded correctly (currently only testing a ConstantParameter is returned)
- Add a couple more rows to the example data, so that we're testing the correct row is loaded.

jetuk:
Done.

Also added support for multi-indexing by passing a list to either `column` or `index`.

PR #264, opened 2016-09-27
Fixed bug in MeanFlowRecorder related to days->timesteps

snorfalorpagus:
Fixed a bug in MeanFlowRecorder in the conversion of days to timesteps.

Need to add a test for this to prevent regressions.


snorfalorpagus:
There was already a test, it just wasn't enough. This is ready for review.

Issue #159, opened 2016-06-28
Should registry be a class method?


snorfalorpagus:
When creating a new parameter/node/etc the API looks like this:

```
parameter_registry.add(ControlCurveIndexParameter)
```

How about this instead:

```
ControlCurveIndexParameter.register()
```

Presumably there would be an equivalent `deregister` or `unregister` function.

jetuk:
What's the benefit of this? Just style?

snorfalorpagus:
Yes.

snorfalorpagus:
This was implemented in #265.

It could also be done for recorders...

PR #265, opened 2016-09-27
parameter_registry is now a dict

### LINKED ISSUE #215: Redefining parameter types causes problems for parameter_registry
When interactively developing a new parameter type (in Jupyter) we can end up with many different versions of the parameter stored in `parameter_registry`. Should `parameter_registry` be a dict? I can't remember why it's a set.

Same applies for other registries.

#159.

snorfalorpagus:
This PR changes `parameter_registry` from a set to a dict in order to close #215.

Additionally, a `Parameter.register()` method is added to manage the registry, closing #159.

### DIFF
--- a/pywr/parameters/_control_curves.pyx
+++ b/pywr/parameters/_control_curves.pyx
@@ -72,7 +72,7 @@ cdef class BaseControlCurveParameter(Parameter):
         return node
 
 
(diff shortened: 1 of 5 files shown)


snorfalorpagus:
This is a significant change to the API - let's put it in 0.2.

PR #271, opened 2016-10-02
Improvements to TablesRecorder

### LINKED ISSUE #268: TablesRecorder can't handle AggregatedNode
When attaching a `TablesRecorder` to a model without passing it any nodes, it tries to record every node. It can't handle `AggregatedNode`:

```
ValueError: Unrecognised Node type '<class 'pywr.core.AggregatedNode'>' for TablesRecorder
```

jetuk:
Fixes #268

`TablesRecorder` is now a lot more general.

- It can record multiple `Nodes` and `Parameters`.
- User can specify the path within the h5 file they would like each array saving.
- It can be defined from JSON with filters and mode arguments.

### DIFF
--- a/pywr/recorders.py
+++ b/pywr/recorders.py
@@ -87,8 +87,9 @@ class TablesRecorder(Recorder):
         ----------
         model : pywr.core.Model
             The model to record nodes from.
-        h5file : tables.File
-            The tables file handle to attach the CArray objects to.
+        h5file : tables.File or filename
+            The tables file handle or filename to attach the CArray objects to. If a
+            filename is given the object will open and close the file handles.
         nodes : iterable or None
             Nodes to save in the tables database. Can be an iterable of Node objects or
             node names. It can also be a iterable of tuples with a node specific where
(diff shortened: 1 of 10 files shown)


snorfalorpagus:
This looks good. I'll review properly tomorrow.

pywr/recorders.py
```diff
@@ -80,59 +80,156 @@ class TablesRecorder(Recorder):
...
         import tables
         shape = len(self.model.timestepper), len(self.model.scenarios.combinations)
-        atom = tables.Float64Atom()
         # Create a CArray for each node
         self._arrays = {}
 
         # Default to all nodes if None given.
         if self.nodes is None:
-            nodes = self.model.node.values()
+            nodes = [(self.where, n) for n in self.model.node.values()]
```
snorfalorpagus: Please use `self.model.nodes` instead of `self.model.node`. (check for other instances)

pywr/recorders.py
```diff
@@ -80,59 +80,156 @@ class TablesRecorder(Recorder):
...
+                try:
+                    where, node = n
+                except (TypeError, ValueError):
+                    node = n
+                    where = self.where
+
+                # Accept a str, and lookup node by name instead.
+                if isinstance(node, basestring):
+                    node = self.model.nodes[node]
+                # Otherwise assume it is a node object anyway
```
snorfalorpagus: Check this?

pywr/recorders.py
snorfalorpagus: #272

pywr/recorders.py
snorfalorpagus: Actually I don't think this is needed.

Issue #270, opened 2016-09-30
Scenarios in JSON


snorfalorpagus:
We don't currently have a way to represent scenarios in JSON.

PR #274, opened 2016-10-03
Loading scenarios and TablesArrayParameter

### LINKED ISSUE #269: Add tables.Array based parameter
Support reading array style data from a pytables array node. This may be more memory efficient when running a large number of scenarios as the data will be read progressively from disk rather than kept entirely in memory.

jetuk:
This PR fixes #269 and #270.

- Scenarios can be defined in JSON
- Per scenario data can be read from a `pandas.DataFrame` using `DataFrameParameter` and passing a `Scenario` object.
- Above can be loaded from JSON using "scenario" key in the parameter defintion.
- New `TablesArrayParameter` that can read array data using PyTables. Key advantage is that this does not load all the data in to memory, but reads it sequentially in `value()`
- Fixed a couple of bugs with `Scenario`
  - Checking `size` >= 1
  - Making the `_name`  basestring for better Python 2 support when loading from JSON (i.e. unicode).


tests/test_scenarios.py
```diff
@@ -142,3 +143,57 @@ def test_scenario_storage(solver):
     assert_allclose(i.flow, [5, 10])
     assert_allclose(s_rec.data[0], [505, 510])
     assert_allclose(s_rec.data[1], [510, 520])
+
+
+def test_scenarios_from_json(solver):
+
+    model = load_model('simple_with_scenario.json', solver=solver)
+s
```
snorfalorpagus: Typo

snorfalorpagus:
Need to rebase this on master and make use of the code in `pywr.h5tools`.

pywr/_core.pyx
```diff
@@ -820,6 +832,18 @@ cdef class Storage(AbstractStorage):
             except ZeroDivisionError:
                 self._current_pc[i] = np.nan
 
+    cpdef finish(self):
+        """Called at the beginning of a run"""
```
jetuk: Docstring is wrong.

PR #275, opened 2016-10-03
Moved file-opening into reset method for CSV and H5 recorders.

snorfalorpagus:
This PR moves the code that opens the file recorders write to into the `reset` method.

A new class, `pywr.h5tools.H5Store`, is used to manage the open HDF5 file handle.

### DIFF
--- a/pywr/h5tools.py
+++ b/pywr/h5tools.py
@@ -3,7 +3,7 @@ import sys
 from past.builtins import basestring
 
 class H5Store(object):
-    def __init__(self, filename, filter_kwds, mode):
+    def __init__(self, filename, filter_kwds, mode, title=''):
         filter_kwds = filter_kwds
         mode = mode
         self._opened = False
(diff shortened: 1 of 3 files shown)


pywr/h5tools.py
```diff
@@ -0,0 +1,32 @@
...
+        if isinstance(filename, basestring):
+            # filename is a path to open
+            self.filename = filename
+            # Note sure how else to deal with str / unicode requirements in pytables
+            # See this issue: https://github.com/PyTables/PyTables/issues/522
+            import sys
+            if sys.version_info[0] == 2 and 'complib' in filter_kwds:
+                filter_kwds['complib'] = filter_kwds['complib'].encode()
+            filters = tables.Filters(**filter_kwds)
+            self.file = tables.open_file(filename, mode=mode, filters=filters)
```
jetuk: There are other keywords we can add to this. `title` and `description` I think.

pywr/recorders.py
```diff
@@ -203,12 +182,23 @@ def setup(self):
...
                 atom = tables.Int32Atom()
             else:
                 atom = tables.Float64Atom()
-            self._arrays[node] = self.h5file.create_carray(where, node.name, atom, shape, createparents=True)
+            self.h5store.file.create_carray(where, node.name, atom, shape, createparents=True)
+
+        self.h5store = None
+
+    def reset(self):
+        mode = "a" # always need to append, as file already created in setup
```
jetuk: Should be "r+" to ensure the file already exists.

pywr/h5tools.py
jetuk: It's just `title` actually.

PR #279, opened 2016-10-03
Add test for Jupyter notebooks

snorfalorpagus:
This PR adds a basic test for drawing a network diagram in d3 from Jupyter.

Also fixes a regression introduced in the notebook module (related to `model.node`).

This adds quite a few dependencies to the pywr package - is this something we're OK with? I don't see any need to install jupyter on the cluster...

Build is failing on appveyor for no reason...

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -2,5 +2,5 @@ set LIB=%LIBRARY_LIB%;.\lib;%LIB%
 set LIBPATH=%LIBRARY_LIB%;.\lib;%LIBPATH%
 set INCLUDE=%LIBRARY_INC%;%INCLUDE%
 
-"%PYTHON%" setup.py build_ext -I"%LIBRARY_INC%" -L"%LIBRARY_LIB%" install --with-glpk --with-lpsolve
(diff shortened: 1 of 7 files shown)


jetuk:
This is ready for review and merge despite the appveyor failure?

snorfalorpagus:
I believe so. I've checked - the extra dependencies only add a minute on to the build time. In the future we may have more detailed tests that need Jupyter.

conda-recipe/bld.bat
```diff
@@ -2,5 +2,5 @@ set LIB=%LIBRARY_LIB%;.\lib;%LIB%
 set LIBPATH=%LIBRARY_LIB%;.\lib;%LIBPATH%
 set INCLUDE=%LIBRARY_INC%;%INCLUDE%
 
-"%PYTHON%" setup.py build_ext -I"%LIBRARY_INC%" -L"%LIBRARY_LIB%" install --with-glpk --with-lpsolve
+"%PYTHON%" setup.py build_ext -I"%LIBRARY_INC%" -L"%LIBRARY_LIB%" install --with-glpk --with-lpsolve --single-version-externally-managed --record=record.txt
```
jetuk: `record.txt` isn't an external that needs to be provided?

snorfalorpagus: No, it's an output (that isn't used). I just copy+pasted from a conda-forge build. It's something to do with conda's handling off eggs I think.

PR #280, opened 2016-10-04
Changed syntax for table recorder names

### LINKED ISSUE #278: Improvement to TableRecorder naming
Don't append (pywr) node name to (pytables) node. e.g.

```
["/storage/abc", "myreservoir"]
```

Should produce `"/storage/abc"` not `"/storage/abs/myreservoir"`.

snorfalorpagus:
Closes #278.

Need to document how this works.

If a list of nodes are passed the arrays are named: `/where/node_name`, where where is a constant.

If tuples of (where, node) are used the arrays are named: `/where`

### DIFF
--- a/pywr/recorders.py
+++ b/pywr/recorders.py
@@ -2,7 +2,7 @@ import sys
 from pywr._recorders import *
 from past.builtins import basestring
 from .h5tools import H5Store
-
+import os
 
 class CSVRecorder(Recorder):
     """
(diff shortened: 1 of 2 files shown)


pywr/recorders.py
```diff
@@ -196,7 +202,10 @@ def setup(self):
                 atom = tables.Int32Atom()
             else:
                 atom = tables.Float64Atom()
-            self.h5store.file.create_carray(where, node.name, atom, shape, createparents=True)
+            group_name, node_name = os.path.split(where)
```
jetuk: Is it correct to treat these as filesystem path? Will that cause any problems with `\\` or similar? Why not just `where.split('/', 1)`?

jetuk: `rsplit` then

PR #293, opened 2016-10-10
Added a optional time Table output to TablesRecorder.

jetuk:
As title.


pywr/recorders.py
```diff
@@ -224,6 +241,14 @@ def save(self):
         ts = self.model.timestepper.current
         idx = ts.index
 
+        if self._time_table is not None:
+            entry = self._time_table.row
+            entry['year'] = ts.datetime.year
```
snorfalorpagus: Cache datetime object locally?

PR #297, opened 2016-10-13
TablesArrayParameter raises IndexError in reset() if length of data is shorted than timeseries

### LINKED ISSUE #296: TablesArrayParameter doesn't check table length
`TablesArrayParameter` doesn't currently check the number of rows in the table it uses. This can result in an `IndexError`, potentially at the very end of a long run...

https://github.com/pywr/pywr/blob/master/pywr/parameters/parameters.py#L188-L202

snorfalorpagus:
This PR makes two changes:

- If the length of data in the h5 file passed to a TablesArrayParameter is shorter than the required run length (e.g. if you had a year of data, but tried to run two years) an `IndexError` is raised during `reset()`. Closes #296.
- If an exception is raises during `model.run()`, `model.finish()` is called before raising, which gives time for parameters etc to tear down (e.g. closing file handles). Without this change, the above change results in the h5 file being left open. It doesn't get closed until the test script finishes, at which point you get a warning.


snorfalorpagus:
Merged as a68cb4683ee7b4cd3a5bc73bd334611573c2ff2a.

PR #301, opened 2016-10-14
Fix247 & Added "tables" section to JSON

jetuk:
This allows `DataFrame` objects to be read once by providing them in a `"tables"` section of the JSON. Reference to these objects can be made in individual `Parameter` definitions by using a `"table"` key instead of the `"url"`. The `"index"` and `"column"` keys should stay with the `Parameter` definition but other keys passed as keywords to pandas read functions should be moved to the `"tables"` definition.

### DIFF
--- a/tests/models/simple_df_shared.json
+++ b/tests/models/simple_df_shared.json
@@ -1,0 +1,49 @@
+{
+    "metadata": {
+        "title": "Simple 1",
+        "description": "A very simple example.",
+        "minimum_version": "0.1"
+    },
+    "timestepper": {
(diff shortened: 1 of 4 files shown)


tests/models/simple_df_shared.json
```diff
@@ -0,0 +1,49 @@
+{
+    "metadata": {
+        "title": "Simple 1",
+        "description": "A very simple example.",
```
snorfalorpagus: Update this.

snorfalorpagus:
Merged.

PR #302, opened 2016-10-16
Performance improvements

jetuk:
Two performance improvements related to pytables.

- First is an easy win the `TablesRecorder`
- Second is cythoning the `TablesArrayParameter` this also reads the entire array on init. It's not ideal for indices due to have to type to `double` and cast back to `int`.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -5,6 +5,7 @@ import pandas
 from libc.math cimport cos, M_PI
 from libc.limits cimport INT_MIN, INT_MAX
 from past.builtins import basestring
+from pywr.h5tools import H5Store
 
 cdef enum Predicates:
     LT = 0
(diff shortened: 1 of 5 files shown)


snorfalorpagus:
Is there anything we can do to avoid the casting? Check the type of the node, hand have a `_values_double` and `_values_int` memoryview? Or does this complicate `value()` too much?

jetuk:
Use two classes.

snorfalorpagus:
> Use two classes.

Instead we could have `value()` read the doubles array and `index()` read the integer array. If the respective array is `None` then cast the value from the other array.

pywr/recorders.py
```diff
@@ -248,23 +248,29 @@ def save(self):
...
             if isinstance(node, AbstractStorage):
                 ca[idx, :] = node.volume
             elif isinstance(node, AbstractNode):
                 ca[idx, :] = node.flow
             elif isinstance(node, IndexParameter):
+                a = np.empty(len(self.model.scenarios.combinations), dtype=np.int)
                 for si in self.model.scenarios.combinations:
-                    ca[idx, si.global_id] = node.index(ts, si)
+                     a[si.global_id] = node.index(ts, si)
+                ca[idx, :] = a
```
jetuk: Each call to `ca[idx, si.global_id]` creates lots of overhead in pytables. Doing it once saves this.

jetuk:
That sounds a much better solution.

snorfalorpagus:
I've made the change. There are existing tests that use both the `value()` and `index()` method of `TablesArrayParameter`, so I think this is good to go (once you review my changes).

pywr/parameters/_parameters.pyx
```diff
@@ -261,6 +262,111 @@ cdef class ArrayIndexedScenarioParameter(Parameter):
...
+    cpdef reset(self):
+        self.h5store = H5Store(self.h5file, None, "r")
+        node = self.h5store.file.get_node(self.where, self.node)
+
+        # detect data type and read into memoryview
+        if node.dtype == np.float32 or node.dtype == np.float64:
+            self._values_dbl = node.read()
+            self._values_int = None
+            shape = self._values_dbl.shape
+        else:
```
jetuk: Should we be explicit about which types to use as ints here? And throw an error it's something else?

snorfalorpagus: Yes. I'll make the change.

PR #305, opened 2016-10-18
Removed some unused subclasses from domains.river

### LINKED ISSUE #232: Remove `domains.river.RiverAbstraction`
At the workshop someone had some trouble using a `RiverAbstraction` node without realising this was an `Output` not a `Link`. I myself was confused, as we don't ever use this type. I think we should remove it (and from corresponding .json examples).

snorfalorpagus:
Closes #232.

### DIFF
--- a/pywr/domains/river.py
+++ b/pywr/domains/river.py
@@ -190,23 +190,6 @@ class Discharge(Catchment):
     """
     pass
 
-
-class DemandDischarge(River):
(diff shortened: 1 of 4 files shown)


snorfalorpagus:
Apparently we have quite a few tests which uses these classes...

jetuk:
Should be easy to change?

Issue #259, opened 2016-09-23
Make core.py more modular


snorfalorpagus:
The `core.py` file is now 1500+ lines. I think this is too much, and would benefit from being split into smaller parts, perhaps:

```
core/__init__.py
core/timestepper.py
core/model.py
core/nodes/__init__.py
core/nodes/basic.py
core/nodes/storage.py
core/nodes/???
core/scenarios.py
core/domains/__init__.py
```

Also, maybe drop the "core" name entirely.

```
from pywr.model import Model
from pywr.nodes import Input, Output, Link
```

jetuk:
I agree with this.

jetuk:
This is complete now?

snorfalorpagus:
I'm undecided if we should actually remove core.py now.

jetuk:
I think I made a comment about a `pywr.api` module. Some other projects use that approach.

snorfalorpagus:
You mentioned it in #311. I'm not familiar with this approach. Does the `api` module just provide access to anything and everything, with the details of the sub-modules hidden? Can you give me an example project?

snorfalorpagus:
The current setup is good enough for now. The code is more manageable and keeping core around keeps everything working for older code.

PR #311, opened 2016-10-25
Refactor of core.py

snorfalorpagus:
This PR is an attempt at a major refactor of `core.py` into smaller, more manageable sub-modules as suggested in #259.

Currently `core.py` still exists and imports the required objects in order to keep the current API functioning. The next step will be to go through all of the tests and switch to the new API.

I was surprised at how easy this was to do. Hopefully the next step is similarly easy (just changing some imports on the tests).


jetuk:
Should we have an `pywr.api` module? We could rename core? I'll look at this when I have a bigger screen.

jetuk:
This looks fine.

Issue #312, opened 2016-11-02
Passing arbitrary keywords to pandas via json is broken


jetuk:
```
"tables" : {
    "My Control Curves" : {
        "url": "Reservoir target and drought curves.xlsx",
        "sheetname": "My Reservoir"
    }
},
   
```

The above doesn't work because "sheetname" is not cleaned up in `load_dataframe` (see https://github.com/pywr/pywr/blob/master/pywr/parameters/_parameters.pyx#L1095). This then causes a failure later on when `Parameter.load` checks for remaining keywords.

```
d:\dev_cntr\pywr\pywr\parameters\_parameters.pyx in pywr.parameters._parameters.Parameter.load (pywr/parameters/_parameters.c:7949)()
    144         if data:
    145             key = list(data.keys())[0]
--> 146             raise TypeError("'{}' is an invalid keyword argument for this function".format(key))
    147         return cls(values, name=name, comment=None)
    148 

TypeError: 'sheetname' is an invalid keyword argument for this function
```

Do we really want a whitelist of keywords in `load_dataframe`? `**data` is passed to all the `pandas.read_*` functions. Should we not just remove all keys from `data` at the end of `load_dataframe`. I.e. assume they are consumed by pandas?

snorfalorpagus:
I read the PR before this issue and came to the same conclusion. All of the pairs are passed on pandas, so that should error if it doesn't like one. I think we can just call `data.clear()` at the end of that function.

PR #313, opened 2016-11-02
Temporary workaround issue #312.

jetuk:
As title. I'd quite like to use the "sheetname" keyword soon.


snorfalorpagus:
See comment in #312. I think we can jump straight to the proper fix.

PR #315, opened 2016-11-03
Fix #312

### LINKED ISSUE #312: Passing arbitrary keywords to pandas via json is broken

jetuk:


snorfalorpagus:
Should we add a test for the feature which this was implemented for ("sheetname")?

snorfalorpagus:
:+1:

Issue #205, opened 2016-08-11
Dependency graph for Parameters and Recorders


snorfalorpagus:
The solver(s) currently update `Parameters` recursively. This results in some inefficiencies (e.g. #107). This is partly resolved using the `CachedParameter` object. I think a better solution would be to resolve the dependency tree of the `Parameters` during `setup`, then at the start of each `solve` calculate the value of every parameter and use the cached result instead of calling `node.get_max_flow` etc.

A similar problem exists for recorders, which means it's not possible to have a meta-recorder (a recorder dependent on another recorder) as the order of resolution is not necessarily sensible.

The resolution order can be solved using the `nx.dfs_postorder_nodes` function.

A down-side to this is for `ControlCurveIndex` parameters, which have a short-circuit optimisation - when they find a control curve the reservoir is above, they don't evaluate the rest.

snorfalorpagus:
Here's the prototype code I wrote for this.
```
from pywr._core import Storage, Node
from pywr.parameters._parameters import Parameter
from pywr.core import AggregatedNode, AggregatedStorage
import networkx as nx

G = nx.DiGraph()

all_parameters = set()

for node in model.nodes:
    if isinstance(node, Storage):
        if isinstance(node.max_volume, Parameter):
            all_parameters.add(node.cost)
        if isinstance(node.cost, Parameter):
            all_parameters.add(node.cost)
    elif isinstance(node, AggregatedStorage):
        if isinstance(node.max_volume, Parameter):
            all_parameters.add(node.cost)
    elif isinstance(node, AggregatedNode):
        if isinstance(node.max_flow, Parameter):
            all_parameters.add(node.max_flow)
        if isinstance(node.min_flow, Parameter):
            all_parameters.add(node.min_flow)
    else:
        if isinstance(node.max_flow, Parameter):
            all_parameters.add(node.max_flow)
        if isinstance(node.min_flow, Parameter):
            all_parameters.add(node.min_flow)
        if isinstance(node.cost, Parameter):
            all_parameters.add(node.cost)
            
to_process = list(all_parameters)
            
visited = set()

while len(to_process) > 0:
    parameter = to_process.pop()
    if not parameter.parents:
        G.add_edge("root", parameter)
    visited.add(parameter)
    for child in parameter.children:
        G.add_edge(parameter, child)
        if not child in visited:
            to_process.append(child)

check = set()
            
for node in nx.dfs_postorder_nodes(G, "root"):
    if node == "root":
        break
    
[...]

PR #323, opened 2016-11-14
WIP: Initial commit of parameter dependency tree

snorfalorpagus:
This is an initial implementation of the parameter tree we've been talking about, which could potentially solve a number of issues we've been having. The tree is generated by first getting the "top level" parameters (i.e. "max_flow", "min_flow", etc.) from all the nodes in the model, then traversing the children (`Parameter.children`) of those parameters recursively. The tree is then used to call the `setup` and `reset` methods for the parameters.

The next step is to implement the same for recorders, which don't currently use the parents/children structure. As is, this probably doesn't work correctly for any parameters that are only referenced by a recorder (which is possible!).

After that, we could extend this for use as the primary refresh method for parameters. This would allow all parameter values to be cached. This would remove the need for `CachedParameter` (and the assoicated issues with detecting parameter type.

I expect the result of this should generally be an small increase in performance as parameters are only ever evaluated once (with exception to the issue raised in #205).

See the `build_parameter_tree` and `flatten_parameter_tree` methods. The flattened tree is generated in such a way that any dependencies should be handled correctly. There is currently no handling of cyclic dependencies - this should be identified and raised as an exception.


snorfalorpagus:
Still can't spell initial...

snorfalorpagus:
The current support for recorders is limited. They aren't added to the tree, but some special cases are used to add parameters held by ParameterRecorder/IndexParameterRecorder into the tree. I think we'll need a hybrid parameter/recorder tree if we want to use it to call `after()` on recorders in the correct order for dependencies.

snorfalorpagus:
This doesn't catch parameters that are held by TablesArrayRecorder.

jetuk:
Sounds good. What's left to complete before merging? Can we make a list?

snorfalorpagus:
I'm not sure how to create any tests specifically for this. The risk is the setup/reset methods aren't called for a parameter.

It could do with some tidying / commenting. Some of the variable names are a bit hack-y.

Before merging it would be good to get the recorders in the tree. To do that they will need the parent/child structure that parameters have currently.

jetuk:
We could test with some dummy parameters that simply register when they're called. Then check it's all in the correct order?

Here's a task list.

- [ ] Dependency tree tests
  - [ ] test setup/reset called
  - [ ] test before/after
- [ ] Tidy up comments and variable names
- [ ] Add recorders to the the tree. Requires parent/child structure to be borrow from parameters.

snorfalorpagus:
I'm going to try and tackle this over the christmas break.

jetuk:
The good news is it ran successfully in its current state through one of our complex models.

jetuk:
Where are we with this?

snorfalorpagus:
I haven't had a chance to look at it much. I think I had a go at adding recorders into the tree, which is the main blocker, but it wasn't as easy as expected. I can't remember why.

pywr/model.py
```diff
@@ -584,6 +607,62 @@ def to_dataframe(self):
         df.columns.set_names('Recorder', level=0, inplace=True)
         return df
 
+    def build_parameter_tree(self):
```
snorfalorpagus: This function is no longer needed.

pywr/_core.pyx
```diff
@@ -530,57 +530,21 @@ cdef class Node(AbstractNode):
             self._cost = self._cost_param.value(ts, scenario_index)
 
     cpdef setup(self, model):
```
snorfalorpagus: These implementations aren't doing anything other than calling their parent method. Do we still need them?

pywr/_core.pyx
```diff
@@ -843,11 +807,11 @@ cdef class Storage(AbstractStorage):
         AbstractStorage.setup(self, model)
         # Setup any Parameters and Recorders
         if self._cost_param is not None:
-            self._cost_param.setup(model)
+            self._cost_param.setup()
```
snorfalorpagus: I think this is no longer needed. The parameter.setup method will be called by model directly.

pywr/parameters/_parameters.pxd
```diff
@@ -1,31 +1,23 @@
 from pywr.recorders._recorders cimport Recorder
+from pywr._component cimport Component
 
 # Forward declations
-cdef class Parameter
-cdef class ArrayIndexedParameter
-cdef class ConstantScenarioParameter
-cdef class ArrayIndexedScenarioMonthlyFactorsParameter
+#cdef class Parameter
```
snorfalorpagus: Remove these if they aren't required (I don't think they are).

pywr/model.py
```diff
@@ -454,7 +471,38 @@ def find_all_routes(self, type1, type2, valid=None, max_length=None, domain_matc
...
     def step(self):
-        """Step the model forward by one day"""
+        """ Step the model forward by one day
+
+        This method progresses the model by one time-step. The anatomy
+        of a time-step is as follows:
+          1. Call `Model.setup` if the `Model.dirty` is True.
+          2. Progress the `Model.timestepper` by one step.
+          3. Call `Model.before` to ensure all nodes and components are ready for solve.
+          4. Call `Model.solve` to solve the linear programme
```
jetuk: Updating parameter values should go here. Between phases 3 and 4.

snorfalorpagus: I've made a start with this here: https://github.com/pywr/pywr/tree/ocptt

snorfalorpagus:
A list of parameters which depend on other parameters:

AggregatedParameter
AggregatedIndexParameter (index)
MinParameter
MaxParameter
NegativeParameter
MaxNegativeParameter
CachedParameter (is this now defunct?)
TablesArrayParameter
IndexArrayParameter
ControlCurveInterpolatedParameter
ControlCurveIndexParameter
Polynomial1DParameter
Polynomial2DStorageParameter
ScaledProfileParameter
DataFrameParameter

This doesn't include those that use .index instead of .value.

jetuk:
We should also check/change how the variables code works. With this change the model will have access to all the `Parmeter` objects and therefore could easily get a list of the variables directly from the model, rather than from each node.

jetuk:
I've "fixed" `DataFrameParameter`. It had a problem because it used to create its own parameter instance during setup. This doesn't work in this new approach because that new parameter is not in the dependency tree.

Instead I've just simplified the approach to pull the value from a `DataFrame` during `value()` rather than use a `NumpyArrayParameter`. There might be a performance hit for doing this.

jetuk:
The AppVeyor builds are failing on GLPK with Python 3.6

pywr/parameters/parameters.py
```diff
@@ -100,34 +100,36 @@ def load(cls, model, data):
...
-
-        param.parents.add(self)
-        self._param = param
 
     def value(self, ts, scenario_index):
-        return self._param.value(ts, scenario_index)
-DataFrameParameter.register()
-
-
+        print(ts)
```
snorfalorpagus: Remove this.

pywr/parameters/parameters.py
```diff
@@ -100,34 +100,36 @@ def load(cls, model, data):
...
+        df = self._df
 
+        if df.ndim == 1:
+            # Single timeseries for the entire run
+            value = df.iloc[i]
+        elif df.shape[1] == 1:
+            value = df.iloc[i, :]
+        else:
+            # We assume the columns are in the correct order for the scenario.
+            j = scenario_index.global_id
```
snorfalorpagus: This isn't correct. See how ArrayIndexedScenarioParameter works with this.

```
return self.values[ts._index, scenario_index._indices[self._scenario_index]]
```

We need to add a test that catches this.

pywr/parameters/parameters.py
```diff
@@ -100,34 +100,36 @@ def load(cls, model, data):
...
-DataFrameParameter.register()
-
-
+        print(ts)
+        i = ts.index
+        df = self._df
 
+        if df.ndim == 1:
+            # Single timeseries for the entire run
+            value = df.iloc[i]
```
snorfalorpagus: Is there a benefit to cythonising this and accessing values more directly? This is what the old approach did by handing off to ArrayIndexedParameter.

snorfalorpagus:
I've refactored DataFrameParameter into a cython parameter.

snorfalorpagus:
I think that's everything except `RecorderThresholdParameter` now.

jetuk:
What about the `.index()` calls? They weren't in your list?

snorfalorpagus:
There are still some calls in `control_curves.py` and `licenses.py`.

jetuk:
I think I've done the ones in `control_curves.py`

pywr/_component.pyx
```diff
@@ -0,0 +1,102 @@
...
+    cpdef before(self):
+        pass
+
+    cpdef after(self):
+        pass
+
+    cpdef finish(self):
+        pass
+
+    property model:
```
snorfalorpagus: This could be exposed in Cython as a `readonly` member, e.g.

```
cdef readonly object model
```

This would provide some consistency between Cython/Python components as everything would access as `self.model`.

pywr/model.py
```diff
@@ -82,18 +84,33 @@ def __init__(self, **kwargs):
...
-        self.parameters = NamedIterator()
-        self.recorders = NamedIterator()
         self.tables = {}
         self.scenarios = ScenarioCollection(self)
 
         if kwargs:
             key = list(kwargs.keys())[0]
             raise TypeError("'{}' is an invalid keyword argument for this function".format(key))
 
+        self.parameter_tree = None
```
snorfalorpagus: This isn't used anywhere and should be removed.

pywr/_component.pyx
jetuk: Agreed. Model is not an extension type.

pywr/parameters/_parameters.pyx
```diff
@@ -594,6 +608,28 @@ cdef class IndexParameter(Parameter):
...
+            num_comb = len(self._model.scenarios.combinations)
+        else:
+            num_comb = 1
+        self.__indices = np.empty([num_comb], np.int32)
+
+    cpdef calc_values(self, Timestep timestep):
+        cdef ScenarioIndex scenario_index
+        for scenario_index in self._model.scenarios.combinations:
+            self.__values[<int>(scenario_index.global_id)] = self.value(timestep, scenario_index)
+            self.__indices[<int>(scenario_index.global_id)] = self.index(timestep, scenario_index)
```
snorfalorpagus: I wonder if the index should be calculated before the value? The value tends to depend on the index.

pywr/parameters/_parameters.pyx
```diff
@@ -613,8 +649,8 @@ cdef class IndexedArrayParameter(Parameter):
     -----
     Float arguments `params` are converted to `ConstantParameter`
     """
-    def __init__(self, index_parameter=None, params=None, **kwargs):
-        super(IndexedArrayParameter, self).__init__(**kwargs)
+    def __init__(self, model, index_parameter=None, params=None, **kwargs):
```
snorfalorpagus: I don't think `index_parameter` or `params` are optional - remove `None`?

tests/test_core.py
```diff
@@ -222,17 +222,16 @@ def test_timeseries_excel(simple_linear_model, filename):
...
+    model.timestepper.start = ts.dataframe.index[0] + 5
+    model.timestepper.end = ts.dataframe.index[-1] - 12
 
     # need to assign parameter for it's setup method to be called
     model.nodes["Input"].max_flow = ts
-    model.setup()
 
-    # test accessing a specific value
-    timestep = build_timestep(model, '2015-01-15')
+    model.run(until_date=datetime.datetime(2015, 1, 14))
```
snorfalorpagus: This will need to be changed for #421, but let's leave it for then.

tests/test_parameters.py
```diff
@@ -682,7 +718,7 @@ class NewParameter(Parameter):
     class NewParameter(IndexParameter):
         DATA = 43
         def __init__(self, *args, **kwargs):
-            IndexParameter.__init__(self)
+            pass
```
snorfalorpagus: Remove the `__init__` method entirely? It's not in the previous definition.

tests/test_recorders.py
```diff
@@ -169,14 +169,17 @@ def test_numpy_parameter_recorder(simple_linear_model):
 def test_numpy_index_parameter_recorder(simple_storage_model):
     """
     Test the NumpyArrayIndexParameterRecorder
+    
+    Note the parameter is recorder at the start of the timestep, while the
```
snorfalorpagus: recorded

pywr/_component.pyx
```diff
@@ -0,0 +1,102 @@
...
+    See also
+    --------
+    `Model`
+
+    """
+    def __init__(self, model, name=None, comment=None):
+        self._model = model
+        self._name = name
+        self.comment = comment
+        model.component_graph.add_edge("root", self)
```
jetuk: Should this reset `model.component_tree_flat = None`? I.e. a new component is added so the flattened tree is out of date.

pywr/model.py
jetuk: Should the docstring be more explicit about the call to calc_values during (3)?

pywr/parameters/_parameters.pyx
```diff
@@ -65,33 +29,28 @@ cdef class Parameter:
...
-        cdef Parameter child
-        for child in self.children:
-            child.reset()
-
-    cpdef before(self, Timestep ts):
-        cdef Parameter child
-        for child in self.children:
-            child.before(ts)
+    cpdef setup(self):
+        cdef int num_comb
```
jetuk: Doesn't call `super`

pywr/parameters/_parameters.pyx
```diff
@@ -623,7 +659,7 @@ cdef class IndexedArrayParameter(Parameter):
         for p in params:
             if not isinstance(p, Parameter):
                 from pywr.parameters import ConstantParameter
-                p = ConstantParameter(p)
+                p = ConstantParameter(model, p)
             self.params.append(p)
 
         for param in params:
```
jetuk: This should be `for param in self.params`. It will miss any that are coerced to a `ConstantParameter` above.

pywr/parameters/_parameters.pyx
```diff
@@ -829,8 +853,8 @@ cdef class AggregatedParameter(AggregatedParameterBase):
         The aggregation function. Must be one of {"sum", "min", "max", "mean",
         "product"}, or a callable function which accepts a list of values.
     """
-    def __init__(self, parameters, agg_func=None, **kwargs):
-        super(AggregatedParameter, self).__init__(**kwargs)
+    def __init__(self, model, parameters, agg_func=None, **kwargs):
```
jetuk: The `parameters` argument could be moved to the parent class.

pywr/parameters/_parameters.pyx
```diff
@@ -890,8 +914,8 @@ cdef class AggregatedIndexParameter(AggregatedParameterBase):
         The aggregation function. Must be one of {"sum", "min", "max", "any",
         "all"}, or a callable function which accepts a list of values.
     """
-    def __init__(self, parameters, agg_func=None, **kwargs):
-        super(AggregatedIndexParameter, self).__init__(**kwargs)
+    def __init__(self, model, parameters, agg_func=None, **kwargs):
```
jetuk: See above comment re parent class.

pywr/parameters/_parameters.pyx
```diff
@@ -1078,7 +1103,7 @@ cdef class RecorderThresholdParameter(IndexParameter):
 
     cpdef double value(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
         """Returns a value from the values attribute, using the index"""
-        cdef int ind = self.index(timestep, scenario_index)
+        cdef int ind = self.index(timestep, scenario_index) # don't use get_index!
```
jetuk: Does this need fixing?

pywr/recorders/_recorders.pyx
```diff
@@ -64,22 +60,9 @@ cdef class Recorder:
...
-        pass
-
-    cpdef reset(self):
-        pass
-
-    cpdef int save(self) except -1:
-        return 0
-
-    cpdef finish(self):
+    cpdef after(self):
```
jetuk: This can be deleted?

jetuk:
I've reviewed the code, but not the tests. Will do the tests a bit later.

pywr/_component.pyx
```diff
@@ -0,0 +1,102 @@
...
+     one component that is a parent of another is called first.
+
+    See also
+    --------
+    `Model`
+
+    """
+    def __init__(self, model, name=None, comment=None):
+        self._model = model
+        self._name = name
```
snorfalorpagus: Check for name collision.

pywr/parameters/_parameters.pyx
snorfalorpagus: If we swapped the order that `get_values`/`get_indices` get called in `before` then we could.

pywr/_component.pyx
```diff
@@ -0,0 +1,102 @@
...
+        self.comment = comment
+        model.component_graph.add_edge("root", self)
+        self.parents = GraphInterface(self)
+        self.children = GraphInterface(self)
+
+    property name:
+        def __get__(self):
+            return self._name
+        def __set__(self, value):
```
snorfalorpagus: If the name is not None we should probably check for collision (here, and in `__init__`).

pywr/model.py
```diff
@@ -82,18 +84,33 @@ def __init__(self, **kwargs):
             # use default solver
             solver = solver_registry[0]
         self.solver = solver()
+        self.component_graph = nx.DiGraph()
+        self.component_graph.add_node("root")
```
snorfalorpagus: SGTM. It should live in `_components`.

pywr/_component.pyx
snorfalorpagus: This breaks a bunch of the control curve tests.

tests/test_recorders.py
```diff
@@ -186,18 +189,18 @@ def test_numpy_index_parameter_recorder(simple_storage_model):
     assert(res_rec.data.shape == (5, 1))
     assert_allclose(res_rec.data, np.array([[7, 4, 1, 0, 0]]).T, atol=1e-7)
     assert (lvl_rec.data.shape == (5, 1))
-    assert_allclose(lvl_rec.data, np.array([[0, 1, 2, 2, 2]]).T, atol=1e-7)
+    assert_allclose(lvl_rec.data, np.array([[0, 0, 1, 2, 2]]).T, atol=1e-7)
```
jetuk: Why has this result changed?

tests/test_recorders.py
```diff
@@ -186,18 +189,18 @@ def test_numpy_index_parameter_recorder(simple_storage_model):
...
 
 
     df = lvl_rec.to_dataframe()
     assert df.shape == (5, 1)
-    assert_allclose(df.values, np.array([[0, 1, 2, 2, 2]]).T, atol=1e-7)
+    assert_allclose(df.values, np.array([[0, 0, 1, 2, 2]]).T, atol=1e-7)
```
jetuk: And here?

tests/test_recorders.py
jetuk: Discussed this. Happy that these have changed.

tests/test_recorders.py
jetuk: Discussed this. Happy that these have changed.

tests/test_recorders.py
snorfalorpagus: See comment at the top of the test. Essentially the parameter is recorded at the start of the timestep, but the storage is recorded at the end.

PR #326, opened 2016-11-17
Add support for literals in AggregatedParameters. Closes #325.

### LINKED ISSUE #325: AggregatedParameter doesn't support literals as parameters
The following parameter is not valid, as the constant is not wrapped as a "constant" parameter in JSON.
```
        "agg_inflow": {
            "type": "aggregated",
            "agg_func": "product",
            "parameters": [
                "my_inflow",
                0.520
            ]
        }
```
This raises the exception:
```
AttributeError: 'float' object has no attribute 'setup'
```

I think this is something we should support. Perhaps related to #323.

snorfalorpagus:
This PR adds support for writing a literal into the parameter list for AggregatedParameter. The current requirement is that this is wrapped in a {"type": "constant", "value": X} parameter.

No changes are made to initalisation from Python space. Although it would be possible to do the same thing (convert floats to ConstantParameter), there might be some oddities with the .add/.remove methods, e.g. how should this work:

```
p = AggregatedParameter()
p.add(3.0)
p.remove(3.0)
```

You might expect this to work, but if 3.0 has been converted to a ConstantParameter that is what you'd need to pass to .remove. Also, not just equivalent (e.g. another constant == 3.0) as the set remove function works on the id() of the object, which the user doesn't even have a reference to!

Thoughts @jetuk?

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -667,6 +667,11 @@ _agg_func_lookup = {
     "all": AggFuncs.ALL,
 }
 
+def wrap_const(value):
+    if isinstance(value, (int, float)):
+        value = ConstantParameter(value)
+    return value
+
 cdef class AggregatedParameterBase(IndexParameter):
     """Base class for aggregated parameters
 
@@ -677,7 +682,8 @@ cdef class AggregatedParameterBase(IndexParameter):
         parameters_data = data.pop("parameters")
         parameters = set()
         for pdata in parameters_data:
-            parameters.add(load_parameter(model, pdata))
+            parameter = load_parameter(model, pdata)
+            parameters.add(wrap_const(parameter))
 
         agg_func = data.pop("agg_func", None)
         return cls(parameters=parameters, agg_func=agg_func, **data)
--- a/tests/test_parameters.py
+++ b/tests/test_parameters.py
@@ -250,10 +250,7 @@ class TestAggregatedParameter:
             "type": "aggregated",
             "agg_func": "product",
             "parameters": [
-                {
-                    "type": "constant",
-                    "value": 0.8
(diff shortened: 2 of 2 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -667,6 +667,11 @@ _agg_func_lookup = {
     "all": AggFuncs.ALL,
 }
 
+def wrap_const(value):
```
jetuk: Should `BaseControlCurveParameter` use this helper function too?

jetuk:
Hmm not sure about that. How would you remove `3.0` anyway? What if you did 

```
p.add(3.0)
p.add(3.0)
p.remove(3.0)
```

What happens?

snorfalorpagus:
The aggregated parameters behave like sets. If `p` was a normal set the second call to add 3.0 wouldn't do anything as 3.0 is already a member.

Our intended usage of the aggregated parameters is a little different. Two members of an aggregated parameter may have the same value, but be different objects. e.g. if you have `demand = 3 * 3 * value`. This isn't the same as `demand = 3 * value`.

I think we can preserve the existing behaviour, but as a convenience convert any literals to ConstantParameter instances. If the user needs to remove a parameter, they can get a reference to the actual parameter.

jetuk:
OK. That's fine it's just a bit "magic" the `float` gets converted to `ConstantParameter`, but I guess that's fine if it's documented.

snorfalorpagus:
The alternative is that we leave it as-is and just do the magic conversion in JSON?

jetuk:
I'm lost on this one. Do we merge now or is there more to do?

snorfalorpagus:
Merged as-is without making any changes to the Python API.

Issue #322, opened 2016-11-14
Add support for writing JSON


snorfalorpagus:
Add support for create a JSON representation of a `Model` from an existing model.

e.g.

```
data = model.dumps()
```

snorfalorpagus:
Implementing this should highlight where parameters don't allow introspection, i.e. once created their information is trapped in cython-space.

jetuk:
This will require a write method on each object?

snorfalorpagus:
Yes. Any object that currently has a load method will require a dump method.

jetuk:
It might be possible to use marshmallow: https://marshmallow.readthedocs.io/en/latest/

Not sure how fiddly this will get because of how pywr actually loads the different objects (i.e. non-linear). How that fits with the `@post_load` decorate I'm not sure.

PR #327, opened 2016-11-17
WIP: Initial commit of JSON dump method.

snorfalorpagus:
This PR begins to address #322. There is still much to do. In particular:

- [ ] Implement `dump` method for all parameters
- [ ] Handle named parameters (i.e. parameters with a reference)
- [ ] Add a test for all JSON files (using pytest magic?)


jetuk:
Why the close?

PR #333, opened 2016-12-01
Improvements to DataFrame frequency handling.

jetuk:
- Updated DataFrame frequency inferrence only happen if there is no existing frequency.
- Made the alignment and resampling function handle a special case of `Week` frequency by treating it as a 7D frequency.
- Added weekly tests based on the above and timeseries1.csv


tests/test_run.py
```diff
@@ -119,16 +120,18 @@ def test_run_timeseries1(solver):
         assert_allclose(demand1.flow, min(expected, 23.0), atol=1e-7)
 
 
-def test_run_timeseries1_xlsx(solver):
-    model = load_model('timeseries1_xlsx.json', solver=solver)
+# Contains an out of range date for pandas.to_datetime
+@pytest.mark.parametrize("json_file", ['timeseries1_weekly.json', 'timeseries1_weekly_hdf.json'])
+def test_run_timeseries1(solver, json_file):
```
snorfalorpagus: This needs to be renamed as it has the same name as the previous test.

snorfalorpagus:
This looks OK, provided that the test that isn't being run still passes when you fix it.

PR #349, opened 2017-01-12
Added casting in TablesArrayParameter.reset

snorfalorpagus:
Without the change in this commit you get a type error if you try to read an HDF5 table using the TablesArrayParameter with an 8 or 16 bit integer types or 32 bit float type.

I haven't added a test for this yet.

This is required to work with the DSL tables we've recently generated.


jetuk:
Fix looks fine. Merge whenever the conda stuff is fixed.

PR #355, opened 2017-01-19
Added a property to set the `agg_func`

jetuk:
in both `AggregatedParameter` and `AggregatedRecorder`

Also added "median" as an option for the `AggregatedRecorder`


snorfalorpagus:
These changes seem OK to me.

The default value of agg_func on recorder seems to be "mean". This seems a little strange? The recorder may not even have any aggregation.

PR #356, opened 2017-01-19
Added ArrayIndexedScenarioMonthlyFactorsParameter JSON support

### LINKED ISSUE #339: ArrayIndexedScenarioMonthlyFactorsParameter needs JSON support
See the change in value() made to MonthlyProfileParameter - this was an optimisation that should be copied across to the scenario variable version.

Also... wow. What a parameter name...

snorfalorpagus:
This adds support for creating `ArrayIndexedScenarioMonthlyFactorsParameter` objects from JSON.

I've used a new syntax here which is different to anything we've done before. Some parameters have an option to load their data from an external file. But this parameter has two (values and factors) that could be external. Can you have a look at how I've handled this and let me know what you think @jetuk?

Closes #339.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -428,7 +428,7 @@ cdef class ArrayIndexedScenarioMonthlyFactorsParameter(Parameter):
         if factors.ndim != 2:
             raise ValueError("Factors must be two dimensional.")
 
-        if scenario._size != factors.shape[0]:
+        if factors.shape[0] != scenario._size:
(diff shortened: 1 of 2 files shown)


jetuk:
Why is this new syntax?

snorfalorpagus:
Previously the entire parameter would be passed to load_parameter_values. This use the same function, but called twice. Once for the time series values, once for the function. So it's an extension of the existing syntax. I think it works quite well but wanted your opinion.

jetuk:
I think it looks fine. Will double check and merge.

PR #360, opened 2017-01-21
Add Polynomial parameters

### LINKED ISSUE #334: y=mx+c parameter
I've seen quite a few inflows that are based on regression from another inflow. It's ugly/slow to create these using the existing AggregatedParameter objects. It would be nice to have a parameter that could calculate y=mx+c where y is the new flow and x is the original flow.

In theory we could implement something more complex, that took an arbitrary equation. I don't think this is needed right now though.

jetuk:
Two generic polynomial parameters.

The 1D parameter can be related to either a node (for flow), storage (for volume) or another parameter. The 2D parameter is a bit more specialised and uses a storage and parameter for the 2 dimensions. I couldn't think of a clean way to make the choice of the dimensions general while also being efficient.

Fixes #334

### DIFF
--- a/pywr/parameters/_polynomial.pyx
+++ b/pywr/parameters/_polynomial.pyx
@@ -1,0 +1,84 @@
+""" Module contains Polynomial Parameters """
+from ._parameters import load_parameter
+import numpy as np
+cimport numpy as np
+
+
+cdef class Polynomial1DParameter(Parameter):
+    """ Parameter that returns the result of 1D polynomial evaluation
+
+    The input to the polynomial can be either:
+     - The previous flow of the attached node (default)
+     - The previous flow of another `AbstractNode`
+     - The current storage of an `AbstractStorage` node
+     - The current value of another `Parameter`
+
+    Parameters
+    ----------
+    node : `AbstractNode`
+        An optional `AbstractNode` the flow of which is used to evaluate the polynomial.
+    storage_node : `Storage`
+        An optional `Storage` node the volume of which is used to evaluate the polynomial.
+    parameter : iterable of Parameter objects or single Parameter
+        An optional `Parameter` the value of which is used to evaluate the polynomial.
+    use_proportional_volume : bool
+        An optional boolean only used with a `Storage` node to switch between using absolute
+         or proportional volume when evaluating the polynomial.
+    """
+    def __init__(self, coefficients, *args, **kwargs):
+        self.coefficients = np.array(coefficients, dtype=np.float64)
+        self._other_node = kwargs.pop('node', None)
+        self._storage_node = kwargs.pop('storage_node', None)
+        self._parameter = kwargs.pop('parameter', None)
+        self.use_proportional_volume = kwargs.pop('use_proportional_volume', False)
+        # Check only one of the above is given
+        arg_check = [
+            self._other_node is not None,
+            self._storage_node is not None,
+            self._parameter is not None,
+        ]
+        # Check we haven't been given an ambiguous number of objects
+        if arg_check.count(True) > 1:
+            raise ValueError('Only one of "node", "storage_node" or "parameter" keywords should be given.')
+        super(Polynomial1DParameter, self).__init__(*args, **kwargs)
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int i
+        cdef double x, y
+
+        # Get the 'x' value to put in the polynomial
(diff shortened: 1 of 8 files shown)


pywr/parameters/_polynomial.pyx
```diff
@@ -0,0 +1,140 @@
...
+        self._parameter = kwargs.pop('parameter', None)
+        self.use_proportional_volume = kwargs.pop('use_proportional_volume', False)
+        # Check only one of the above is given
+        arg_check = [
+            self._other_node is not None,
+            self._storage_node is not None,
+            self._parameter is not None,
+        ]
+        # Check we haven't been given an ambiguous number of objects
+        if arg_check.count(True) > 1:
```
snorfalorpagus: Should this be == 1? What if the user doesn't pass anything?

pywr/parameters/_polynomial.pyx
```diff
@@ -0,0 +1,140 @@
...
+    parameter : iterable of Parameter objects or single Parameter
+        An `Parameter` the value of which is used to evaluate the polynomial.
+    use_proportional_volume : bool
+        An optional boolean only used with a `Storage` node to switch between using absolute
+         or proportional volume when evaluating the polynomial.
+    """
+    def __init__(self, coefficients, storage_node, parameter, *args, **kwargs):
+        self.coefficients = np.array(coefficients, dtype=np.float64)
+        self._storage_node = storage_node
+        self._parameter = parameter
```
snorfalorpagus: The parameter should be added to this parameter's children. `self.children.add(parameter)`

pywr/parameters/_polynomial.pyx
```diff
@@ -0,0 +1,140 @@
...
+        An optional `Parameter` the value of which is used to evaluate the polynomial.
+    use_proportional_volume : bool
+        An optional boolean only used with a `Storage` node to switch between using absolute
+         or proportional volume when evaluating the polynomial.
+    """
+    def __init__(self, coefficients, *args, **kwargs):
+        self.coefficients = np.array(coefficients, dtype=np.float64)
+        self._other_node = kwargs.pop('node', None)
+        self._storage_node = kwargs.pop('storage_node', None)
+        self._parameter = kwargs.pop('parameter', None)
```
snorfalorpagus: The parameter should be added to this parameter's children. `self.children.add(parameter)`

snorfalorpagus:
This looks OK, with the exception of the comments I've added. It should work nicely with #362. If you make those changes I'll give it another read through before merging when I'm more awake!

pywr/parameters/_polynomial.pyx
jetuk: Passing nothing is valid.  It uses `self._node.flow` instead. See line 62.

pywr/parameters/_polynomial.pyx
jetuk: True, and applies to the 1D class too.

jetuk:
I've made the changes to the above, except the one I don't agree with.

I've also added the parameters discussed in #362 to this PR.

pywr/parameters/_polynomial.pyx
snorfalorpagus: This change has been addressed (on a different line).

tests/test_parameters.py
```diff
@@ -686,3 +688,162 @@ def test_invalid_parameter_values():
...
+
+    def test_1st_order_with_parameter(self, model):
+        """ Test 1st order with a `Parameter` """
+        p1 = Polynomial1DParameter([0.5, np.pi], parameter=ConstantParameter(2.0))
+        si = ScenarioIndex(0, np.array([0], dtype=np.int32))
+        ts = model.timestepper.current
+        np.testing.assert_allclose(p1.value(ts, si), 0.5 + np.pi*2.0)
+
+    def test_2nd_order_with_parameter(self, model):
+        """ Test 2nd order with a `Parameter` """
```
snorfalorpagus: This test could be made more clear if you used a variable when defining the constant parameter (e.g. `x = 2.0` `p1 = ConstantParameter(x)` ... `0.5 + np.pi*x + 3.0*x**2`). Also, don't redefine p1.

snorfalorpagus:
I've made one more comment. I can't see anything wrong with the code, it's just a style thing.

tests/test_parameters.py
jetuk: Addressed in the latest commit.

PR #361, opened 2017-01-23
Fixed JSON loading for RiverSplitWithGauge

snorfalorpagus:
RiverSplitWithGauge was incorrectly using `load_parameter` in it's init method. This moves the calls into `load` where they belong.

### DIFF
--- a/pywr/domains/river.py
+++ b/pywr/domains/river.py
@@ -171,9 +171,6 @@ class RiverSplitWithGauge(RiverSplit):
          the number of extra slots required.
     """
     def __init__(self, model, name, mrf=0.0, cost=0.0, mrf_cost=0.0, **kwargs):
-        """Initialise a new RiverSplit instance
-
-        """
         kwargs['cost'] = [mrf_cost, cost]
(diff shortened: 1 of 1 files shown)


pywr/domains/river.py
```diff
@@ -170,17 +170,23 @@ class RiverSplitWithGauge(RiverSplit):
         The identifiers to refer to the slots when connect from this Node. Length must be one more than
          the number of extra slots required.
     """
-    def __init__(self, *args, **kwargs):
+    def __init__(self, model, name, mrf=0.0, cost=0.0, mrf_cost=0.0, **kwargs):
         """Initialise a new RiverSplit instance
```
jetuk: This docstring can be deleted. We should defer to the class docstring here, but has no argument descriptions.

snorfalorpagus: I can't remember what the rule for this is although I remember talking about it with you (and you being correct and me being mistaken).

jetuk: The numpy docs permit documenting `__init__` under class docstring. I only meant that the current `__init__` docstring is wrong and should be removed.

Issue #332, opened 2016-11-23
Evaporation and minimum storage


snorfalorpagus:
Our current approach to dead storage - setting the minimum volume of the node - doesn't play nicely with evaporation, which may in theory draw down a reservoir below the dead storage level. The effect is likely to be minor.

Perhaps it would be possible to have different minimum storage volumes for different storage outputs in the same compound node?

jetuk:
This doesn't seem very trivial to do in a clean fashion?

PR #372, opened 2017-01-31
Storage current_pc now accounts for min_volume != 0

### LINKED ISSUE #368: Minimum storage used by control curve parameters
Some of the control curve parameters do an interpolation when a storage falls below it's control curve, like this:

```
bottom = 0.0
weight = (current percent - bottom) / (control curve - bottom)
```

Should we be using the minimum storage attribute for the value of `bottom` instead of the control curve? Two storages with the same amount of available water could report completely different resource states.

jetuk:
Partially fixes #368

Change fixes `Storage`, but I'm not sure what to do about `AggregatedStorage`. @snorfalorpagus any ideas?

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -792,6 +792,7 @@ cdef class Storage(AbstractStorage):
     cpdef _reset_storage_only(self):
         cdef int i
         cdef double mxv = self._max_volume
+        cdef double mnv = self._min_volume
         cdef ScenarioIndex si
 
         for i, si in enumerate(self.model.scenarios.combinations):
@@ -799,8 +800,11 @@ cdef class Storage(AbstractStorage):
             # Ensure variable maximum volume is taken in to account
             if self._max_volume_param is not None:
                 mxv = self._max_volume_param.value(self.model.timestepper.current, si)
+            # Ensure variable minimum volume is taken in to account
+            if self._min_volume_param is not None:
+                mnv = self._min_volume_param.value(self.model.timestepper.current, si)
             try:
-                self._current_pc[i] = self._volume[i] / mxv
+                self._current_pc[i] = self._volume[i] / (mxv - mnv)
             except ZeroDivisionError:
                 self._current_pc[i] = np.nan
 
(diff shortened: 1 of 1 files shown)


snorfalorpagus:
I'm going to need some persuading on this. The difficultly with `AggregateStorage` highlights the non-physicalness of negative volumes.

jetuk:
The code we have is effectively a weighted average of current %. The aggregated version could be the same approach. This way it would account for `min_volume` in a consistent manner.

snorfalorpagus:
I think this could cause some problems for us. All of the control curves in Aquator are given in percentage assuming 0% = empty. If we change the calculation of storage percent, all the control curves will be in the wrong place in absolute terms, assuming we're using min_volume as dead storage.

jetuk:
Do we use min_volume as dead storage? What about evaporation?

snorfalorpagus:
I have used min_volume as dead storage in some models. Yes there is an issue with evaporation. See #332.

jetuk:
Hmm I guess this needs some thought.

jetuk:
So what are the choices here?

1. Define `current_pc` as based on `max_volume` and zero.
   - This means `max_volume < 0` should not be supported? Do we test for this?
   - It also means `min_volume != 0` would produce unexpected results?
   - Should it mean we don't support negative volumes at all?
2. Redefine `current_pc` to be based on `max_volume - min_volume`.
   - Different to other software. 
   - Makes non-zero hard limits in reservoirs more difficult if their control curves are 0 based proportions.
3. Add a second `current_pc` type attribute that does (2), leave `current_pc` as it is.
4. Make the choice between (1) and (2) user defined by adding a flag to the `Storage` node.
   - Potentially a performance hit for checking this during `after`?

jetuk:
I think our consensus was to do (1) and not support negative volumes?

snorfalorpagus:
Yes. I don't think storages with a negative volume have a sensible meaning.

I think we'll need to support some of the control curves using a non-zero value for their "bottom", but it needs to be optional.

Closing this in favour of something that makes negative min_volume invalid.

PR #373, opened 2017-02-01
Added scaling and offset factors to polynomial parameters.

jetuk:


snorfalorpagus:
Looks OK, but needs a test. Obviously they are being tested implicitly, but I'd like something that explicitly tests the non-defaults.

jetuk:
I think this is good to go now.

PR #375, opened 2017-02-01
Fixed bug with default flow for Catchment when loading from JSON

jetuk:


snorfalorpagus:
Is there a reason we're not using the normal Catchment `__init__` method when loading from JSON? (Don't ask me - all I did was write this function!)

jetuk:
It's the call to `load_parameter` that needs to happen when loading the JSON.

snorfalorpagus:
This is OK. Needs cherry picking.

snorfalorpagus:
Merged.

PR #381, opened 2017-02-07
Fixed a bug when passing a tuple of strings and Parameter instances t…

jetuk:
…o `TablesRecorder`


snorfalorpagus:
Can you explain what's going on here?

jetuk:
Basically `param` should already be a `Parameter` objective if `param` is not a basestring. Setting `param = p` undoes the previous few lines  if `p` is a tuple.

PR #383, opened 2017-02-08
Fixed a mistake in the tests for the FDC deviation recorder 

jetuk:
It wasn't properly testing the scenarios. Ooops.


pywr/_recorders.pyx
```diff
@@ -494,7 +494,8 @@ cdef class FlowDurationCurveDeviationRecorder(FlowDurationCurveRecorder):
                 # Cache the target FDC to use in this combination
                 trgt_fdc = self._target_fdc[:, j]
                 # Finally calculate deviation
-                self._fdc_deviations[:, i] = np.divide(np.subtract(self._fdc[:, i], trgt_fdc), trgt_fdc)
+                for k in range(trgt_fdc.shape[0]):
+                    self._fdc_deviations[k, i] = (self._fdc[k, i] - trgt_fdc[k])  / trgt_fdc[k]
```
snorfalorpagus: I don't follow what's different between these two without playing with some data.

jetuk: I can't remember exactly what was happening, but the slicing operations on the memory view were behaving weird... at least on windows.

jetuk:
Can this be merged?

Issue #385, opened 2017-02-09
Could ScenarioCollection iterate through user defined combinations?


jetuk:
Currently `ScenarioCollection` iterates through the product of all scenario ensembles. It would be useful if the user could specify their own combinations to run either for debugging or just analysis on a subset. 

What will break if this changes? Presumably `len(model.scenarios.combinations)` is used a lot to check the correct size of things that vary by scenario. In this proposal there would be two lengths: the maximum possibly and the number to run. These could then be different.

Thoughts?

snorfalorpagus:
I think this could be useful, especially if you could use slices for each scenario.

model.scenarios[0:50,(1,2)]

I don't think this should be too difficult to make transparent to most of the existing code.

snorfalorpagus:
Turns out `model.scenario.combinations` is just a list.

https://github.com/pywr/pywr/blob/master/pywr/_core.pyx#L73-L79

It should be trivial to filter this list as it's created. Any suggestions for a syntax? I have a use case for this right now.

jetuk:
I think there's two ways to setup,

1. With a truncated range of each scenario.
2. A predefined list of tuples to run.

JSON format for (1) could be like so. Where slice could be an integer for a single scenario or anything that could be passed to `slice`. We would have to parse a string for that I think.
```
"scenarios": [
  { "name": "A", "size": 10, "slice": ":2"},
  { "name": "B", "size": 4, "slice": "0"},
]
```
For (2) you could provide an optional list of lists. The one below would only run 4 combinations. 

```
"scenario-combinations": [
  [0, 0],
  [0, 2],
  [4, 0],
  [4, 2],
]
```
I think these could be passed to the `ScenarioCollection` and if no predefined combinations are set it iterates through per scenario defined slices if available, otherwise does everything.

snorfalorpagus:
https://github.com/pywr/pywr/commit/2902452222fdac420cff4f4e2df110b2410dd3b8

Looks like `ScenarioCombinations` isn't used but was never removed.

jetuk:
Closed by #387

PR #387, opened 2017-02-10
Added support for slicing scenarios

snorfalorpagus:
Added support for slicing scenarios, in order to run only a subset.

Example usage:

```
scenario_weather = Scenario(model, size=200)
scenario_demand = Scenario(model, size=3)
scenario_weather.slice = slice(0, 50, 1) # only the first 50 weather scenarios
scenario_demand.slice = slice(2, 3, 1) # only the last demand scenario
```

This partially addresses #385. Not implemented is JSON support (is this needed?) or support for arbitrary combinations.

I'd like to use this next week. We can add the other bits as needed.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -10,7 +10,7 @@ cdef class Scenario:
     """ Represents a scenario in the model.
 
     Typically a scenario will be used to run many similar models simultaneously. A small
-     number of `Parameter` objects in the model will be return different values depending
+     number of `Parameter` objects in the model will return different values depending
      on the scenario, but many will not. Multiple scenarios can be defined such that
      some `Parameter` values vary with one scenario, but not another. Scenarios are defined
      with a size that represents the number of ensembles in within that scenario.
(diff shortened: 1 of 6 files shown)


pywr/_core.pyx
```diff
@@ -64,19 +46,22 @@ cdef class ScenarioCollection:
...
-            combinations = [ScenarioIndex(0, np.array([0], dtype=np.int32)), ]
+            # model has no scenarios defined, implicitly has 1 scenario of size 1
+            combinations = [ScenarioIndex(0, np.array([0], dtype=np.int32))]
         else:
-            combinations = []
-            for i, indices in enumerate(product([sc._size for sc in self._scenarios])):
-                combinations.append(ScenarioIndex(i, indices))
-        self.combinations = combinations
+            # product of all scenarios, taking into account Scenario.slice
+            iter = itertools.product(*[range(scenario._size)[scenario.slice] if scenario.slice else range(scenario._size) for scenario in self._scenarios])
```
jetuk: This line is nuts; I like it.

jetuk:
I think this is fine. I might do the explicit list on top this weekend.

JSON support would be good. I guess it'll be forgotten about if not done now.

jetuk:
Did the user defined combinations, but no JSON support.

snorfalorpagus:
I'll review this tomorrow.

PR #388, opened 2017-02-11
Added ModelResult class

### LINKED ISSUE #386: Object returned by `Model.run`
The `run` method currently returns the timestep object from the last step in the model. It would be useful if something with more information was returned, e.g. number of timesteps (already available from timestepper) but also speed (= steps / second).

snorfalorpagus:
Added `ModelResult` class, returned by `Model.run()`.

```
res = model.run()
print("{.0f} steps/second".format(res.speed))
```

Closes #386.

Anything further to add?

### DIFF
--- a/pywr/model.py
+++ b/pywr/model.py
@@ -7,6 +7,7 @@ import copy
 from packaging.version import parse as parse_version
 import warnings
 import inspect
+import time
 
 import pywr
 from pywr.timestepper import Timestepper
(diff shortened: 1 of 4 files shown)


jetuk:
I guess it could take the solver states if we rolled that out as a generic dictionary available from every solver.

snorfalorpagus:
I think it's reasonable to just put the solver stats dictionary as a member on the results object. Each solver could provide different stats, relevant to the solver specifics.

I'll fix the divide by zero error, returning infinity.

snorfalorpagus:
It would be nice to give it a string and HTML repr to show a little table.

jetuk:
I think I've done most of what we wanted in here now.

PR #392, opened 2017-02-15
Removed redundant calls to load_recorders

snorfalorpagus:
I'm not really sure what's going on here. I was having some problems with a custom variant of ParameterRecorder, where the recorder was failing to load from JSON. I looked at the code to find this section which seems redundant - the recorders are already loaded by this point. Trying to load them again was causing an exception.


snorfalorpagus:
The appveyor failure on branch is fixed in master.

PR #394, opened 2017-02-16
Added an option for H5Store and TablesRecorder to create directories

jetuk:
Default is the same as it was, but you can now pass `create_directories=True` and the reorder will make any missing directories for you.


snorfalorpagus:
Looks OK. Test please.

PR #410, opened 2017-02-21
ScenarioCollection.multiindex now respects scenario slices. Closes #409.

### LINKED ISSUE #409: ScenarioCollection.multiindex doesn't use scenario slices
ScenarioCollection.multiindex doesn't respect the new slices property of scenario objects. This results in indexes that aren't the right shape for the data.

snorfalorpagus:


jetuk:
This doesn't use the names if they're available.

jetuk:
This is ready?

snorfalorpagus:
I think so.

PR #420, opened 2017-03-01
WIP: Event recorder

jetuk:
Created PR for discussion.

Currently does the following,
- Adds EventRecorder and Event object.
- Basic test for capturing the correct events on a scenario basis.
- Creates a pywr.parameters._thresholds module for threshold parameters.
- Moved RecorderThresholdParameter to new module.
- Added NodeThresholdParameter and StorageThresholdParameter.

Todo 
- Track other things during an event.

Also #323 would break some things in here.

### DIFF
--- a/pywr/parameters/_thresholds.pyx
+++ b/pywr/parameters/_thresholds.pyx
@@ -171,10 +171,11 @@ cdef class RecorderThresholdParameter(AbstractThresholdParameter):
 
     """
 
-    def __init__(self,  model, Recorder recorder, *args, **kwargs):
+    def __init__(self,  model, Recorder recorder, *args, initial_value=1, **kwargs):
         super(RecorderThresholdParameter, self).__init__(model, *args, **kwargs)
         self.recorder = recorder
         self.recorder.parents.add(self)
+        self.initial_value = initial_value
 
     cpdef double _value_to_compare(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
         # TODO Make this a more general API on Recorder
@@ -187,7 +188,7 @@ cdef class RecorderThresholdParameter(AbstractThresholdParameter):
         if index == 0:
             # on the first day the recorder doesn't have a value so we have no
             # threshold to compare to
-            ind = 1
+            ind = self.initial_value
         else:
             ind = super(RecorderThresholdParameter, self).index(timestep, scenario_index)
         return ind
--- a/pywr/parameters/_thresholds.pyx
+++ b/pywr/parameters/_thresholds.pyx
@@ -1,0 +1,161 @@
+cimport numpy as np
+import numpy as np
+
+cdef enum Predicates:
+    LT = 0
+    GT = 1
+    EQ = 2
+    LE = 3
+    GE = 4
+_predicate_lookup = {
+    "LT": Predicates.LT, "<": Predicates.LT,
+    "GT": Predicates.GT, ">": Predicates.GT,
+    "EQ": Predicates.EQ, "=": Predicates.EQ,
+    "LE": Predicates.LE, "<=": Predicates.LE,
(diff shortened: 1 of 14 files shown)


pywr/recorders/events.py
```diff
@@ -0,0 +1,92 @@
...
+
+     See also
+     --------
+     `pywr.parameters._thresholds`
+
+
+     """
+    def __init__(self, model, threshold, *args, **kwargs):
+        super(EventRecorder, self).__init__(model, *args, **kwargs)
+        self.threshold = threshold
```
jetuk: Add a TODO comment regarding setting parent relationship with `self.threshold`

pywr/parameters/_thresholds.pyx
```diff
@@ -0,0 +1,203 @@
...
+        return self.recorder.data[timestep._index - 1, scenario_index._global_id]
+
+    cpdef int index(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
+        """Returns 1 if the predicate evalutes True, else 0"""
+        cdef int index = timestep.index
+        cdef int ind
+        if index == 0:
+            # on the first day the recorder doesn't have a value so we have no
+            # threshold to compare to
+            ind = 1
```
snorfalorpagus: This should be configurable. `self.initial_value`?

pywr/recorders/_recorders.pxd
```diff
@@ -6,6 +6,7 @@ from pywr.parameters._parameters cimport Parameter, IndexParameter
 cdef class Recorder(Component):
     cdef bint _is_objective
     cdef public double epsilon
+    cdef public double ignore_nan
```
snorfalorpagus: This is the wrong type. Should be `bint`.

pywr/parameters/_thresholds.pyx
```diff
@@ -0,0 +1,203 @@
...
+    """
+
+    def __init__(self,  model, Recorder recorder, *args, **kwargs):
+        super(RecorderThresholdParameter, self).__init__(model, *args, **kwargs)
+        self.recorder = recorder
+        self.recorder.parents.add(self)
+
+    cpdef double _value_to_compare(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
+        # TODO Make this a more general API on Recorder
+        return self.recorder.data[timestep._index - 1, scenario_index._global_id]
```
snorfalorpagus: Should this use `self.recorder.values()`?

pywr/recorders/events.py
snorfalorpagus: This has been done.

pywr/recorders/events.py
```diff
@@ -0,0 +1,176 @@
...
+        # Optional different method for aggregating across self.recorders scenarios
+        agg_func = kwargs.pop('recorder_agg_func', kwargs.get('agg_func'))
+        self.recorder_agg_func = agg_func
+
+        super(EventDurationRecorder, self).__init__(model, **kwargs)
+        self.event_recorder = event_recorder
+        self.event_recorder.parents.add(self)
+
+    def setup(self):
+        self._values = np.zeros(len(self.model.scenarios.combinations))
```
snorfalorpagus: This can use `np.empty` as `setup` calls `reset`.

pywr/parameters/_thresholds.pyx
snorfalorpagus: I'm currently using this recorder with "MeanFlowRecorder". I would need to implement `.values` on that, but it shouldn't be difficult.

jetuk: Well I am not sure about this. `Recorder.values` is currently intended to produce a single value for each scenario at the end of the model run. We've not used as a general API for getting a value from a `Recorder` during a run.

PR #428, opened 2017-03-08
Made the event recorders more robust to aggregation method and no events

jetuk:
There was also a stupid bug with the event duration being negative which was in both the code and tests!

### DIFF
--- a/tests/test_recorders.py
+++ b/tests/test_recorders.py
@@ -828,7 +828,10 @@ def cyclical_linear_model(simple_linear_model):
 
 class TestEventRecorder:
     """ Tests for EventRecorder """
-    def test_event_capture_with_storage(self, cyclical_storage_model):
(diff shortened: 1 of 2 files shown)


tests/test_recorders.py
```diff
@@ -887,8 +905,32 @@ def test_event_capture_with_node(self, cyclical_linear_model):
...
+        arry = NumpyArrayStorageRecorder(m, strg)
+
+        # Create the trigger using a threhsold parameter
+        trigger = StorageThresholdRecorder(m, strg, -1.0, predicate='<')
+        evt_rec = EventRecorder(m, trigger)
+        evt_dur = EventDurationRecorder(m, evt_rec, recorder_agg_func=recorder_agg_func, agg_func='max')
+
+        m.run()
+
+        # Ensure there is at least one event
```
snorfalorpagus: This comment is wrong?

PR #435, opened 2017-03-10
Added minimum_event_length to EventRecorder

### LINKED ISSUE #429: EventRecorder should have a minimum length for hysteresis
The user should be able set a minimum of number (default =0) timesteps for an event to be recorded. This could be done either,

1. that an event stays open for N timesteps before it is considered closed, or
2.  simply has to last >= N timesteps to be saved.

jetuk:
- This sets a minimum duration in time-steps that an event must last to be recorded.
- Also added a test for this.

Fixes #429

### DIFF
--- a/pywr/recorders/events.py
+++ b/pywr/recorders/events.py
@@ -31,6 +31,11 @@ class EventRecorder(Recorder):
     ----------
     threshold - IndexParameter
        The parameter that defines the start and end of an event.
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
Python 2.7 doesn't support keyword arguments with default values after variable length positional arguments. Just drop the `*args`.

PR #436, opened 2017-03-10
Added AnnualReturnPeriodRecorder

jetuk:


snorfalorpagus:
I don't totally follow the maths in this. Does this have a problem with leap days in long timeseries?

jetuk:
It entirely depends on what `Recorder` you give it. All the math is doing is attempting to get a integer number of years to divide by. You just do `start.year - end.year`. For example,

1900-01-01 -> 1900-12-31 = 1 year
1900-01-01 -> 1901-01-01 = 1 year
1900-01-01 -> 1901-12-31 = 2 years
1900-01-01 -> 1902-01-01 = 2 years

Generally this should work with a set of `Annual...` recorders. We have one at the moment: ` AnnualCountIndexParameterRecorder`.

snorfalorpagus:
This illustrates the problem:
```
>>> start = pandas.Timestamp("1800-01-01")
>>> end = pandas.Timestamp("2200-01-01")
>>> dt = end - start
>>> remainder_of_year = dt.days / 365.0 % 1.0
>>> remainder_of_year
0.2657534246575324
>>> threshold = 10.0/365.0
>>> threshold
0.0273972602739726
>>> threshold < remainder_of_year < (1 - threshold)
True
```

PR #437, opened 2017-03-12
Performance improvement in calc_values.

jetuk:
- Results from using ScenarioIndex._global_id rather than ScenarioIndex.global_id.
 - Also typed the ScenarioCollection so the loop knows .collections is a list.

There is a bunch of time still spent in `calc_values`, but I think most of this is overhead from it being called from Python. I.e. you would need to type the `Parameter` objects in model to make this faster.

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -20,7 +20,7 @@ cdef class ScenarioCombinations:
     cdef ScenarioCollection _collection
 
 cdef class ScenarioIndex:
-    cdef int _global_id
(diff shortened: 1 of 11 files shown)


snorfalorpagus:
Another solution to this would be to make `scenario_id` and `indices` `readonly` cython attributes. This would mean the syntax is the same in both Cython and Python (making it easier to port Parameters when required).

jetuk:
I think I prefer the readonly for `global_id` but keeping it as is for `indices`. I'll update it.

PR #442, opened 2017-03-14
Vectorised AggregatedParameter.

jetuk:
Here's a benchmark from ASV

```
    before     after       ratio
  [919dd7e7] [5e2e9cfc]
-    1.32μs   802.27ns      0.61  benchmarks.AggregatedParameter.time_calc_values(10)
-    8.50μs     1.18μs      0.14  benchmarks.AggregatedParameter.time_calc_values(100)
-   40.52μs     2.68μs      0.07  benchmarks.AggregatedParameter.time_calc_values(500)
```

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -858,14 +858,14 @@ cdef class AggregatedParameter(AggregatedParameterBase):
                 for i in range(n):
                     accum[i] += values[i]
         elif self._agg_func == AggFuncs.MAX:
-            accum[...] = float("-inf")
+            accum[...] = np.NINF
             for parameter in self.parameters:
                 values = parameter.get_all_values()
                 for i in range(n):
                     if values[i] > accum[i]:
                         accum[i] = values[i]
         elif self._agg_func == AggFuncs.MIN:
-            accum[...] = float("inf")
+            accum[...] = np.PINF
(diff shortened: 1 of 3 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -828,44 +828,64 @@ cdef class AggregatedParameter(AggregatedParameterBase):
...
-            value = 0
+            accum[...] = 0.0
             for parameter in self.parameters:
-                value += parameter.get_value(scenario_index)
+                values = parameter.get_all_values()
+                for i in range(n):
+                    accum[i] += values[i]
         elif self._agg_func == AggFuncs.MAX:
-            value = float("-inf")
+            accum[...] = float("-inf")
```
snorfalorpagus: We should probably just define `inf` at the start of the module (see below also)

snorfalorpagus:
The appveyor failures are usual conda rubbish. Need to test on Windows before merge.

This looks good to me. I was worried that we'd have two implementations of the algorithm (one for value one for all values) but I see you've just removed value. I think we should still implement value, but have it raise an error (something bespoke) to avoid tripping up.

jetuk:
> I think we should still implement value, but have it raise an error (something bespoke) to avoid tripping up.

We discussed that the default `value` method will raise an error already.

PR #445, opened 2017-03-16
Modified TableRecorder so that it can record node deficits

Batch21:


snorfalorpagus:
@jetuk pointed out that this would be handled better as a parameter that returns the current timestep's deficit. I'd still like to have the deficit keyword for TablesRecorder - you can create the required parameters in init. @jetuk can you help @Batch21 with this?

jetuk:
Sorry I didn't get around to talking to @Batch21 about this.

I think we can quickly make a new `Parameter` using something similar to `TotalDeficitNodeRecorder`, see some untested hacking below (i.e. it definitely needs an `__init__`). This parameter computes the timestep's deficit during after. It must be zero in the first time-step (until after). Therefore the `TablesRecorder` must make sure it becomes a parent of this parameter so after is computed before it gets the value.

```
cdef class DeficitParameter(Parameter):
    """
    Recorder to total the difference between modelled flow and max_flow for a Node
    """
    cpdef reset(self):
       super(....)
       # First timestep should be zero
       self.__values[...] = 0

    cpdef calc_values(self, Timestep ts):
        # Don't do anything during calculation 
        pass

    cpdef after(self):
        cdef double max_flow
        cdef ScenarioIndex scenario_index
        cdef Timestep ts = self.model.timestepper.current
        cdef int days = self.model.timestepper.current.days
        cdef AbstractNode node = self.node
        for scenario_index in self.model.scenarios.combinations:
            max_flow = node.get_max_flow(ts, scenario_index)
            # Instead update values here.
            self.__values[scenario_index.global_id] = (max_flow - node._flow[scenario_index.global_id])*days
```

PR #446, opened 2017-03-16
Added Model.find_orphaned_components method. Closes #444.

### LINKED ISSUE #444: Search for unused parameters
Feature request: add a method to the model which returns all parameters in the model that aren't referenced anywhere.

Now that all parameters are being calculated, even if they aren't used, I'm noticing parameters lurking in aggregated models.

snorfalorpagus:
This commit adds a method helpful in searching for parameters in
the model that aren't used. AbstractNode now provides a components
property which yields all of the components it uses. This is done
using a class attribute, component_attrs, which lists all of the
properties on a class which may hold a component (e.g. max_flow).
This will allow nodes to define new properties in the future,
although most will be caught anyway as all nodes in the model
are queried, including the children of complex nodes.

Also something (I used git stash?) seemed to remove a bunch of blank lines. Hopefully that's not going to create merge issues... I'll sort if it does.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -1,4 +1,5 @@
 from pywr._core cimport *
+from pywr._component cimport Component
 import itertools
 import numpy as np
 cimport numpy as np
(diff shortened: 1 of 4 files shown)


pywr/_core.pyx
```diff
@@ -663,6 +683,8 @@ cdef class Storage(AbstractStorage):
         self._domain = None
         self._allow_isolated = True
 
+    component_attrs = ["cost", "initial_volume", "min_volume", "max_volume", "level"]
```
jetuk: `initial_volume` can only ever be a `double`.

snorfalorpagus:
> Should we call find_orphaned_parameters during check?

If we did this I think it should raise a warning only.

snorfalorpagus:
This is ready for merge, I think. Conflict is minor. Checked that something doesn't raise a warning is a PITA.

Issue #427, opened 2017-03-08
Dump linear program to file when non-optimal solution found


snorfalorpagus:
"solution is feasible" (i.e. non-optimal)

```
  File "pywr/solvers/cython_glpk.pyx", line 351, in pywr.solvers.cython_glpk.CythonGLPKSolver.solve (pywr/solvers/cython_glpk.c:9895)
  File "pywr/solvers/cython_glpk.pyx", line 357, in pywr.solvers.cython_glpk.CythonGLPKSolver.solve (pywr/solvers/cython_glpk.c:9786)
  File "pywr/solvers/cython_glpk.pyx", line 447, in pywr.solvers.cython_glpk.CythonGLPKSolver._solve_scenario (pywr/solvers/cython_glpk.c:10920)
RuntimeError: solution is feasible
```

This specific error is probably a symptom of a larger problem in the model. It would be nice if the program could automatically spit out a .lp file with the problem, to check any issues with GLPK.

Any ideas where the file should be saved? How to trigger this?

jetuk:
Current working directory? One issue is making a unique name if multiple runs are being done.

I had thought the solvers should have a generic method to dump the LP (though in theory some solvers might not use a LP). Perhaps a generic "dump some useful debug information" method would be better. 

We would need to catch the `RuntimeError` and clean up by calling this methods during `step`.

jetuk:
This has been done in simple manner in #455

PR #455, opened 2017-04-03
Added a timelimit to GLPK simplex method. 

jetuk:
If this occurs a new basis is initialised and solve retried before giving up.

As I mentioned had a weird error with GLPK getting stuck in a seemingly infinite loop. However the dumped LP was easily solved by using `glpsol`from the command line. 

This fix works around this problem by giving a timelimit to GLPK of 5 seconds (not unreasonable to add anyway). If this fails a new basis is created before a resolve. This solves the current problem. If a second failure occurs the solve raises a `RuntimeError` as previously.

@snorfalorpagus comments?

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -53,6 +53,7 @@ cdef class CythonGLPKSolver:
         # disable console messages
         glp_init_smcp(&self.smcp)
         self.smcp.msg_lev = GLP_MSG_ERR
+        self.smcp.tm_lim = 5000  # 5 second limit
 
     def __init__(self):
         self.stats = None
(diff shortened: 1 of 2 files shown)


jetuk:
This is also solves #427 but it probably needs a better naming convention.

snorfalorpagus:
Time limit seems like a good idea.

The re-solve will trigger if the solution is infeasible. Is this desired?

Any preference on *.mps vs *.lp?

jetuk:
Not really. MPS was the first API call I came from. 

I noticed previously we weren't storing the return value of `glp_simplex` this gives some additional information about how the solve ended. E.g. we could only do the above on a timeout. I think I added the relevant constants to the header. I would be happy with modifying it only to resolve on the timeout. 

Some of these things should be configurable by passing keywords to the solver I think.

PR #458, opened 2017-04-16
Fix moea example

jetuk:
`InspyredOptimisationModel` doesn't obey the `minimise` / `maximise` flag on `is_objective`. I'll open an issue about that. However the example works again.


jetuk:
Can this be merged?

PR #460, opened 2017-04-19
Glpk determinism

jetuk:
This now does a few things to the GLPK solver

- Added ability to dump MPS file for the last LP.
- Added reset API to solvers
- Tracks the basis (row and column status) for each scenario. This is an attempt to make the solver deterministic when the model is rerun or run with different scenarios. Previously the basis was only crashed once on the first solve and then retained for each solve. Now the basis is crashed on the first time-step of each run for each scenario. The basis is then saved and reapplied on a scenario basis. In theory this makes each scenario entirely deterministic in the behaviour of the basis regardless of whether it is run multiple times or with other scenarios.
- Enabled presolver

Also added a simple NaN check to `ControlCurveInterpolatedParameter`

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -466,7 +466,7 @@ cdef class CythonGLPKSolver:
             if status != GLP_OPT or simplex_ret != 0:
                 print("Simplex solve returned: {}".format(simplex_ret))
                 print("Simplex status: {}".format(status))
-                glp_write_mps(self.prob, GLP_MPS_FILE, NULL, 'pywr_glpk_debug.mps')
+                self.dump_mps('pywr_glpk_debug.mps')
                 raise RuntimeError(status_string[status])
         self.stats['lp_solve'] += time.clock() - t0
         t0 = time.clock()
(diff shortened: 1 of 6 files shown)


pywr/solvers/cython_glpk.pyx
```diff
@@ -355,6 +369,47 @@ cdef class CythonGLPKSolver:
             'number_of_nodes': len(self.all_nodes)
         }
 
+    cdef _init_basis_arrays(self, model):
+        """ Initialise the arrays use for storing the LP basis by scenario """
```
snorfalorpagus: used

pywr/solvers/cython_glpk.pyx
```diff
@@ -47,16 +47,27 @@ cdef class CythonGLPKSolver:
     cdef list aggregated
     cdef public object stats
 
+    # Internal representation of the basis for each scenario
+    cdef int[:, :] row_stat
+    cdef int[:, :] col_stat
+    cdef bint is_first_solve
+    cdef bint has_presolved
+    cdef bint use_presolve
```
snorfalorpagus: It would be good to expose this as an option in JSON. We already have a "solver" section.

```
"solver": {
    "name": "glpk",
    "presolve": true
}
```

I don't know if it should default on or off. I remember trying it in the past and found it was slower or no difference with presolving.

jetuk: I didn't help much in our biggest model. So I wouldn't be too upset if this was removed or turned off by default. Currently we have no easy mechanism to pass arbitrary arguments to solvers.

PR #462, opened 2017-04-21
Adds support for Parameter based min & max flow in AggregatedNode

### LINKED ISSUE #461: AggregatedNode min & max flow can not be parameters
#174 added support for min and max flow constraints on `AggregatedNode`. However these are constants and can not be set with a `Parameter`.

jetuk:
- Implemented in GLPK solver only
- Small refactor of the aggregated nodes in the solver
- Found a small bug in the setting of `None` to min_flow & max_flow on Node

Fixes #461

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -462,7 +462,10 @@ cdef class Node(AbstractNode):
             return self._min_flow_param
 
         def __set__(self, value):
-            if isinstance(value, Parameter):
+            if value is None:
+                self._min_flow = 0
+                self._min_flow_param = None
+            elif isinstance(value, Parameter):
                 self._min_flow_param = value
             else:
                 self._min_flow_param = None
@@ -489,6 +492,7 @@ cdef class Node(AbstractNode):
         def __set__(self, value):
             if value is None:
                 self._max_flow = inf
+                self._max_flow_param = None
             elif isinstance(value, Parameter):
                 self._max_flow_param = value
             else:
@@ -559,6 +563,8 @@ cdef class AggregatedNode(AbstractNode):
         self._factors = None
         self._min_flow = -inf
         self._max_flow = inf
+        self._min_flow_param = None
+        self._max_flow_param = None
 
     property nodes:
         def __get__(self):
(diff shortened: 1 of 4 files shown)


pywr/_core.pyx
```diff
@@ -591,37 +597,76 @@ cdef class AggregatedNode(AbstractNode):
...
+            if self._min_flow_param is None:
+                return self._min_flow
+            return self._min_flow_param
+
         def __set__(self, value):
             if value is None:
-                value = inf
-            self._max_flow = value
-            self.model.dirty = True
+                self._min_flow = -inf
```
snorfalorpagus: Why is this `= -inf` here but `= 0` previously (L466)

pywr/solvers/cython_glpk.pyx
```diff
@@ -321,7 +313,7 @@ cdef class CythonGLPKSolver:
                 ind[1+i] = 1+col
                 val[1+i] = 1.0
             glp_set_mat_row(self.prob, row, length, ind, val)
-            glp_set_row_bnds(self.prob, row, constraint_type(min_flow, max_flow), min_flow, max_flow)
+            glp_set_row_bnds(self.prob, row, GLP_FX, 0.0, 0.0)
```
snorfalorpagus: This line isn't needed?

tests/test_agg_constraints.py
```diff
@@ -104,6 +106,27 @@ def test_aggregated_node_max_flow(model):
...
+    """Nodes constrained by the max_flow of their AggregatedNode using a Parameter """
+    A = Input(model, "A", max_flow=20.0, cost=1)
+    B = Input(model, "B", max_flow=20.0, cost=2)
+    Z = Output(model, "Z", max_flow=100, cost=-10)
+
+    A.connect(Z)
+    B.connect(Z)
+
+    agg = AggregatedNode(model, "agg", [A, B])
+    agg.max_flow = ConstantParameter(model, 30.0)
```
snorfalorpagus: We should test with a parameter that actually changes it's value. I wonder if we should roll these tests up with parameterization?

pywr/_core.pyx
jetuk: It defaulted to -inf in `AggregatedNode` before. Default in `Node` was/is zero.

pywr/solvers/cython_glpk.pyx
jetuk: I just followed the pattern from the previous section which includes this line.

tests/test_agg_constraints.py
jetuk: Yes, we should do both. I didn't have time to do anything better.

PR #464, opened 2017-04-22
Make TablesArrayParameter load it's data during setup

jetuk:
Current behaviour means the `TablesArrayParameter` loads data during `reset`. This is inefficient in certain algorithms that reset a model and rerun. The current implementation of the parameter does not expect the data to change between `setup` and `reset`. This moves the loading to `setup` instead.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -303,7 +303,6 @@ cdef class TablesArrayParameter(IndexParameter):
         if self.scenario is not None:
             self._scenario_index = self.model.scenarios.get_scenario_index(self.scenario)
 
-    cpdef reset(self):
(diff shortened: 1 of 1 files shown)


snorfalorpagus:
I think my original logic for putting it in reset was that setup would be used for changes to the model structure (which are generally expensive to recalculated, e.g. routes, LP shape) and reset would be used for changes to the values in the model. The tables array parameter doesn't work well with this though as it's a change in values but potentially takes a long time.

jetuk:
> The tables array parameter doesn't work well with this though as it's a change in values but potentially takes a long time.

It seems a bit niche to support these files changing their content bewteen setup and reset?

Setting dirty is not widely used at the moment. However this parameter would have to probe the file to know if there are any changes during reset.

PR #466, opened 2017-04-28
Fixed #465. TablesRecorder sets Parameters as children.

### LINKED ISSUE #465: TablesRecorder doesn't define it's children
`TablesRecorder` does not define the parameters it records as it's children. This isn't as bad as it could be, as parameter values are calculated in *before*, while the recorder saves the values in *after*, so there should never be a race condition.

It's awkward to fix because `TablesRecorder` accepts a list of parameters _or_ parameter names, which aren't evaluated until `setup` is called which is after the component tree is defined. The easiest workaround I can think of is to only support passing parameters or at least evaluate the list in `__init__` rather than `setup`.

snorfalorpagus:
Parameters added as children to TablesRecorder.

@jetuk It'd be good to have your eyes on this in case I've missed something.

### DIFF
--- a/pywr/recorders/recorders.py
+++ b/pywr/recorders/recorders.py
@@ -7,6 +7,7 @@ from .events import *
 from .calibration import *
 from past.builtins import basestring
 from pywr.h5tools import H5Store
+from ..parameters import load_parameter
 
 def assert_rec(model, parameter, name=None):
     """Decorator for creating AssertionRecorder objects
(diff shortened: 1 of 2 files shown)


jetuk:
Do you need this urgently?

On 28 Apr 2017 11:30 am, "Joshua Arnott" <notifications@github.com> wrote:

@snorfalorpagus <https://github.com/snorfalorpagus> requested your review
on: pywr/pywr#466 <https://github.com/pywr/pywr/pull/466> Fixed #465
<https://github.com/pywr/pywr/issues/465>. TablesRecorder sets Parameters
as children..

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/pywr/pywr/pull/466#event-1061899969>, or mute the thread

snorfalorpagus:
No. In the next week though.

pywr/recorders/recorders.py
```diff
@@ -198,14 +199,46 @@ def __init__(self, model, h5file, nodes=None, parameters=None, where='/', time='
...
 
+    def _add_parameter(self, parameter):
+        try:
+            where, param = parameter
+        except (TypeError, ValueError):
+            where = None
+            param = parameter
+        if isinstance(param, basestring):
+            param = load_parameter(self.model, param)
+        if not param.name:
```
jetuk: Can we add a check in here for parameter names containing "/". I've been bitten by that before. It creates random subgroups in the H5.

snorfalorpagus: Auto-replace with `_`?

jetuk: Probably not a bad idea, but in theory could create a conflict.

snorfalorpagus: I've added a warning and test for this.

PR #468, opened 2017-04-28
Use node._prev_flow in NodeThresholdParameter. Fixes #454

### LINKED ISSUE #454: NodeThresholdParameter should use _prev_flow
The test is not good enough here.

snorfalorpagus:
Fixes #454.

In the initial timestep the previous flow is undefined. Rather than returning a NaN, the parameter will always return 0 for it's index on the initial timestep.

AssertionRecorder doesn't let you check the index directly, so needed a workaround. See #467.


snorfalorpagus:
Good spot. Removed.

PR #471, opened 2017-05-02
Added new Transfer node

### LINKED ISSUE #449: Compound node for transfers
I've been using a compound node which contains a storage with zero volume to reduce the number of routes in highly interconnected models.

The node looks like this, where X and Y are actually many nodes:

```
X ->- | -- storage ->- link -- | ->- Y
```

Internally the storage is an input and output which crucially are *not* connected. This means that for X1..3 and Y1..3 there are 6 routes in the model (3+3), not 9 (3*3).

The node uses iter_slots to expose the correct node when connecting, so you can do this:

```
T = Transfer(max_flow=1.5, cost=10)
X1.connect(T)
T.connect(Y1)
```

I'll submit a PR for this on Monday. What do you think about the node name `Transfer`? It's really applicable anywhere you want to collapse the routes, not just inter-zone transfers (which is what it was designed for).

snorfalorpagus:
This PR adds a compound transfer node, composed of a Storage with zero volume and a Link.

Closes #449.

### DIFF
--- a/pywr/nodes.py
+++ b/pywr/nodes.py
@@ -746,20 +746,31 @@ class AggregatedNode(with_metaclass(NodeMeta, Drawable, _core.AggregatedNode)):
         super(AggregatedNode, self).__init__(model, name, **kwargs)
         self.nodes = nodes
 
-class Transfer(Node):
+class BreakLink(Node):
     """Compound node used to reduce the number of routes in a model
-    
+
+    Parameters
+    ----------
+    model : `pywr.model.Model`
+    name : string
+    min_flow : float or `pywr.parameters.Parameter`
+    max_flow : float or `pywr.parameters.Parameter`
+    cost : float or `pywr.parameters.Parameter`
+
+    Notes
+    -----
+
     In a model with form (3, 1, 3), i.e. 3 (A,B,C) inputs connected to 3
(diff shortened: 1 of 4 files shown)


pywr/nodes.py
```diff
@@ -746,4 +746,86 @@ def __init__(self, model, name, nodes, **kwargs):
         super(AggregatedNode, self).__init__(model, name, **kwargs)
         self.nodes = nodes
 
+class Transfer(Node):
```
jetuk: I think this is a good implementation. My only comment is whether to stick with the name `Transfer`. This is a quite generic extenstion to the basic `Link`. Should it derive its name from that instead?

snorfalorpagus: If we can think of something better, yes. Transfer doesn't really describe what it does either, just what we've used it for.

ZeroStorageLink?
NullStorage?
BreakLink?

jetuk: BufferedLink?

snorfalorpagus: Buffered implies some carryover between timesteps to me.

jetuk: True. I guess BreakLink then?

snorfalorpagus: We agreed on `BreakLink`.

PR #475, opened 2017-05-06
WIP to cythonise Model class

jetuk:
Any opinion on this? It gives about 8% speed improvement on a big model because all the before, after and calc_values calls are now typed.


snorfalorpagus:
This seems sensible to me. I wasn't expecting such a large speed improvement, so it's probably worth doing just because of that. It's not actually that big a change. I wonder if we will end up with everything in cython by default.

snorfalorpagus:
We might need to periodically check for signals: http://stackoverflow.com/a/16778026/1300519

jetuk:
This code isn't effectively pure C so won't it do this interpreter stuff as it's looping?

jetuk:
I just "fixed" this test to work. However rereading it I am not sure what it is testing?

snorfalorpagus:
It was the test I wrote when I was writing the `calc_values` function - I tested it there before putting it into Model proper. It can probably just be removed.

PR #476, opened 2017-05-12
Revised the implemented of the FlowDurationCurveDeviationRecorder

jetuk:
The recorder now supports an upper and lower target FDC. Deviation is calculated as positive above
and negative below the respective FDC. If flow is between the two zero deviation is returned.

@Batch21 could give this a read please.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -548,17 +548,23 @@ cdef class FlowDurationCurveDeviationRecorder(FlowDurationCurveRecorder):
         else:
             return self._agg_user_func(np.array(self._fdc_deviations), axis=0)
 
-    def to_dataframe(self):
-        """ Return a `pandas.DataFrame` of the recorder data
-
-        This DataFrame contains a MultiIndex for the columns with the recorder name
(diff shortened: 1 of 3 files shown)


Batch21:
It looks good. I know it's not needed currently, but would it be worth recording the direction of the deviation?

jetuk:
That's a good point. It also doesn't have a mechanism to find which percentile has the worst deviation.

jetuk:
I think this does what we want now.

pywr/recorders/_recorders.pyx
```diff
@@ -442,66 +442,90 @@ FlowDurationCurveRecorder.register()
...
+                raise ValueError("The number of upper target FDCs does not match the number of scenarios")
 
     cpdef finish(self):
         super(FlowDurationCurveDeviationRecorder, self).finish()
 
         cdef int i, j, k, sc_index
         cdef ScenarioIndex scenario_index
-        cdef double[:] trgt_fdc
+        cdef double[:] utrgt_fdc, ltrgt_fdc
+        cdef udev, ldev
```
snorfalorpagus: These should be typed double.

PR #477, opened 2017-05-15
Moved time table creation to reset() in TablesRecorder

jetuk:
This avoids a bug whereby resetting the model doesn't remove all the rows
in the time table. Subsequent appends then duplicate the entries. Time
table creation is now done in the reset method.

### DIFF
--- a/pywr/recorders/recorders.py
+++ b/pywr/recorders/recorders.py
@@ -314,14 +314,6 @@ class TablesRecorder(Recorder):
                 group_name = self.h5store.file.root
             self.h5store.file.create_carray(group_name, node_name, atom, shape, createparents=True)
 
-        # Create time table
-        if self.time is not None:
-            group_name, node_name = self.time.rsplit('/', 1)
-            if "group_name" == "/":
-                group_name = self.h5store.file.root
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
Related: is there a way to get a pandas datetime series from the time table in a nice way?

jetuk:
It is something like:
```
tbl = h5.get_node('/time')
dates = pandas.to_datetime({k: tbl.col(k) for k in ('year', 'month', 'day')})
```

pywr/recorders/recorders.py
```diff
@@ -343,15 +335,31 @@ def setup(self):
...
         self._time_table = None
+        # Create time table
+        # This is created in reset so that the table is always recreated
         if self.time is not None:
-            self._time_table = self.h5store.file.get_node(self.time)
+            group_name, node_name = self.time.rsplit('/', 1)
+            if group_name == "":
+                group_name = "/"
+            description = {c: tables.Int64Col() for c in ('year', 'month', 'day', 'index')}
+            print(group_name, node_name)
```
snorfalorpagus: Remove this.

PR #485, opened 2017-05-17
SeasonalFlowDurationCurveRecorder

Batch21:


tests/test_recorders.py
```diff
@@ -97,6 +97,29 @@ def test_fdc_recorder():
...
+def test_seasonal_fdc_recorder():
+    """
+    Test the FlowDurationCurveRecorder
+    """
+    model = load_model("timeseries4.json")
+    input = model.nodes['catchment1']
+
+    percentiles = np.linspace(20., 100., 5)
+    months = [6, 7, 8]
+    rec = SeasonalFlowDurationCurveRecorder(model, input, percentiles, months)
```
jetuk: Lets load this from JSON.

tests/models/timeseries3.csv
```diff
@@ -0,0 +1,366 @@
+Timestamp,Data
```
jetuk: Let's have a random timeseries rather than this data,

Batch21: Created a random hydrograph using Pandas to use instead

PR #488, opened 2017-05-24
Updates to storage max and initial volume

### LINKED ISSUE #453: It should be possible to define initial volume in percentage terms
I suggest a second attribute to `Storage`. We would have to have `initia_volume` default to `nan` and check the `initial_volume_pc` in that case. If both are `nan` then error.

```
s = Storage(initial_volume_pc=1.0)
assert s.initial_volume_pc == 1.0
```

jetuk:
Two features

- Allow some simple aggregated style parameters to be used as max_volume.
- Added feature to set initial storage volume as a proportion of maximum volume. Fixes #453

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -95,6 +95,7 @@ cdef class AbstractStorage(AbstractNode):
 cdef class Storage(AbstractStorage):
     cdef double _cost
     cdef double _initial_volume
+    cdef double _initial_volume_pc
     cdef double _min_volume
     cdef double _max_volume
     cdef double _level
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
Such hack.

Skim read looks OK, but I'll do a proper review when I get a chance.

jetuk:
Unfortunately I need slightly more complex parameters for some current
work. Wasn't sure how else to do it

On 24 May 2017 4:51 pm, "Joshua Arnott" <notifications@github.com> wrote:

Such hack.

Skim read looks OK, but I'll do a proper review when I get a chance.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/pywr/pywr/pull/488#issuecomment-303767251>, or mute the
thread

Issue #489, opened 2017-05-24
A smarter parameter for loading timeseries data


snorfalorpagus:
We currently have two parameters capable of loading data from an HDF5 file: `TablesArrayParameter` and `DataFrameParameter`. A hybrid of the two, with some extra features, could be useful.

Existing features:
- Fast access
- Automatic date alignment
- Resample on-the-fly
- Scenario-aware
- Works with float/int data

New features:
- Only load data for the required scenarios
- Option to read from disk as-required, rather than in setup, to reduce memory requirement?
- IndexParameter variant for those other parameters that expect an IndexParameter?

snorfalorpagus:
Following #510 it'd be good to support weekly and monthly timeseries (i.e. 52 week / 12 month year) also.

PR #490, opened 2017-05-24
Only load required data in TablesArrayParameter

snorfalorpagus:
This PR begins to address #489, by only loading data required for the sliced scenario in `TablesArrayParameter`.

TODO:

- [x] slice in time as well as scenario
- [x] support for user combinations of scenarios
- [x] apply slice to integer datasets

### DIFF
--- a/pywr/parameters/_parameters.pxd
+++ b/pywr/parameters/_parameters.pxd
@@ -78,6 +78,7 @@ cdef class TablesArrayParameter(IndexParameter):
     cdef public object where
 
     cdef int _scenario_index
+    cdef int[:] _scenario_ids
 
 cdef class IndexedArrayParameter(Parameter):
     cdef public IndexParameter index_parameter
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -290,7 +290,7 @@ cdef class TablesArrayParameter(IndexParameter):
         self.where = where
         self.scenario = scenario
 
-        # Private attributes, initialised during reset()
+        # Private attributes, initialised during setup()
         self._values_dbl = None
         self._values_int = None
         self._scenario_index = -1
@@ -309,6 +309,10 @@ cdef class TablesArrayParameter(IndexParameter):
         self.h5store = H5Store(self.h5file, None, "r")
         node = self.h5store.file.get_node(self.where, self.node)
 
+        if not node.dtype in (np.float32, np.float64, np.int8, np.int16, np.int32):
+            raise TypeError("Unexpected dtype in array: {}".format(node.dtype))
+
+        # check the shape of the data is valid
         if self.scenario is not None:
             if node.shape[1] < self.scenario.size:
                 raise RuntimeError("The length of the second dimension of the tables Node should be the same as the size of the specified Scenario.")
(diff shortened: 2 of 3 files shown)


snorfalorpagus:
This is ready for review now. It should provide a speedup without any change from the user's perspective.

It'd be good to get it merged ASAP as it will speed up my runs this week. :)

pywr/parameters/_parameters.pxd
```diff
@@ -78,6 +78,7 @@ cdef class TablesArrayParameter(IndexParameter):
     cdef public object where
 
     cdef int _scenario_index
+    cdef int[:] _scenario_ids
```
jetuk: Pedantic. I guess we should start commenting some of these attributes. I'm starting to forget what everything is!

pywr/parameters/_parameters.pyx
```diff
@@ -290,12 +290,15 @@ cdef class TablesArrayParameter(IndexParameter):
         self.where = where
         self.scenario = scenario
 
-        # Private attributes, initialised during reset()
+        # Private attributes, initialised during setup()
         self._values_dbl = None
         self._values_int = None
         self._scenario_index = -1
+        self._scenario_ids = None
```
jetuk: See above. Would be good to summarise in a comment what these are all used for in the upcoming code.

pywr/parameters/_parameters.pyx
```diff
@@ -306,50 +309,82 @@ cdef class TablesArrayParameter(IndexParameter):
...
+        # check the shape of the data is valid
         if self.scenario is not None:
-            if shape[1] < self.scenario.size:
+            if node.shape[1] < self.scenario.size:
                 raise RuntimeError("The length of the second dimension of the tables Node should be the same as the size of the specified Scenario.")
-            elif shape[1] > self.scenario.size:
+            elif node.shape[1] > self.scenario.size:
                 warnings.warn("The length of the second dimension of the tables Node is greater than the size of the specified Scenario. Not all data is being used!")
-        if shape[0] < len(self.model.timestepper):
+        if node.shape[0] < len(self.model.timestepper):
```
jetuk: We don't have the equivalent warning re the length of the model run being shorter than the data.

pywr/parameters/_parameters.pyx
```diff
@@ -306,50 +309,82 @@ cdef class TablesArrayParameter(IndexParameter):
...
+        self._values_int = None
+        if self.scenario:
+            # if possible, only load the data required
+            scenario_indices = None
+            if self.model.scenarios.user_combinations:
+                scenario_indices = set()
+                for user_combination in self.model.scenarios.user_combinations:
+                    scenario_indices.add(user_combination[self._scenario_index])
+                scenario_indices = sorted(list(scenario_indices))
+                self._scenario_ids = np.ones(self.scenario.size, dtype=np.int32)
```
jetuk: Seems a bit naughty that the entries not used in this array are left as `1`

pywr/parameters/_parameters.pyx
```diff
@@ -306,50 +309,82 @@ cdef class TablesArrayParameter(IndexParameter):
...
+                for n, i in enumerate(scenario_indices):
+                    self._scenario_ids[i] = n
+                if node.dtype in (np.float32, np.float64):
+                    self._values_dbl = node[:len(self.model.timestepper), scenario_indices]
+                else:
+                    self._values_int = node[:len(self.model.timestepper), scenario_indices]
+            elif self.scenario.slice:
+                self._scenario_ids = np.ones(self.scenario.size, dtype=np.int32)
```
jetuk: See above re entries left with `1`

pywr/parameters/_parameters.pyx
```diff
@@ -306,50 +309,82 @@ cdef class TablesArrayParameter(IndexParameter):
...
+                scenario_indices = range(*self.scenario.slice.indices(self.scenario.slice.stop))
+                for n, i in enumerate(scenario_indices):
```
jetuk: This bit could be abstracted with the same portion in the block above. It's quite readable as is though.

pywr/parameters/_parameters.pyx
jetuk: This area is set with a type of `Py_ssize_t` but its dtype is `np.int32`, is that a problem?

pywr/parameters/_parameters.pyx
```diff
@@ -306,50 +309,82 @@ cdef class TablesArrayParameter(IndexParameter):
...
+            if self._values_int is None:
+                self._values_int = node.read().astype(np.int32)
+
     cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
-        cdef int i = ts._index
-        cdef int j
+        cdef Py_ssize_t i = ts._index
+        cdef Py_ssize_t j
         if self._values_dbl is None:
             return float(self.get_index(scenario_index))
```
jetuk: Is it correct to call `get_index` rather than `index` here?

pywr/parameters/_parameters.pyx
snorfalorpagus: I agree. I'm not sure what else to do though. They can't be set to None/NaN as there isn't an integer representation. They could be set to a negative number, but that's still a valid index. I guess they could be set to `node.shape[1]` so that it would at least raise an `IndexError` if it was ever used.

snorfalorpagus: The array is being used to store array indices, but it's not possible to set the dtype to Py_ssize_t. I guess that Py_ssize_t is an int64_t on 64-bit systems. Although it's unlikely there would ever be enough columns in the dataset to require more than 32-bit indexing.

tl;dr I don't think it's a problem.

pywr/parameters/_parameters.pyx
snorfalorpagus: It was like that where I got there! I think it's actually a bug. This is actually OK, because index is calculated before value. But the other (`get_value` call in `index`) will be broken.

https://github.com/pywr/pywr/blob/e74a0bd71964fdb8fdf214ef5e7d755100b985b1/pywr/parameters/_parameters.pyx#L595-L600

This should be fixed.

pywr/parameters/_parameters.pyx
jetuk: Would `np.empty` be any better/worse?

jetuk: +1

snorfalorpagus: Worse. The array would still contain integers, just uninitialised ones. Probably zeros, but could be any integer.

jetuk: The reason where' not using a dictionary is because of speed?

jetuk: OK

jetuk:
I've tweaked this a bit. I'll merge unless anyone shouts.

PR #493, opened 2017-05-29
WIP: Hydropower example

jetuk:
If someone has a few minutes it would be good to get an opinion on this new example.

Several requests have come in for hydropower. We should have a good answer to how this is done in pywr.

### DIFF
--- a/pywr/nodes.py
+++ b/pywr/nodes.py
@@ -406,7 +406,7 @@ class Storage(with_metaclass(NodeMeta, Drawable, Connectable, _core.Storage)):
 
         if level is not None:
             level = load_parameter(model, level)
-        print(level)
+
         node.level = level
 
         return node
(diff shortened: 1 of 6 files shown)


jetuk:
This all fixes some issues with loading storage levels from JSON.

pywr/nodes.py
```diff
@@ -401,6 +404,11 @@ def load(cls, data, model):
             cost = 0.0
         node.cost = cost
 
+        if level is not None:
+            level = load_parameter(model, level)
+        print(level)
```
snorfalorpagus: Remove.

pywr/parameters/_parameters.pyx
```diff
@@ -305,7 +305,6 @@ cdef class TablesArrayParameter(IndexParameter):
 
         self.h5store = H5Store(self.h5file, None, "r")
         node = self.h5store.file.get_node(self.where, self.node)
-
```
snorfalorpagus: This will conflict with #490

jetuk:
Cool. I'll update this. 

Do you think the `HydroPowerRecorder` should be in the main library?

jetuk:
@aphurford this is the example we talked about.

jetuk:
I've rebased this against current master and fixed the comments.

PR #496, opened 2017-06-07
Improved the FDC tests by checking my aggregation functions.

jetuk:
Also the deviation test wasn't really testing anything because it was looking at the input node rather than flow after the abstraction.


jetuk:
I think this one is good to go.

PR #500, opened 2017-07-04
Routes recorder

jetuk:
- Added routes flows output to `TablesRecorder`
- Added utility function to convert this saved file into d3js format.
- Added `user_combinations` support to `TablesRecorder`
  - Requires supporting ScenarioCollection.shape when user_combinations is given.
  - Stores data in a flattened formed.
  - Adds a table of scenarios combinations stored.

### DIFF
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -58,6 +58,14 @@ else:
             return self._cy_solver.dump_mps(filename)
 
         @property
+        def save_routes_flows(self):
+            return self._cy_solver.save_routes_flows
+
+        @save_routes_flows.setter
+        def save_routes_flows(self, value):
+            self._cy_solver.save_routes_flows = value
+
+        @property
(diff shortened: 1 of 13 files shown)


snorfalorpagus:
The tests are failing because this hasn't been implemented for lpsolve yet.

Is there a down side to storing the values in `self.route_flows_arr` all of the time? Could this get quite large for big models with many scenarios?

pywr/solvers/__init__.py
```diff
@@ -58,6 +58,22 @@ def dump_mps(self, filename):
...
+
+        @property
+        def routes(self):
+            return self._cy_solver.routes
+
+        @property
+        def routes_flows_array(self):
+            return self._cy_solver.route_flows_arr
+
+        @property
```
snorfalorpagus: A minor point, but I prefer the other syntax for creating properties.

Issue #19, opened 2015-05-27
Display linear program as table


snorfalorpagus:
Excel? Qt?

jetuk:
It would be relatively trivial to dump to a file. For large problems it might become a problem - especially in dense form. 

I suggest we look at this in context of #88 and a review of the `Solver` api.

PR #502, opened 2017-07-05
Displaying linear programme as table

snorfalorpagus:
Added function to display the linear programme as a table (#19). Needs more work on the design and wrapping to make it friendly in Jupyter.

### DIFF
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -57,6 +57,9 @@ else:
         def dump_mps(self, filename):
             return self._cy_solver.dump_mps(filename)
 
+        def dump_html(self):
+            return self._cy_solver.dump_html()
+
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
Also some checks on the state of the linear programme would be good - has it been solved, etc. (or even initialised).

jetuk:
Why didn't you dump this to a dataframe and let pandas/jupyter do the HTML?

snorfalorpagus:
> Why didn't you dump this to a dataframe and let pandas/jupyter do the HTML?

The structure didn't seem to fit in the column/row format. The matrix itself does, but then there are multiple "header" rows (objective value, upper and lower bounds, etc).

jetuk:
Is this ready to review / merge?

PR #508, opened 2017-07-18
More determinism in GLPK solver

jetuk:
Some changes here:

- Use a fully qualified node name for sorting the nodes & routes in the solver. This should ensure that the order of the columns & rows in the LP is deterministic. Previously this was not the case due to, for example, `Storage` sublinks having the same name.
- Add an absolute epsilon check against (1e-8) before passing row bounds and objective coefficients to GLPK. This was following some advice on the GLPK wiki: https://en.wikibooks.org/wiki/GLPK/Modeling_tips#Close-to-zero_values
- Sorted a `set` that's used for defining the min / max flow constraints on `AggregatedNode`

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -304,6 +304,13 @@ cdef class AbstractNode:
             # apply new name
             self._name = name
 
+    property fully_qualified_name:
+        def __get__(self):
+            if self._parent is not None:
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
This all looks good. I might do a quick check to see what Cython produces for the `np.abs` calls vs abs from the math library. Let me know when you're done.

PR #517, opened 2017-09-13
The threshold in *ThresholdParameter can now be a Parameter.

### LINKED ISSUE #516: Control curve parameter based on another parameter instead of storage
I have need for an `IndexParameter` which compares a control curve (daily profile, etc) to another parameter's value rather than a `Storage` node volume. I don't think we have anything that does this currently?

snorfalorpagus:
Closes #516.

### DIFF
--- a/pywr/parameters/_thresholds.pyx
+++ b/pywr/parameters/_thresholds.pyx
@@ -21,7 +21,7 @@ cdef class AbstractThresholdParameter(IndexParameter):
 
     Parameters
     ----------
-    threshold : double
+    threshold : double or Parameter
         Threshold to compare the value of the recorder to
     values : iterable of doubles
         If the predicate evaluates False the zeroth value is returned,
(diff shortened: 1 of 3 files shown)


pywr/parameters/_thresholds.pyx
```diff
@@ -42,9 +42,14 @@ cdef class AbstractThresholdParameter(IndexParameter):
     the previous day. In this case the predicate evaluates to True.
 
     """
-    def __init__(self, model, threshold, *args,  values=None, predicate=None, **kwargs):
+    def __init__(self, model, threshold, *args, values=None, predicate=None, **kwargs):
         super(AbstractThresholdParameter, self).__init__(model, *args, **kwargs)
-        self.threshold = threshold
+        threshold = load_parameter(model, threshold)
```
snorfalorpagus: This should be done in a `load` method.

pywr/parameters/_thresholds.pyx
```diff
@@ -42,9 +42,14 @@ cdef class AbstractThresholdParameter(IndexParameter):
...
+        if isinstance(threshold, Parameter):
+            self._threshold_parameter = threshold
```
jetuk: Self should be made a parent of the threshold.

snorfalorpagus:
I've addressed the two comments.

jetuk:
This looks good. I think it is the right way around otherwise the test would fail.

PR #522, opened 2017-09-16
Added optional checksum 

### LINKED ISSUE #497: Add optional checksum for TablesParameter
If given this would checksum the file before loading the data and error if it differed. This is useful for catching corruption during transfer and is also a version control check. It might be feasible to do it on some other parameters that accept file based input, e.g. `DataFrameParameter`?

We would have to do it like this because of big file: https://stackoverflow.com/a/31278890

jetuk:
- Only supports `DataFrameParameter` and `TablesArrayParameter`.
- Multiple hashes can be given for the same file in JSON.
- Uses hashlib and reads the data in chunks.

Fixes #497

### DIFF
--- a/pywr/hashes.py
+++ b/pywr/hashes.py
@@ -1,0 +1,34 @@
+""" Generic utilities functions useful for pywr, but not specific to any of the core modules
+"""
+import io
+import hashlib
+
+
+def compute_hash(filename, algorithm='md5', chunk_size=io.DEFAULT_BUFFER_SIZE):
+    """ Compute the hash of a large file using hashlib """
+    h = hashlib.new(algorithm)
+
+    with io.open(filename, mode='rb') as fh:
+        for chunk in iter(lambda: fh.read(chunk_size), b''):
+            h.update(chunk)
+
+    return h.hexdigest()
+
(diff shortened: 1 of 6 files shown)


jetuk:
One thing this could do, for urls that are files, is to see if a corresponding checksum file (e.g. .md5) exists and use that for a comparison?

snorfalorpagus:
I like this.

> One thing this could do, for urls that are files, is to see if a corresponding checksum file (e.g. .md5) exists and use that for a comparison?

How do we know which to check for (e.g. md5 vs shaXXX)?

There are two benefits to checksums: 1) detecting corrupt files, 2) detecting out-of-date files. This approach works for 1, but not so well for 2. Whereas if the checksum is in the model JSON and the data isn't updated it will error (correctly).

jetuk:
> How do we know which to check for (e.g. md5 vs shaXXX)?

Try a few extensions? Or have a special name (e.g. "file") in the JSON to check for a file.

> Whereas if the checksum is in the model JSON and the data isn't updated it will error (correctly).

Agreed. It's also more files, etc. to fiddle with.

Issue #166, opened 2016-06-30
Model.nodes.__delitem__ doesn't remove subnodes


snorfalorpagus:
A node can be removed from a model as follows:

```
del(model.nodes["nodename"])
```

This works, unless the node is a compound node (reservoirs, gauging stations) as the subnodes are not removed.

https://github.com/pywr/pywr/blob/master/pywr/core.py#L136-L139

In order for this to work we need to also delete any node whos parent is the node being removed. In the short term, a loop that checks every node in the model and deletes it if it's parent is being removed will do. It's not 100% robust though (think multiple levels of compound node).

jetuk:
This code has moved to here:

https://github.com/pywr/pywr/blob/master/pywr/_model.pyx#L745

jetuk:
>  In the short term, a loop that checks every node in the model and deletes it if it's parent is being removed will do.

Suggest we do this for 0.5

PR #527, opened 2017-09-21
Recursive deletion of child nodes when deleting compound nodes.

snorfalorpagus:
This PR allows you to delete compound nodes from a model, including their hidden sub nodes (as suggested in #166).

e.g.

```
del(model.nodes["my reservoir"])
```

It also allows you to delete a node directly rather than by name, e.g. `del(model.nodes[my_reservoir])`. This was required, as the child nodes often don't have unique names (e.g. `[output0]`).

### DIFF
--- a/pywr/_model.pyx
+++ b/pywr/_model.pyx
@@ -743,8 +743,15 @@ class NodeIterator(object):
         raise KeyError("'{}'".format(key))
 
     def __delitem__(self, key):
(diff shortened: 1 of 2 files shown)


snorfalorpagus:
The build failures are in master also (see https://travis-ci.org/pywr/pywr/builds/277930531). This was working before, so something upstream must have broken. Not sure what.

PR #530, opened 2017-09-24
Fix to the csv recorder as it is assumed that the nodes objects are p…

khaledk2:
…rovided but when nodes names are provided (e.g. loaded from the json), it will throw an error

### DIFF
--- a/pywr/recorders/recorders.py
+++ b/pywr/recorders/recorders.py
@@ -97,8 +97,7 @@ class CSVRecorder(Recorder):
         self.scenario_index = scenario_index
         self.nodes = nodes
         self.csv_kwargs = kwargs.pop('csv_kwargs', {})
-        self.node_names = None
-
+        self._node_names = None
         self._fh = None
         self._writer = None
 
@@ -108,31 +107,17 @@ class CSVRecorder(Recorder):
         """
 
         if self.nodes is None:
-            self.node_names = sorted(self.model.nodes.keys())
-        else:
-            # the next line is commented because when the nodes names are provided (e.g. loaded from the json)
-            # it will throw an exception
-            # self.node_names = sorted(n.name for n in self.nodes)
-            nodes = []
+            self._node_names = sorted(self.model.nodes.keys())
+        else:            
             node_names = []
             for node_ in self.nodes:
                 # test if the node name is provided
                 if isinstance(node_, basestring):
-                    try:
-                        # lookup node by name
-                        nodes.append(self.model.nodes[node_])
-                        node_names.append(node_)
-                    except:
-                        raise ValueError ("node name is not in the model: {}".format(node_))
+                    # lookup node by name                        
+                    node_names.append(node_)                    
                 else:
-                    try:
-                        nodes.append(node_)
-                        node_names.append((node_.name))
-                    except:
-                        # raise an error if the object is not recognized
-                        raise ValueError("Unrecognised node object: {}".format(node_))
-            self.node_names = node_names
-            self.nodes = nodes
+                    node_names.append((node_.name))                    
+            self._node_names = node_names           
 
     def reset(self):
         import csv
@@ -142,7 +127,7 @@ class CSVRecorder(Recorder):
             self._fh = open(self.csvfile, 'w')
         self._writer = csv.writer(self._fh, **self.csv_kwargs)
         # Write header data
-        self._writer.writerow(['Datetime']+self.node_names)
+        self._writer.writerow(['Datetime']+self._node_names)
 
     def after(self):
         """
@@ -151,7 +136,7 @@ class CSVRecorder(Recorder):
         from pywr._core import Node, Storage
 
         values = [self.model.timestepper.current.datetime.isoformat()]
-        for node_name in self.node_names:
+        for node_name in self._node_names:
             node = self.model.nodes[node_name]
             if isinstance(node, Node):
                 values.append(node.flow[self.scenario_index])
--- a/pywr/recorders/recorders.py
+++ b/pywr/recorders/recorders.py
@@ -1,6 +1,7 @@
 import sys
 import numpy as np
 from functools import wraps
+from pywr._core import AbstractNode, AbstractStorage
 from ._recorders import *
 from ._thresholds import *
 from .events import *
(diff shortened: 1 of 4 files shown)


pywr/recorders/recorders.py
```diff
@@ -110,7 +110,29 @@ def setup(self):
         if self.nodes is None:
             self.node_names = sorted(self.model.nodes.keys())
         else:
-            self.node_names = sorted(n.name for n in self.nodes)
+            # the next line is commented because when the nodes names are provided (e.g. loaded from the json)
```
jetuk: Remove these comments and the commented code if it's not needed.

pywr/recorders/recorders.py
```diff
@@ -110,7 +110,29 @@ def setup(self):
...
+                        raise ValueError ("node name is not in the model: {}".format(node_))
+                else:
+                    try:
+                        nodes.append(node_)
+                        node_names.append((node_.name))
+                    except:
+                        # raise an error if the object is not recognized
+                        raise ValueError("Unrecognised node object: {}".format(node_))
+            self.node_names = node_names
+            self.nodes = nodes
```
jetuk: I'm not sure I like updating `self.nodes`. If you pass a list of strings to `__int__` then `self.nodes` is that list of strings. However after `setup` is called it is now a new list of `Node` objects. For me this is odd that this changes. It appears the remainder of the code on relies on `self.node_names` so we only need to set that here.

pywr/recorders/recorders.py
```diff
@@ -110,7 +110,29 @@ def setup(self):
...
-            self.node_names = sorted(n.name for n in self.nodes)
+            # the next line is commented because when the nodes names are provided (e.g. loaded from the json)
+            # it will throw an exception
+            # self.node_names = sorted(n.name for n in self.nodes)
+            nodes = []
+            node_names = []
+            for node_ in self.nodes:
+                # test if the node name is provided
+                if isinstance(node_, basestring):
+                    try:
```
jetuk: We can probably drop this try-except, or at least it is inconsistent with `TablesRecorder` which just lets the `KeyError` from `NodeIterator` raise.

pywr/recorders/recorders.py
```diff
@@ -110,7 +110,29 @@ def setup(self):
         if self.nodes is None:
             self.node_names = sorted(self.model.nodes.keys())
         else:
-            self.node_names = sorted(n.name for n in self.nodes)
+            # the next line is commented because when the nodes names are provided (e.g. loaded from the json)
+            # it will throw an exception
+            # self.node_names = sorted(n.name for n in self.nodes)
+            nodes = []
```
jetuk: See comment below. We probably don't need to update `self.nodes` so this local variable can go also.

pywr/recorders/recorders.py
```diff
@@ -110,7 +110,29 @@ def setup(self):
...
+            node_names = []
+            for node_ in self.nodes:
+                # test if the node name is provided
+                if isinstance(node_, basestring):
+                    try:
+                        # lookup node by name
+                        nodes.append(self.model.nodes[node_])
+                        node_names.append(node_)
+                    except:
```
jetuk: `except:` without a specific subclass is too broad.

pywr/recorders/recorders.py
```diff
@@ -110,7 +110,29 @@ def setup(self):
...
+                        raise ValueError ("node name is not in the model: {}".format(node_))
+                else:
+                    try:
+                        nodes.append(node_)
+                        node_names.append((node_.name))
+                    except:
```
jetuk: except: without a specific subclass is too broad. This is only capturing an expected `AttributeError`? We can probably let that raise and drop the try-except?

jetuk:
The build failures are due to a mistake with two recent PR. It is fixed in #531.

pywr/recorders/recorders.py
khaledk2: I agree, there is no need to update the list as it will not be used after that

pywr/recorders/recorders.py
khaledk2: I was trying to provide more information to the user when something is going wrong. But if you think the current approach is fine, I will drop try and catch

pywr/recorders/recorders.py
khaledk2: I kept them to help checking the code modifications but I will remove them

pywr/recorders/recorders.py
jetuk: Fair point, but I think you'll get the same information from the `KeyError` traceback i.e. it'll tell you whcih node was not found and that this happened in the `CSVRecorder`. Generally we haven't caught these errors internally in the other recorders.

khaledk2:
I have found another issue related to get recorders nodes from the model when they are not provided by the user, it will take all the models nodes but this will raise a valueError if the node type is not Node or Storage, e.g. AggregatedNode. 
I think it is needed to check the nodes types and add only Node and Storage nodes only. Somthing like that:
if self.nodes is None:
         for node in self.model.nodes:
                if isinstance(node, Node) or isinstance(node, Storage):
                       node_names.append(node.name)
What do you think?

pywr/recorders/recorders.py
```diff
@@ -129,7 +136,7 @@ def after(self):
         from pywr._core import Node, Storage
 
         values = [self.model.timestepper.current.datetime.isoformat()]
-        for node_name in self.node_names:
+        for node_name in self._node_names:
             node = self.model.nodes[node_name]
             if isinstance(node, Node):
```
jetuk: This should check against AbstractNode and AbstractStorage

See here for example from TablesRecorder.
https://github.com/pywr/pywr/blob/5375bb39b7cf541ee704146ffeba3ca17a6e0205/pywr/recorders/recorders.py#L458

jetuk:
With that change above and a JSON based test of this recorder this can be merged.

tests/models/CSV_Recorder.json
```diff
@@ -0,0 +1,38 @@
...
+	"recorders":  {
+        "model_out": {
+            "comment": "flow to opt node",
+            "type": "CSVRecorder",
+			"nodes": ["inpt","otpt"],
+			"csvfile": "output.csv"
+        }		
+		
+	}
+}
```
snorfalorpagus: Formatting in this is a bit odd. Run this through pretty print.

eg http://jsonprettyprint.com/

tests/test_recorders.py
```diff
@@ -399,7 +399,29 @@ def test_csv_recorder(simple_linear_model, tmpdir):
                 assert np.all((np.array([float(v) for v in row[1:]]) - 10.0) < 1e-12)
             assert expected == actual
 
-
+def test_loading_csv_recorder_from_json():
+    """
+    Test the CSV Recorder which is loaded from json
+    """
+    model = Model.load('CSV_Recorder.json')
```
snorfalorpagus: Use the helper function for loading models in tests.

PR #534, opened 2017-10-02
Renamed InterpolatedLevelParameter to InterpolatedParameter

### LINKED ISSUE #526: InterpolatedLevel parameter should be renamed
It can be used for `area` and other things. Should have a generic class of interpolated parameters.

jetuk:
Fixes #526

### DIFF
--- a/pywr/domains/groundwater.py
+++ b/pywr/domains/groundwater.py
@@ -1,5 +1,5 @@
 from ..core import Storage
-from ..parameters import InterpolatedParameter
+from ..parameters import InterpolatedVolumeParameter
 from ..parameters.groundwater import KeatingStreamFlowParameter
 
 import numbers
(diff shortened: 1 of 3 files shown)


snorfalorpagus:
We should also change the name of the "levels" property. "values"?

jetuk:
I'm not sure this is still the correct name. The parameter uses storage volumes as x. Perhaps it should be a `InterpolatedVolumeParameter`? Perhaps also the `interp1d` init should be done in setup.

snorfalorpagus:
> I'm not sure this is still the correct name. The parameter uses storage volumes as x. Perhaps it should be a InterpolatedVolumeParameter?

Yes. In theory we could have other interpolated parameters (e.g. one based on another parameter).

> Perhaps also the interp1d init should be done in setup.

Yes.

PR #537, opened 2017-10-10
Added WeeklyProfileParameter

snorfalorpagus:
This PR adds a profile parameter for weekly data (i.e. 52 weeks in a year).

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -547,6 +547,9 @@ cdef class WeeklyProfileParameter(Parameter):
         v = np.squeeze(np.array(values))
         if v.ndim != 1:
             raise ValueError("values must be 1-dimensional.")
+        if len(values) == 53:
+            values = values[:52]
+            warnings.warn("Truncating 53 week profile to 52 weeks.")
         if len(values) != 52:
             raise ValueError("52 values must be given for a weekly profile.")
         self._values = v
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -537,6 +537,34 @@ cdef class DailyProfileParameter(Parameter):
         return self._values[i]
 DailyProfileParameter.register()
 
+cdef class WeeklyProfileParameter(Parameter):
+    """Weekly profile (52-week year)
+
+    The last week of the year will have more than 7 days, as 365 / 7 is not whole.
+    """
+    def __init__(self, model, values, *args, **kwargs):
+        super(WeeklyProfileParameter, self).__init__(model, *args, **kwargs)
+        v = np.squeeze(np.array(values))
+        if v.ndim != 1:
+            raise ValueError("values must be 1-dimensional.")
+        if len(values) != 52:
+            raise ValueError("52 values must be given for a weekly profile.")
+        self._values = v
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int i = ts.dayofyear - 1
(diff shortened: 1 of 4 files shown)


snorfalorpagus:
Consider using pandas.Timestamp.week.

Also, consider using a 53 week year in the same way we have a 366 day profile. If a 52 week profile is passed, the 53 week should == 52.

pywr/parameters/_parameters.pyx
```diff
@@ -537,6 +537,34 @@ cdef class DailyProfileParameter(Parameter):
...
+
+    The last week of the year will have more than 7 days, as 365 / 7 is not whole.
+    """
+    def __init__(self, model, values, *args, **kwargs):
+        super(WeeklyProfileParameter, self).__init__(model, *args, **kwargs)
+        v = np.squeeze(np.array(values))
+        if v.ndim != 1:
+            raise ValueError("values must be 1-dimensional.")
+        if len(values) != 52:
+            raise ValueError("52 values must be given for a weekly profile.")
```
jetuk: We could accept both 52 and 53 week iterables.

pywr/parameters/_parameters.pyx
```diff
@@ -537,6 +537,34 @@ cdef class DailyProfileParameter(Parameter):
...
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int i = ts.dayofyear - 1
+        if not is_leap_year(<int>(ts._datetime.year)):
+            if i > 58: # 28th Feb
+                i += 1
+        cdef Py_ssize_t week
+        if i >= 364:
+            # last week of year is slightly longer than 7 days
+            week = 51
```
jetuk: Then this would depend on whether a 53rd value existed or not.

tests/test_parameters.py
```diff
@@ -251,6 +251,25 @@ def test_daily_profile_leap_day(model):
...
+    model.timestepper.end = "2005-05-01"
+    model.timestepper.delta = 7
+
+    values = np.arange(0, 52) ** 2 + 27.5
+
+    p = WeeklyProfileParameter.load(model, {"values": values})
+
+    @assert_rec(model, p)
+    def expected_func(timestep, scenario_index):
+        week = int(min((timestep.dayofyear - 1) // 7, 51))
```
jetuk: Perhaps make the test use the pandas implementation of `.week` to see if these are the same?

snorfalorpagus:
The way we count days/weeks differs from pandas due to handling of leap years. We call the 31st December day 366 every year, whereas pandas calls it 365 in non-leap years.

```python
>>> pandas.Timestamp("2004-03-08").week
11                                     
>>> pandas.Timestamp("2003-03-08").week
10
>>> pandas.Timestamp("2004-03-08").dayofyear
68 
>>> pandas.Timestamp("2003-03-08").dayofyear
67
```

I think we should carry this behavior from daily to weekly profiles, e.g. I think the 8th March should *always* be week 11. That what the current implementation in this PR does.

Also, I think I prefer assuming a 52 week year, rather than 53. I can see the value in accepting a 53-value profile though, as this is what something like pandas might generated. I think we should just drop the last value in this case, rather than jumping through hoops in the rest of the code.

jetuk:
> I think we should just drop the last value in this case, rather than jumping through hoops in the rest of the code.

This should raise a warning in that case.

snorfalorpagus:
>> I think we should just drop the last value in this case, rather than jumping through hoops in the rest of the code.

> This should raise a warning in that case.

Done.

PR #542, opened 2017-10-13
Added GZ/BZ2 compression for CSVRecorder. Closes #532

### LINKED ISSUE #532: Add optional compression to CSVRecorder
As title. Suggest we make use of the standard library compressions: gzip, bz2, lzma, zip.

snorfalorpagus:


snorfalorpagus:
Python 2 being a pain...

jetuk:
```
>       self._writer.writerow([u'Bananas', u'and', u'Custard'])
E       TypeError: write() argument 1 must be unicode, not str

pywr/recorders/recorders.py:169: TypeError
```

Care to explain this, because I have no idea what is going on.

snorfalorpagus:
> Care to explain this, because I have no idea what is going on.

In Python 3.x you can use `bz2.open`. This wraps the `BZ2File` in a TextIOWrapper, so you can write encoded text directly to a compressed binary file. Python 2.x doesn't provide this feature. I thought it would be easy to backport just that little bit, but it's throwing up these odd errors. I'm not sure what else to try.

jetuk:
What's wrong with my fix? It seems to easy?

jetuk:
I still don't quite understand where the unicode to binary black magic happens in 2.7? Is there just no unicode because we don't explicitly use unicode anyway? I wonder if this breaks if you set a node name to `u'my node'` and run in Python 2?

snorfalorpagus:
Good point. Can you add a test for this? Just rename one of the nodes in the CSV recorder test.

jetuk:
I've added the unicode test. It fails under Python 2.

jetuk:
Gah this is confusing me. Shouldn't we write in binary mode if we're manually encoding the names? But this seems to break the `csv.writer`. It appears to internally convert `bytes` to `str` and then fails to write to the files when opened in binary mode binary mode.

snorfalorpagus:
Hmmmmmmmmmmmmmmmm

jetuk:
I give up :(

snorfalorpagus:
We could require `bz2file` for Python 2.x? https://pypi.python.org/pypi/bz2file

Or just throw an exception. We don't actually use 2.x any more and I don't think it should hold us back.

jetuk:
I think this now works in Python 2 and 3, but Python 2 doesn't support unicode characters very well. Not sure if I can be bothered to fix that. It'll require a bit of boiler plate and use `io` to do the encoding I think.

jetuk:
The test failure is 3.4 still being built for the branch. Are we happy with this now?

snorfalorpagus:
The failure in 3.4 is not related to the changes in this PR, but rather a lack of support for 3.4 by anaconda/conda-forge.

I'll do a final review now and merge if OK.

PR #544, opened 2017-10-17
Added JSON support for license parameters.

snorfalorpagus:
This PR adds support for license parameters (e.g. `pywr.parameters.licenses.AnnualLicense`) from JSON.

### DIFF
--- a/pywr/parameters/parameters.py
+++ b/pywr/parameters/parameters.py
@@ -10,6 +10,7 @@ from ._parameters import (
     align_and_resample_dataframe, DataFrameParameter,
     IndexParameter, AggregatedParameter, AggregatedIndexParameter,
     load_parameter, load_parameter_values, load_dataframe)
+from . import licenses
 from ._polynomial import Polynomial1DParameter, Polynomial2DStorageParameter
 from ._thresholds import StorageThresholdParameter, RecorderThresholdParameter
 from past.builtins import basestring
--- a/tests/test_license.py
+++ b/tests/test_license.py
@@ -4,7 +4,9 @@ import pytest
 from datetime import datetime
 from pywr.core import Timestep, ScenarioIndex
 from pywr.parameters.licenses import License, TimestepLicense, AnnualLicense, AnnualExponentialLicense, AnnualHyperbolaLicense
+from pywr.recorders import NumpyArrayNodeRecorder
 from fixtures import simple_linear_model
+from helpers import load_model
 from numpy.testing import assert_allclose
 import numpy as np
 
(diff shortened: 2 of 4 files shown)


tests/test_license.py
```diff
@@ -63,6 +65,20 @@ def test_simple_model_with_annual_licence(simple_linear_model):
     remaining -= remaining / (365 - 2)
     assert_allclose(lic.value(m.timestepper._next, si), remaining / (365 - 3))
 
+def test_annual_license_json(solver):
+    model = load_model("annual_license.json")
+
+    model.timestepper.start = "2001-01-01"
+    model.timestepper.end = "2001-01-31"
+    model.timestepper.delta = 5
```
jetuk: Test is convenient when the time-step fits in to the year exactly. I assume this gets a little messy when that's not the case. Do the existing licence tests cover this situation?

tests/test_license.py
```diff
@@ -63,6 +65,20 @@ def test_simple_model_with_annual_licence(simple_linear_model):
...
+
+    rec = NumpyArrayNodeRecorder(model, model.nodes["supply1"])
+
+    model.run()
+
+    assert_allclose(rec.data[0], 200.0 / 365.0)
+    assert_allclose(rec.data[0], (200.0 - rec.data[0] * 5) / 360.0)
```
jetuk: These two tests are mathematically identical. Was the second one meant to test something else?
```
x = (200 - 5x) / 360
360x = 200 - 5x
x = 200 / 365
```

pywr/parameters/parameters.py
```diff
@@ -10,6 +10,7 @@
     align_and_resample_dataframe, DataFrameParameter,
     IndexParameter, AggregatedParameter, AggregatedIndexParameter,
     load_parameter, load_parameter_values, load_dataframe)
+from . import licenses
```
jetuk: Should some of the other `Parameter` groups be in submodules rather than all imported in to the `pywr.parameters` namespace? E.g. thresholds.

tests/test_license.py
snorfalorpagus: The second one was supposed to test the second timestep. This is a mistake.

tests/test_license.py
snorfalorpagus: I did think about this when I was writing this test. I'm not sure what the behavior is (or even what the correct behavior should be!)

snorfalorpagus:
I've improved the test of the annual license. The behavior at the end of the year is still undefined, but that's not worse than it was before...

snorfalorpagus:
Opened #546 re appveyor failures.

jetuk:
Do we want to merge as is with the end of year stuff? If so that should be opened as an issue for not already.

snorfalorpagus:
> Do we want to merge as is with the end of year stuff?
Yes.
> If so that should be opened as an issue for not already.
#549

PR #547, opened 2017-11-02
 Add parameter statistic tracking to EventRecorder

jetuk:
- Adds the ability for Event objects to hold a simple list of values from
   a Parameter for the duration of the event.
   - Currently only supports Parameters and not other components or nodes.
 - New EventStatisticRecorder to compute arbitrary aggregation of the
   tracked values and return via .values() and .aggregated_value()
   methods.
 - Test of the new object and functionality.

There's also another commit tacked on that adds a non-finite value sense check to TablesArrayParameter

### DIFF
--- a/pywr/recorders/events.py
+++ b/pywr/recorders/events.py
@@ -10,6 +10,7 @@ class Event(object):
         self.start = start
         self.scenario_index = scenario_index
         self.end = None
+        self.values = None  # to record any  tracked values
 
     @property
     def duration(self):
@@ -43,7 +44,7 @@ class EventRecorder(Recorder):
 
 
      """
-    def __init__(self, model, threshold, minimum_event_length=1, **kwargs):
+    def __init__(self, model, threshold, minimum_event_length=1, tracked_parameter=None, **kwargs):
         super(EventRecorder, self).__init__(model, **kwargs)
         self.threshold = threshold
         self.threshold.parents.add(self)
@@ -52,11 +53,16 @@ class EventRecorder(Recorder):
         self.minimum_event_length = minimum_event_length
         self.events = None
         self._current_events = None
+        # TODO make this more generic to track components or  nodes (e.g. storage volume)
+        self.tracked_parameter = tracked_parameter
+        if self.tracked_parameter is not None:
+            self.tracked_parameter.parents.add(self)
 
     def setup(self):
         pass
 
     def reset(self):
+        print('reset')
         self.events = []
         # This list stores if an event is current active in each scenario.
         self._current_events = [None for si in self.model.scenarios.combinations]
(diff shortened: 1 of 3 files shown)


pywr/recorders/events.py
```diff
@@ -52,11 +53,16 @@ def __init__(self, model, threshold, minimum_event_length=1, **kwargs):
...
+        # TODO make this more generic to track components or  nodes (e.g. storage volume)
+        self.tracked_parameter = tracked_parameter
+        if self.tracked_parameter is not None:
+            self.tracked_parameter.parents.add(self)
 
     def setup(self):
         pass
 
     def reset(self):
+        print('reset')
```
snorfalorpagus: Remove this.

pywr/recorders/events.py
```diff
@@ -197,3 +211,70 @@ def finish(self):
         # ... and update the internal values
         for index, row in grouped.iterrows():
             self._values[index] = row['duration']
+
+
+class EventStatisticRecorder(Recorder):
+    """ Recorder for the duration of events found by an EventRecorder
+
+    This Recorder uses the results of an EventRecorder to calculate aggregated statistics
+    of those events in each scenario. This requires the EventRecorder to be given a tracked_parameter
```
snorfalorpagus: "tracked_parameter" should be in backticks

pywr/recorders/events.py
```diff
@@ -197,3 +211,70 @@ def finish(self):
...
+    in order to save an array of values during each event. This recorder uses `event_agg_func` to aggregate
+    those saved values in each event before applying `recorder_agg_func` to those values in each scenario.
+    Aggregation by scenario is done via the pandas.DataFrame.groupby() method.
+
+    Any scenario which has no events will contain a NaN value regardless of the aggregation function defined.
+
+    Parameters
+    ----------
```
snorfalorpagus: model

pywr/recorders/events.py
```diff
@@ -197,3 +211,70 @@ def finish(self):
...
+        self._values = np.empty(len(self.model.scenarios.combinations))
+
+        if self.event_recorder.tracked_parameter is None:
+            raise ValueError('To calculate event statistics requires the parent `EventRecorder` to have a `tracked_parameter`.')
+
+    def reset(self):
+        self._values[...] = np.nan
+
+    def after(self):
+        pass
```
snorfalorpagus: Remove this method.

pywr/recorders/events.py
```diff
@@ -197,3 +211,70 @@ def finish(self):
...
+
+    def values(self):
+        return self._values
+
+    def finish(self):
+        """ Compute the aggregated value in each scenario based on the parent `EventRecorder` events """
+        events = self.event_recorder.events
+        # Return NaN if no events found
+        if len(events) == 0:
+            self._values[...] = np.nan
```
snorfalorpagus: This line shouldn't be required, as it's already set to `np.nan` in `reset`.

snorfalorpagus:
This looks good overall. Some of the `Event*` recorders are missing documentation for the `agg_` keyword arguments. Even if they're implemented by the parent class I think we need to include them as the documentation will be read in isolation.

jetuk:
Is this OK now?

PR #555, opened 2017-11-23
[WIP] platypus optimisation support.

jetuk:
Currently does the following,

- Refactors the existing inspyred model subclass to a baseclass.
- Implements a new platypus model subclass.
- Refactors the optimisation example to use either library.

We still have no tests for the `opimisation` submodule.

### DIFF
--- a/examples/two_reservoir.json
+++ b/examples/two_reservoir.json
@@ -1,0 +1,139 @@
+{
+    "metadata": {
+        "title": "Two reservoirs",
+        "description": "A simple problem with two reservoirs.",
+        "minimum_version": "0.5"
+    },
+    "timestepper": {
+        "start": "2100-01-01",
+        "end": "2110-01-01",
+        "timestep": 7
+    },
+    "nodes": [
+        {
+            "name": "catchment1",
+            "type": "catchment",
+            "max_flow": "inflow"
+        },
+        {
+            "name": "catchment2",
+            "type": "catchment",
+            "max_flow": "inflow"
+        },
+        {
+            "name": "reservoir1",
+            "type": "storage",
+            "min_volume": 3000,
+            "max_volume": 20000,
+            "initial_volume": 16000,
+            "cost": -5
+        },
+        {
+            "name": "reservoir2",
+            "type": "storage",
+            "min_volume": 3000,
+            "max_volume": 20000,
+            "initial_volume": 16000,
+            "cost": -5
+        },
+        {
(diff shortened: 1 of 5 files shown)


snorfalorpagus:
I think it should be feasible to run the examples as a test, provided we limit the number of generations and population size.

snorfalorpagus:
Can we switch from argparse to click?

I get this error trying to use platypus (which I installed via pip).

```
  File "two_reservoir_moea.py", line 201, in <module>
    platypus_main(harmonic=args.harmonic)
  File "two_reservoir_moea.py", line 44, in platypus_main
    problem = model.setup()
  File "c:\users\snorfalorpagus\desktop\pywr\pywr\optimisation\platypus.py", line 21, in setup
    p = platypus.Problem(self._variable_map[-1], len(self._objectives), len(self._constraints))
AttributeError: module 'platypus' has no attribute 'Problem'
```

The docstring at the top will need updating.

snorfalorpagus:
Looks like this needs the development version of Platypus (from GitHub). The version on PyPI is too old.

snorfalorpagus:
The current setup doesn't work with the parallel executor in Platypus as it tries to pickle the Model. I think this is a more general problem - we need to ensure the flattened array of variables maps consistently across to the model parameters in the different instances.

jetuk:
I believe the variables are sorted during caching?

I think the best way is to load the model in each instance rather than
trying to send the instance from master to all workers?

snorfalorpagus:
> I think the best way is to load the model in each instance rather than trying to send the instance from master to all workers?
The problem with this is that the current system needs to load the model in the master node to work out how many variables/objectives the problem has.

jetuk:
Why is that necessarily a problem? Otherwise we need to support serialisation of models? 

Generally you also need to be careful all the workers have the same instance loaded.

snorfalorpagus:
I think we're probably both arguing the same thing here. The point I'm trying to make is that in the current setup the evaluator method which platypus uses is a method on an instance of Model. This doesn't work when you try to use Platypus in a parallel way, because it passes the evaluator method to the subprocesses via pickle, which would require the entire model to be pickled. I have a workaround for this, but it avoids using the PlatypusOptimisationModel class completely.

jetuk:
If #601 is merged this will need updating.

snorfalorpagus:
I do not think this should be merged in its current state anyway, due to the aforementioned issues with multiprocessing.

PR #567, opened 2017-12-20
Use conda-forge lp_solve

snorfalorpagus:
lpsolve is available on conda-forge - we should use that instead


snorfalorpagus:
There are some issues upstream. Hopefully resolved soon.

I've also tried to remove the numpy version dependency from this builds. I didn't realise we hadn't already done this!

PR #571, opened 2017-12-21
Combined pyx cc

jetuk:
Combined commits from #569  and #570.


snorfalorpagus:
This looks good to me and the existing tests still pass.

jetuk:
This will need a fix to the docs once it has merged.

Issue #328, opened 2016-11-20
Automatic documentation builds


snorfalorpagus:
I've figured out how to do automatic documentation builds on Travis for Shapely. It shouldn't be too difficult to port this to Pywr. Not much point in doing it until we have some more documentation though?

jetuk:
Agreed, probably not.

snorfalorpagus:
https://github.com/drdoctr/doctr

This simplifies the whole process!

jetuk:
I'm not sure I understand how any of this actually works

snorfalorpagus:
I made some attempt at this in the doctr branch.

https://travis-ci.org/pywr/pywr/jobs/203614484

The issue is that we're building / installing / testing the module using conda-build, so it's all done in an environment. I see a few options:

* Don't use conda-build. Install the dependencies, then install pywr using setup.py and run tests as normal. Then build the docs.
* Build the docs inside conda-build. It's not clear where it would go, but needs to be after tests.
* Build/test in conda as we are. After success, activate the still-existing test environment and build the docs.

All three have merits.

jetuk:
What happened to the doctr branch?

PR #574, opened 2018-01-13
Better docs

jetuk:
This rebases some previous work to use doctr inside the conda build to publish the documentation. Documenation is only built with during python 3.6 tests and then should only deployed to gh-pages when master changes. 

The docs are now using a different theme (`sphinx_rtd_theme`) from "Read the docs". This looks better with the reference documentation I think. 

It's also using `numpydoc` which caused a bunch of issues with badly formatted docstrings. Those a fixed in here too.

Solves #328


jetuk:
That was traumatic. I think this works and I'll merge to master and we can go from there. 

Closes #328

PR #584, opened 2018-01-15
Moved the HydroPowerRecorder in to pywr.recorders.

jetuk:
This was some code in the hydrpower example. I have made it part of `pywr.recorders` so we can use it else where. Tests included.


jetuk:
I have some doubts about this. I think it might be better as `Parameter` rather than recorder.

jetuk:
Actually I've changed my mind. This should probably be two recorders,

1. Based on `NumpyArrayNodeRecorder` to keep the full timeseries, and
2. Based on `BaseConstantNodeRecorder` to keep the total value for a run.

jetuk:
I've done what I said in the previous comment. There are now two recorders both cythonised:

- `HydroPowerRecorder(NumpyArrayNodeRecorder)`
- `TotalHydroEnergyRecorder(BaseConstantNodeRecorder)`

examples/hydropower_example.json
```diff
@@ -27,10 +27,10 @@
             "max_volume": 200000,
             "initial_volume": 170000,
             "level":  {
-              "type": "interpolatedlevel",
+              "type": "interpolatedvolume",
               "node": "reservoir1",
               "volumes": [0, 25000, 50000, 75000, 100000, 150000, 200000],
-              "levels": [0, 29.2, 36.8, 42.2, 46.4, 53.1, 58.5],
+              "values": [0, 29.2, 36.8, 42.2, 46.4, 53.1, 58.5],
               "kind": "cubic"
```
Batch21: to get the example work I had to move the this parameter into the parameter section of the json.

examples/hydropower_example.json
```diff
@@ -98,7 +98,7 @@
             "storage_node": "reservoir1",
             "level": 35.0,
```
Batch21: I had to change `level` to `turbine level` and replace `storage_node` with `water_elevation_parameter`, pointing to the interpolated volume parameter

pywr/recorders/_hydropower.pyx
```diff
@@ -0,0 +1,270 @@
...
+                                             energy_unit_conversion=self.energy_unit_conversion)
+
+            self._values[scenario_index.global_id] += power * days
+
+    @classmethod
+    def load(cls, model, data):
+        from pywr.parameters import load_parameter
+        node = model._get_node_from_ref(model, data.pop("node"))
+        if "water_elevation_parameter" in data:
+            water_elevation_parameter = load_parameter(data.pop("water_elevation_parameter"))
```
Batch21: `model` needs to be the first argument to `load_parameter`

jetuk:
Thanks @Batch21. I'll fix these tomorrow. The conflict is minor; to do with the imports in `test_recorders.py`

jetuk:
I've addressed the issues. A simplified version of the example is now loaded and run as a test, but it has no values checked.

Batch21:
The updates look good. I assume there is no need to check any values in the example test as this is covered in the other tests

jetuk:
Good point. It should probably been done for completeness as that test as not assertions currently.

jetuk:
I think this is done now. The commits can probably be squashed before merging. There's a slightly messy history with merging from master, etc.

jetuk:
@Batch21 are you happy with this?

Batch21:
yeah, all looks good

PR #586, opened 2018-01-16
A few things in there

### LINKED ISSUE #380: Circular reference with group control curves
> [‎07/‎02/‎2017 12:40] Arnott, Joshua: 
> I came across an issue with a circular reference in Pywr yesterday.
> I tried to set the cost of a reservoir to a control curve parameter which used the storage from an aggregated storage which included the original reservoir.
> It didn't like that.
> [‎07/‎02/‎2017 12:40] Tomlinson, James E: 
> I had the same thing this morning
> FYI I came across it ... and worked around it
> [‎07/‎02/‎2017 12:41] Arnott, Joshua: 
> I just put the cost on the WTW instead.
> [‎07/‎02/‎2017 12:41] Tomlinson, James E: 
> If you set it to the node directly it works
> [‎07/‎02/‎2017 12:41] Arnott, Joshua: 
> In Python-space?
> [‎07/‎02/‎2017 12:42] Tomlinson, James E: 
> In JSON
> [‎07/‎02/‎2017 12:42] Arnott, Joshua: 
> You mean not in the parameters section?
> [‎07/‎02/‎2017 12:42] Tomlinson, James E: 
> Yes

### LINKED ISSUE #585: Storage node does not load area parameter from JSON
`Storage` node does not appear to load the "area" parameter passed to it via JSON - see code for level.

https://github.com/pywr/pywr/blob/7b8515fb048a2a6e1900d60251e409a95da3c32d/tests/models/reservoir_evaporation.json#L18

https://github.com/pywr/pywr/blob/7b8515fb048a2a6e1900d60251e409a95da3c32d/pywr/nodes.py#L377

This doesn't actually break the behavior of the model, but is still a bug.

jetuk:
I *think* this solves the circular reference problems. I went down a rabbit hole fixing the `area` loading problem (which had an unrealised circular reference problem in the existing test). This has fixed a few things as follows,

- Load nodes before recorders and parameters from JSON
- Load all parameters after creating the instance when loading Storage
- `load_recorder` consumes JSON data from `_recorders_to_load`.
- Load the area parameter correctly in `Storage`.
- Includes a circular control curve test that now works.

Fixes #380.
Fixes #585.

### DIFF
--- a/pywr/_model.pyx
+++ b/pywr/_model.pyx
@@ -363,9 +363,19 @@ class Model(object):
                     recorders_to_load[key]["name"] = key
         model._recorders_to_load = recorders_to_load
 
+        # load the remaining nodes
+        for node_name in list(nodes_to_load.keys()):
+            node = cls._get_node_from_ref(model, node_name)
+
         # load parameters and recorders
-        for name, rdata in model._recorders_to_load.items():
-            load_recorder(model, rdata)
+        while True:
+            try:
+                name, rdata = model._recorders_to_load.popitem()
+            except KeyError:
+                break
+            recorder = load_recorder(model, rdata)
+
+
         while True:
             try:
                 name, pdata = model._parameters_to_load.popitem()
@@ -374,10 +384,6 @@ class Model(object):
             parameter = load_parameter(model, pdata, name)
             if not isinstance(parameter, BaseParameter):
                 raise TypeError("Named parameters cannot be literal values. Use type \"constant\" instead.")
-
-        # load the remaining nodes
-        for node_name in list(nodes_to_load.keys()):
-            node = cls._get_node_from_ref(model, node_name)
 
         del(model._recorders_to_load)
         del(model._parameters_to_load)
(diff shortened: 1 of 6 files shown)


jetuk:
Hmm I don't really see why this fails in 3.4. Appveyor doesn't run 3.4 which is why it passes I guess.

jetuk:
OK I figured this out. It's the old random ordering of dictionaries. 

Interestingly it's apparently not random in 3.6.1+ and will become part of the standard 3.7. However it is still random in 3.4 and the circular error reoccurs if the nodes are loaded in the "wrong" order.

jetuk:
I'm not convinced this works very well in all cases. There's definitely a regression when loading dataframes from `pandas`. This code passes all of `data` to the `pandas.read_xxx` functions. However the refactor means this can now contain "name" keys which aren't recognised. I think it was probably broken before you passed a "comment" key in some cases. 

Anyway I think this needs some more tests, and probably testing on the bigger models first.

I'm going to cherry pick the #380 commit and the one to solve the area loading problem.

PR #592, opened 2018-01-29
Improved the mean flow recorders.

jetuk:
- Current MeanFlowRecorder renamed to RollingMeanFlowNodeRecorder
  - It is only a rolling mean recorder, and it is meant for nodes.
  - This is a better overall description.
- Added a MeanFlowNodeRecorder that is a simple mean of flow through an
  entire model run.
  - This doesn't give instantaneous mean flow part way through a run
    however.
- Added both to the docs.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -1035,7 +1035,6 @@ cdef class MeanFlowNodeRecorder(BaseConstantNodeRecorder):
     cpdef after(self):
         cdef ScenarioIndex scenario_index
         cdef int i
-        cdef int days = self.model.timestepper.current.days
         for scenario_index in self.model.scenarios.combinations:
             i = scenario_index.global_id
             self._values[i] += self._node._flow[i]*self.factor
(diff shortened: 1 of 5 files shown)


pywr/recorders/_recorders.pyx
```diff
@@ -1022,6 +1022,33 @@ cdef class TotalFlowNodeRecorder(BaseConstantNodeRecorder):
...
+    A factor can be provided to scale the total flow (e.g. for calculating operational costs).
+    """
+    def __init__(self, *args, **kwargs):
+        self.factor = kwargs.pop('factor', 1.0)
+        super(MeanFlowNodeRecorder, self).__init__(*args, **kwargs)
+
+    cpdef after(self):
+        cdef ScenarioIndex scenario_index
+        cdef int i
+        cdef int days = self.model.timestepper.current.days
```
Batch21: doesn't look like the `days` variable is used anywhere

Batch21:
Had a play with the old and new recorders and the updates look good to me

pywr/recorders/_recorders.pyx
jetuk: Fixed.

Issue #258, opened 2016-09-21
Provide a mechanism to retrieve variable values from Parameters


jetuk:
Currently it is possible to use `Parameter.update` with an array of values to update during an optimisation. However there is no methods to retrieve those values in a generic way. Should we implement that? E.g. `Parameter.variables()` though I'm not sure about the name?

snorfalorpagus:
I think this is a probably we have in a few places, whereby once parameters are created it's not possible to modify/view their attributes.

For this specific issue, yes I think it's reasonable that we provide a function that does this. Perhaps rename to `Parameter.get_variables()` and `Parameter.set_variables()`? You could provide a property `variables` also, although I suspect as an API it's probably useful having the functions too.

jetuk:
One thing I have realised is that some parameters would need `int` variables instead of `double`. Because you can't overload the methods in Cython this needs two different API interfaces.

snorfalorpagus:
How do we have parameters that have a mixture? I can't think of an example, but it seems plausible.

jetuk:
Generally a single parameter is one type, but I guess it doesn't have to be. This has a risen with my working a binning scenarios where the bin a scenario is in is an integer value that is updated by an external optimiser.

jetuk:
I started this at the weekend. I've gone for

Class attribute:
```
Parameter.__can_be_variable__
```

For real variables I have used,
```
Parameter.size
Parameter.get_variables()
Parameter.set_variables()
```

For integers I propose,

```
Parameter.integer_size
Parameter.get_integer_variables()
Parameter.set_integer_variables()
```

jetuk:
We agreed yesterday this should be:

Class attribute:
```
Parameter.__can_be_variable__
```

For real variables I have used,
```
Parameter.double_size
Parameter.get_double_variables()
Parameter.set_double_variables()
```

For integers I propose,

```
Parameter.integer_size
Parameter.get_integer_variables()
Parameter.set_integer_variables()
```

jetuk:
I can't find the work I did on this previously. However, I've been thinking about it again. I've recently been working on a `Parameter` that has a date value which becomes the variable component. In the current framework this has been encoded as an integer to select from a list of feasible dates. I think that would work fine as an integer variable in the API above.

However, I think there is a need for another API method that can return the variable values in a `Recorder` specific way. This might return a `datetime` or a date string in the example above. For the `AnnualHarmonicSeriesParameter` it might return a dict like,

```python
[...]

PR #601, opened 2018-03-06
New and improved variable API for Parameters.

jetuk:
This commit changes the API for getting and setting Parameter variables. It implements the changes discussed in #258. The changes have been made to all existing Parameters that supported the old API.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -22,6 +22,8 @@ cdef class Parameter(Component):
     def __init__(self, *args, is_variable=False, **kwargs):
         super(Parameter, self).__init__(*args, **kwargs)
         self.is_variable = is_variable
+        self.double_size = 0
+        self.integer_size = 0
 
     @classmethod
     def register(cls):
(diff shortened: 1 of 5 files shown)


snorfalorpagus:
I will review this.

tests/test_parameters.py
```diff
@@ -32,6 +32,59 @@ def model(solver):
...
+class TestConstantParameter:
+    """ Tests for `ConstantParameter` """
+    def test_basic_use(self, simple_linear_model):
+        """ Test the basic use of `ConstantParameter` using the Python API """
+        model = simple_linear_model
+        # Add two scenarios
+        scA = Scenario(model, 'Scenario A', size=2)
+        scB = Scenario(model, 'Scenario B', size=5)
+
+        p = ConstantParameter(model, np.pi, name='pi', comment='Mmmmm Pi!')
```
snorfalorpagus: lolz

pywr/parameters/_parameters.pyx
```diff
@@ -802,6 +847,8 @@ cdef class AnnualHarmonicSeriesParameter(Parameter):
         self._phase_lower_bounds = np.ones(n)*kwargs.pop('phase_lower_bounds', 0.0)
         self._phase_upper_bounds = np.ones(n)*kwargs.pop('phase_upper_bounds', np.pi*2)
         super(AnnualHarmonicSeriesParameter, self).__init__(model, *args, **kwargs)
+        # Size must be set after call to super.
+        self.size = 1 + 2*n
```
snorfalorpagus: Should this be using self.double_size?

jetuk: I've fixed this. There was also another reference in `ConstantParameter`.

jetuk:
@Batch21 any comment on this?

Batch21:
I had a look through this a while ago but didn't have any specific comments. Would you like me to have a closer look?

jetuk:
If you wouldn't mind. We should get some of these PR merged or closed.

Batch21:
I started looking at this again last night but got sidetracked into looking at the MOEA stuff more broadly. I'll try and do a full review in the next few days.

jetuk:
Thanks. This is also related to #610. Having looked back at #258 this PR does not do the last thing by providing a JSON like API: `get_variables`, `set_variables`.

Batch21:
I've had a detailed look through this and played around with it and the MOEA example. I don't have any specific comments on the code - it all works for me.

jetuk:
Thanks. We'll have to decide whether to do that last thing in #258 as per the previous comment. What do you think?

Batch21:
I can see why it would be nice to have. Would it replace the existing methods for getting and setting variables or would you then be able to use both dicts and arrays?

jetuk:
I think both would be useful. The array based functions are fast. I was mostly thinking the dict output would be useful for dumping the current state to JSON in a meaningful way.

Batch21:
yeah ok. I guess this would help with implementing #602

jetuk:
I am going to merge this, but leave #258 open for the remaining API as discussed here.

PR #603, opened 2018-03-12
Added flow weights to AggregatedNode constraints.

jetuk:


jetuk:
I think these failures are just appveyor being annoying. I don't have permission to restart the jobs.

snorfalorpagus:
Please can you:

* Add a description of flow_weights in the relevant docstring (AggregatedNode)
* Add a test whereby the flow from B is non-zero but the total flow is still constrained.

jetuk:
I've opened #608 for the docstring issue. There's a deeper problem with the docstrings and init methods that needs sorting out. I'll do that in a separate PR for all nodes rather than just AggregatedNode here.

jetuk:
@snorfalorpagus are you happy with this now?

jetuk:
@Batch21 any chance you can give this one a quick look?

Batch21:
I've had a look at this. It was a bit confusing at first at how the weights and factors related to each other, so it'd definitely be good to sort out the docstrings for the other issue. Might be worth adding a test where both factors and weights are set on the agg node (would this actually be done in practice?).

jetuk:
I think the test that uses `aggregated1.json` is an example of both factors and weights being set?

Batch21:
Didn't spot that test. The other thing I was wondering about was how pytest deals with the different solvers? Does it run a test for both lpsolve and glpk if both are available or does it default to one. I can see that some of tests are skipped based on the solver type.

jetuk:
There's a custom option to pytest that allows you to pass the solver. It should patch that as the solver to use, but requires you to make the model via the fixtures and helper functions. Both solvers are used in the CI script (see below). If you look at a travis output you will see it goes through them twice.

```
py.test -v ${SRC_DIR}/tests --solver=glpk
py.test -v ${SRC_DIR}/tests --solver=lpsolve
```
https://github.com/pywr/pywr/blob/master/conda-recipe/run_test.sh

Batch21:
ok cool, thanks for the explanation. Otherwise this all looks good.

PR #610, opened 2018-03-21
Add platypus and pygmo optimisation support.

jetuk:
Platypus refactor to allow pickling and use with ProcessPools.

The refactored approach subclasses the platypus.Problem and creates a module
level cache of the Model for sub-processes to use. These module level values
are accessed through properties on the PywrPlatypusProblem.

Also added tests for running an algorithm with a ProcessPoolEvaluator.


jetuk:
Oh this requires platypus adding to the recipe, but it's a version from master.

jetuk:
This is really waiting on https://github.com/conda-forge/staged-recipes/pull/5650 to be merged so we can do CI with platypus.

jetuk:
I've done some work on this PR. It's been rebased against master to work with the new variable API.

There is now initial support for both `platypus` and `pygmo`. Both use a the same global cache approach to avoid pickling problems in multiprocessing environments. Both are also given limited tests which are conditional on the packages being installed. Currently `pygmo` is added as a test requirement to the conda build recipe, but `platypus` is not. See above comment re conda-forge request for `platypus`. However, CI should pass by skipping the `platypus` tests.

The `pygmo` wrapper still needs to do some things,
- [ ] Expose constraints
- [ ] Not assume all objectives are minimising

jetuk:
Failures are due to pygmo only working on Python >= 3.4.

More reasons to ditch 2.7

pywr/optimisation/platypus.py
```diff
@@ -0,0 +1,85 @@
...
+        """ Setup the constraints. """
+        # Setup the constraints
+        self.problem.constraints[:] = "<=0"
+
+    def evaluate(self, solution):
+        logger.info('Evaluating solution ...')
+
+        for ivar, var in enumerate(self.model_variables):
+            j = slice(self.model_variable_map[ivar], self.model_variable_map[ivar+1])
+            var.update(np.array(solution[j]))
```
jetuk: This is old.

jetuk:
@snorfalorpagus any chance you can check the OSX issue with process pools? I don't know if it is a quirk of Travis or something more fundamental.

snorfalorpagus:
I'll have a look. Hopefully I can reproduce locally as Travis doesn't give much to go on.

snorfalorpagus:
With the current branch the tests pass in my macOS environment. If you look at the latest Travis build it's the Linux branch that fails!

pywr/optimisation/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+def cache_objectives(model):
+    # This is done to make sure the order is fixed during optimisation.
+    objectives = []
+    for r in model.objectives:
+        objectives.append(r)
+    return objectives
+
+
+# Global variables for individual process to cache the model and some of its data.
+# TODO make this a registry using a unique id for each model. This way multiple
```
snorfalorpagus: TODO

jetuk: What unique ID should we use? I have been thinking of adding a UID property to metadata using `uuid`. Though this would perhaps change upon `Model.reset`. Otherwise it could be model's name?

snorfalorpagus: I like the idea of using a UUID (`uuid.uuid4()`) which is initalised in the `__init__` method of `BaseOptimisationWrapper`. The global variables then become dictionaries.

examples/two_reservoir_moea.py
```diff
@@ -10,92 +10,127 @@
 this --plot can be used to generate an animation and PNG of the pareto frontier.
```
Batch21: This text could be updated to include info about the platypus and pygmo options

examples/two_reservoir_moea.py
```diff
@@ -220,6 +258,7 @@ def update_line(gen, dfs, ax, xmax, ymax):
     parser = argparse.ArgumentParser()
     parser.add_argument('--harmonic', action='store_true', help='Use an harmonic control curve.')
     parser.add_argument('--plot', action='store_true', help='Plot the pareto frontier.')
+    parser.add_argument('--library', type=str, choices=['inspyred', 'platypus', 'pygmo'])
     args = parser.parse_args()
 
     if args.plot:
```
Batch21: plot arg only works if inspyred has been used to generate results

pywr/optimisation/inspyred.py
```diff
@@ -1,48 +1,17 @@
 import numpy as np
 import inspyred
-from ..core import Model
+from . import BaseOptimisationWrapper
 
 
-class InspyredOptimisationModel(Model):
+class InspyredOptimisationWrapper(BaseOptimisationWrapper):
     """ A pywr.core.Model subclass to enable optimisation using inspyred.
```
Batch21: not a subclass of model anymore

pywr/optimisation/inspyred.py
```diff
@@ -68,8 +37,8 @@ def generator(self, random, args):
     def evaluator(self, candidates, args):
         fitness = []
         for i, candidate in enumerate(candidates):
```
Batch21: `i` is not used

Batch21:
Are there any clear advantages/disadvantages to choosing any of the optimisation libraries? If so, it might be useful to document these somewhere.

Probably a separate issue, but it would also be useful to have an optimisation section in the cookbook section of the docs. I'm happy to have a go at doing this

jetuk:
I'm not sure we can provide too much guidance. The different libraries have different algorithms and implementations. There will be problem specific considerations and also deployment concerns that make you prefer one library over another.

However, I think we can have some general guidance on how it works. There is the optimisation example which would be a good tutorial for the docs and perhaps the place to start.

jetuk:
@snorfalorpagus @Batch21 I think I've addressed both your comments on this one.

Batch21:
All looks good to me

jetuk:
I'm going to merge this today unless anyone shouts.

snorfalorpagus:
I'm fine with this, although I haven't looked at it since my previous comment(s). You can leave the history as-is and do a normal merge.

jetuk:
Thanks.

PR #611, opened 2018-03-29
WIP: Initial attempt at providing marshmallow schemas for nodes.

jetuk:
I've tried this approach for adding `marshmallow` schemas to nodes. What are your thoughts?

Currently it only does basic validation and most of the fields are `Raw` type which does nothing special. It's mostly checking the incoming data has the correct fields.

### DIFF
--- a/pywr/schema/__init__.py
+++ b/pywr/schema/__init__.py
@@ -1,0 +1,13 @@
+from . import nodes
+
+
+def validate_with(validation_klass):
+    """ Decorator for """
+    def validation_decorator(klass):
+        klass.schema = validation_klass()
+        return klass
+    return validation_decorator
+
+
+
+
--- a/pywr/schema/nodes.py
+++ b/pywr/schema/nodes.py
@@ -8,6 +8,7 @@ class BaseNodeSchema(marshmallow.Schema):
     name = marshmallow.fields.Str()
     type = marshmallow.fields.Str()
     comment = marshmallow.fields.Str()
+    position = marshmallow.fields.Dict()
 
     subclasses = {}
 
@@ -18,12 +19,12 @@ class BaseNodeSchema(marshmallow.Schema):
         cls.subclasses[name] = cls
 
     @classmethod
-    def get_schema(cls, node_type, model, node_to_load_cls=None):
+    def get_schema(cls, node_type, model, node_cls_to_load=None):
         from ..nodes import NodeMeta, Node, Storage
 
         node_cls = NodeMeta.node_registry[node_type]
-        if node_to_load_cls is None:
-            node_to_load_cls = node_cls
+        if node_cls_to_load is None:
+            node_cls_to_load = node_cls
 
         try:
             schema_cls = cls.subclasses[node_type+'schema']
@@ -34,15 +35,15 @@ class BaseNodeSchema(marshmallow.Schema):
                 schema_cls = None
 
         if schema_cls is None:
-            # No schema found try the parents
+            # No schema found. Try the parent's schema instead
             for parent_cls in node_cls.__bases__:
                 if issubclass(parent_cls, (Node, Storage)):
                     return cls.get_schema(parent_cls.__name__.lower(), model,
-                                          node_to_load_cls=node_to_load_cls)
+                                          node_cls_to_load=node_cls_to_load)
             else:
                 raise ValueError('No schema found for node class: "{}"'.format(node_cls))
         else:
-            return schema_cls(context={'model': model, 'cls': node_to_load_cls})
+            return schema_cls(context={'model': model, 'cls': node_cls_to_load})
 
     @marshmallow.validates_schema(pass_original=True)
     def check_unknown_fields(self, data, original_data):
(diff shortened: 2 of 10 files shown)


Batch21:
I had a look through this. Is the ultimate aim of using Marshmallow to allow the whole model to be pickled?

jetuk:
@Batch21 @snorfalorpagus any thoughts on this? It's quite a big change / addition. If agreed it would be done for all the component subclasses too (essentially anything with a load method).

jetuk:
Yes. It makes writing a new `Parameter` more difficult for example.

Batch21:
yeah, though as more parameters are added it is becoming less likely that someone will need to create a custom one.  I think this is only a minor issue that can be mostly mitigated by good documentation

jetuk:
Another option with this is to put the schemas in a separate module. Use a registry and naming convention such that they can be looked up as required on load, but are optional. I.e. if the schema is there it is used for validation, but if it is not it simply raises a warning about a missing validation schema?

jetuk:
@Batch21 any opinion on putting these schemas in their own module. We could then decorate the `load` method with the schema. Something like,

```
@classmethod
@validate_with(schema.Catchment)
def load(cls, data, model):
   ....
```
This way there is a very small change to the node's code (i.e. no internal class). Alternatively we could just search for a `CatchmentSchema` class in a registry of schemas and use it if it exists.

Batch21:
I think decorators would be a nicer way of doing it than having internal classes. It also makes sense to me to separate out the schema code into their own module

jetuk:
On reflection we can't decorate `load` because several subclasses rely on the parent implementation. It would require a class decorate (which makes sense for a future `dump` method too).

jetuk:
OK... something more controversial ... how about the "loading" (i.e. current contents of `load`) method moves to the schemas?

jetuk:
I've updated this (rebased against master also). See the last commit message for more detail. I prefer this refactor due to less interference of `marshmallow` with the "main" code. However, it does introduce some magic class name lookup between the `Node` classes and the `Schema` classes.

jetuk:
Can this be the commit which breaks Python 2 and < 3.6 compatibility?

Batch21:
I think its good to have it as a subpackage. However if an attribute is set after a node has already been loaded then the new/updated attribute won't be validated?

jetuk:
This is true, but also true of the previous approach. Marshmallow will only do checking on (de)serialisation. It's up to the respective classes to behave sanely while in memory.

Any thoughts on the requirement for Python >= 3.6?

Batch21:
I think it should be fine. Its not that long now before some of the dependencies force it and there can't be much benefit in supporting 2.7 up till that point. I'm not actually aware of any examples where 2.7 has been used.

jetuk:
OK. I will do a separate PR to ditch Python < 3.6 from CI and update README etc.

Do you think we can merge this with just the nodes in the schema? Or would you prefer adding the parameters and recorders first too?

jetuk:
Dropping python < 3.6 is now merged (#623) in master, and then merged in here.

jetuk:
If this is accepted as a change / idea the main outstanding item is to extend the approach to (i.e. write schemas for) `Parameter` and `Recorder` subclasses.

pywr/schema/__init__.py
```diff
@@ -0,0 +1,13 @@
+from . import nodes
+
+
+def validate_with(validation_klass):
```
Batch21: Is this used anymore?

Batch21:
Gone through this again and it looks good. Is the plan to add schemas for parameters and recorders after the next release?

jetuk:
Yes I think so. I haven't done any work on that yet.

pywr/schema/nodes.py
```diff
@@ -0,0 +1,168 @@
...
+            schema_cls = cls.subclasses[node_type+'schema']
+        except KeyError:
+            try:
+                schema_cls = cls.subclasses[node_type]
+            except KeyError:
+                schema_cls = None
+
+        if schema_cls is None:
+            # No schema found. Try the parent's schema instead
+            for parent_cls in node_cls.__bases__:
```
snorfalorpagus: This is not nice.

Batch21:
Will the schema still be validated when a model is loaded? Could there be a method something like this for each component:

```python
def validate(self):
    attributes = {"mean": self.mean, "amplitude": self.amplitude, "phases": self.amplitude}
    errors = self.schema().validate(attributes)
    return errors
```

A user could then choose whether to validate with marshmallow or not

snorfalorpagus:
> A user could then choose whether to validate with marshmallow or not
When would the user not want to validate? I was assuming that loading would implicitly validate.

Batch21:
I guess there is no real reason not to validate. Was mainly wondering how/where the validation would be triggered as it doesn't look like this occurs just from including the Schema as an inner class.

jetuk:
I think the plan is to have a @post_load method on a `BaseSchema` class that simply invokes the schema's parent class's `__init__` method. We could then remove each classes `load` classmethod as all the type conversion (e.g. `load_parameter` and getting a node by ref) would be done by the marshmallow fields.

Overall this would be a much tighter integration of marshmallow into the loading procedure than what is currently done here.

jetuk:
FYI my approach here is going to be to revert to https://github.com/pywr/pywr/pull/611/commits/1844550fe85af08c1ec159535c1184c4033417fc and branch from there. I'll close this PR and open a new one when that is done.

PR #612, opened 2018-03-31
WIP: adding functionality to notebook graph function labels

Batch21:
As I’ve been looking at pywr recently, I’ve added a couple of extra options to the notebook graph function. `Label=True` will label all nodes at once and override the exiting name-on-hover functionality. `Attributes=True` will extract attribute information for each node and display it as a table when the node is hovered over. I think the second one needs a bit more work, particularly regarding what attributes should be included in the table.


pywr/notebook/__init__.py
```diff
@@ -16,12 +19,14 @@
 with open(os.path.join(folder, "graph.css"), "r") as f:
     draw_graph_css = f.read()
 
-def pywr_model_to_d3_json(model):
+def pywr_model_to_d3_json(model, attributes):
```
jetuk: Should this default to `attributes=False` to retain previous functionality?

pywr/notebook/__init__.py
```diff
@@ -16,12 +19,14 @@
...
     """Convert a Pywr graph to a structure d3 can display"""
```
jetuk: It would be good to add the new kwarg to the docstring.

pywr/notebook/__init__.py
```diff
@@ -57,19 +61,60 @@ def pywr_model_to_d3_json(model):
...
 
+
     return graph
 
+def get_node_attr(node):
+    """
+    Returns a dictionary that contains node attributes
+    """
+    attrs = inspect.getmembers(node, lambda a:not(inspect.isroutine(a)))
+    filtered_attrs = [attr for attr in attrs if attr[0][0] != "_"]
```
jetuk: Could this filtering be done in the loop below using something like,

```python
if att_name.startswith('_'):
    continue
```

pywr/notebook/__init__.py
```diff
@@ -57,19 +61,60 @@ def pywr_model_to_d3_json(model):
...
+        att_name, att_val = att
+        att_type = type(att_val).__name__ 
+        
+        attrs_to_skip = ["component_attrs", "components", "color", "model",  "input", "output",
+                         "inputs", "outputs", "sub_domain", "sub_output", "sublinks", "visible",
+                         "fully_qualified_name", "allow_isolated"]
+        if not att_val or att_name.lower() in attrs_to_skip:
+            continue    
+        
+        if isinstance(att_val, Component):
```
jetuk: These should only ever be `Parameters` I think.

pywr/notebook/__init__.py
```diff
@@ -57,19 +61,60 @@ def pywr_model_to_d3_json(model):
...
+                         "fully_qualified_name", "allow_isolated"]
+        if not att_val or att_name.lower() in attrs_to_skip:
+            continue    
+        
+        if isinstance(att_val, Component): 
+            att_val = att_val.name
+            if not att_val:
+                att_val = ""
+                
+        if isinstance(att_val, list):
```
jetuk: Why don't you want the list of as a list?

pywr/notebook/__init__.py
```diff
@@ -124,7 +169,7 @@ def create_node_class_trees():
     return node_class_trees
 
 
-def draw_graph(model, width=500, height=400, css=None):
+def draw_graph(model, width=500, height=400, labels=False, attributes=False, css=None):
```
jetuk: Add new kwargs to docstring.

pywr/notebook/draw_graph.js
```diff
@@ -149,10 +163,47 @@ function tick() {
...
+
+{% if attributes %}
+    nodes.on("mouseover", function(d){
+        
+        var table  = d3.selectAll(element)
+                    .append("table")
+                    .classed("table-tooltip", true);
+
+        var thead = table.append('thead')
+        var	tbody = table.append('tbody');
```
jetuk: Random white space.

Batch21:
Thanks for the review - I've started updating the code. Looking through the issues, I can't see any directly related to the notebook stuff. Is there anything else worth doing in this area while I'm looking at it.

jetuk:
You could add something to the documentation and/or add the methods to the
API reference (will need tidy docstrings).

Also update the changelog about this feature.

jetuk:
What's left on this one?

Batch21:
I have not yet added anything to the documentation or any tests.

For the changelog, I created a v0.6 section - is this correct?

jetuk:
I intended it to go under the master heading until v0.6 is released. I'l make that edit quick. We can then add the v0.6 header when it is released.

jetuk:
Done.

Have you managed to build the docs locally before?

Batch21:
I have when I went through it with you a couple of months ago. When I've tried to do it since I've got an error, which I haven't solved yet.

jetuk:
I'm not sure if I've done it on Windows. FYI the travis python 3.6 test on Linux tests building the docs so you can see the commands that are done in that console output. It'll also mean the CI will fail if you break the docs.

Batch21:
I've added a notebook section to the api docs. I was not sure whether to include all the functions or just those that will be directly used in a notebook.

jetuk:
@Batch21 docs build OK here and on CI. My only picky comment is to fix the docstring grammar so they have captial letters and fullstops in the docs. I'm off line now for a couple of days as I move house. I don't mind fixing the grammar later this week if you're also busy.

 For example:
```
converts a json file or a json-derived dict into structure that D3 can use
```

Batch21:
I'll have a go at fixing the grammar this evening and add a couple of tests. Good luck with the move

Batch21:
I've added a couple of simple tests for this. I see there is a notebook in the tests folder but I can't see how it is used by pytest.

jetuk:
It's not run by pytest, but it is used in the conda recipe which is run by the CI. See here, 

https://github.com/pywr/pywr/blob/master/conda-recipe/run_test.sh

You can test this locally by building the recipe locally, or just by just running the same command.

pywr/notebook/__init__.py
```diff
@@ -153,13 +153,25 @@ def pywr_json_to_d3_json(model):
...
             pass
 
+        if attributes:
+            json_node["attributes"] = []
+            for name, val in node.items():
+                val_type = " "
+                if val in params or isinstance(val, dict):
+                    val_type = "parameter"
+                if isinstance(val, dict):
+                    val = "Parameter defined inplace"
```
jetuk: Why do this overwrite the `val`? I'm confused.

pywr/notebook/__init__.py
```diff
@@ -153,13 +153,25 @@ def pywr_json_to_d3_json(model):
 
     nodes = []
     node_classes = create_node_class_trees()
+    params = [param for param in model["parameters"]]
```
jetuk: `params = list(model["parameters"].keys())`

pywr/notebook/__init__.py
```diff
@@ -153,13 +153,25 @@ def pywr_json_to_d3_json(model):
...
 
+        if attributes:
+            json_node["attributes"] = []
+            for name, val in node.items():
+                val_type = " "
+                if val in params or isinstance(val, dict):
+                    val_type = "parameter"
+                if isinstance(val, dict):
+                    val = "Parameter defined inplace"
+                attr_dict = {"Attribute": name, "Type": val_type, "Value": val}
```
jetuk: Why capitalised?

pywr/notebook/draw_graph.js
```diff
@@ -187,7 +187,7 @@ force.start();
             .data(columns).enter()
             .append('th')
             .text(function (column) { return column; });
-            
+        console.log(d)
```
jetuk: Remove.

pywr/notebook/__init__.py
```diff
@@ -153,13 +153,25 @@ def pywr_json_to_d3_json(model):
...
+        json_node = {'name': node.pop("name"), 'clss': node_classes[node["type"].lower()]}
         try:
             json_node['position'] = node['position']['schematic']
         except KeyError:
             pass
 
+        if attributes:
+            json_node["attributes"] = []
+            for name, val in node.items():
+                val_type = " "
```
jetuk: Perhaps this should be `None` by default.

Batch21: I've used an empty string as `None` causes an error when loading the data in js

jetuk:
Is this still WIP?

Batch21:
not really - I don't plan to add anything to it currently. I want to spend some time looking at the water balance issue.

pywr/notebook/__init__.py
```diff
@@ -57,23 +69,76 @@ def pywr_model_to_d3_json(model):
...
+    attrs = inspect.getmembers(node, lambda a:not(inspect.isroutine(a)))
+    attribute_data = []
+    for att in attrs:
+        
+        attr_name, attr_val = att
+        if attr_name.startswith("_"):
+            continue
+        attr_type = type(attr_val).__name__ 
+        
+        attrs_to_skip = ["component_attrs", "components", "color", "model",  "input", "output",
```
jetuk: What attributes are we trying to get here? This seems a bit hackish on a second look. I guess it is related to the schema stuff? Perhaps if that is merged the schema should be used here to dump nodes to JSON data.

jetuk:
I will merge this, but open an issue for the `attrs_to_skip` related to the schemas in #611.

Issue #613, opened 2018-04-12
pypi distribution


snorfalorpagus:
We should do a source release on pypi.

jetuk:
Initial attempt here. 

https://pypi.org/project/pywr/0.5/

It uses the current README for the long description (I need to do the PR for this). It is also sdist only. Apparently binary wheels are broken on Linux.

jetuk:
OK, so this will need some work. `setup.py` is broken with the git repo. I'll try to do a PR so we can do a v0.5.1 release quickly in pypi that actually works.

snorfalorpagus:
Specifically, setup.py doesn't work when not run from a git repository as it tries to access the version. Traceback here:

```
$ pip install pywr-0.6.dev0.tar.gz 
Processing ./pywr-0.6.dev0.tar.gz
    Complete output from command python setup.py egg_info:
    fatal: not a git repository (or any of the parent directories): .git
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-33kiv9d2-build/setup.py", line 143, in <module>
        git_hash = subprocess.check_output(["git", "rev-parse", "HEAD"]).rstrip().decode("utf-8")
      File "/home/snorf/miniconda3/envs/python36/lib/python3.6/subprocess.py", line 336, in check_output
        **kwargs).stdout
      File "/home/snorf/miniconda3/envs/python36/lib/python3.6/subprocess.py", line 418, in run
        output=stdout, stderr=stderr)
    subprocess.CalledProcessError: Command '['git', 'rev-parse', 'HEAD']' returned non-zero exit status 128.
    Using setuptools for setup!
```

This shouldn't be difficult to fix.

snorfalorpagus:
Also need to configure the source distribution to include pyx files, as well as .c files. See how Fiona does this.

jetuk:
So this was a mess. #614 was merged and also attempt to do pypi (and conda) deployment in this repo. Turns out building wheels for Linux is quite difficult and needs its own process. This process is semi-manually and is being added to `pywr-deploy`. The core problems here have been fixed in terms of generating a sdist and being able to generate wheels for pypi.

snorfalorpagus:
We have a source distribution on pypi now:

https://pypi.org/project/pywr/#files

I've just successfully built and run this without doing anything special, so making this as done.

Binary wheels for the various combinations of OS / Python version can be handled in separate issues as they are likely to have their own specific needs.

PR #614, opened 2018-05-02
Fix setup.py and add MANIFEST.in

jetuk:
This starts to addresses #613 by fixing `setup.py` and adding a `MANIFEST.in`. I think it works locally for me. If this is OK we can merge and release 0.5.1 to pypi. 

Do you think it possible to do the deploy to pypi with  the `pywr-deploy`? Specifically it would be good to build wheels for all the platforms and upload those.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: generic
+
 sudo: false
+
 cache:
     directories:
         - ${HOME}/miniconda3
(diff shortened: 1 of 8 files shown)


jetuk:
Oh bugger I've branched from the wrong branch here.

jetuk:
OK I fixed this so it doesn't include a random commit from the `schema` branch. Should be good to review.

PR #619, opened 2018-05-17
Allow specifying solver name via PYWR_SOLVER environment variable.

### LINKED ISSUE #618: Allow specification of solver via environment variable.
@Batch21 highlighted that there's possible an issue in the tests that allows a new test writer to simple initalise a `pywr.model.Model` instance without using the provided fixtures. If this is the case it will always use the default solver (GLPK) despite what is given to `--solver`. 

One way to avoid this could be to allow specification of the solver via an environment variable (e.g. `PYWR_SOLVER`). The fixture could be removed and `Model` look for this variable if no solver is given by default. If no environment and no solver given directly it would then default to GLPK.

What do you think?

jetuk:
- Env variable is used if solver is not given directly.
- If neither given it retains the current default behaviour.
- Removed the "--solver" argument from pytest and the solver fixture.
- Modified all the tests that depended on this fixture.
- Modified the conda recipe to specify solver by environment variable.

Fixes #618.

### DIFF
--- a/conda-recipe/run_test.bat
+++ b/conda-recipe/run_test.bat
@@ -1,5 +1,7 @@
-py.test -v %SRC_DIR%\tests --solver=glpk
-py.test -v %SRC_DIR%\tests --solver=lpsolve
+set PYWR_SOLVER=glpk
+py.test -v %SRC_DIR%\tests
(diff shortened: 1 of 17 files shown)


snorfalorpagus:
LGTM 👍

PR #622, opened 2018-05-22
Fix a bug in the AnnualHarmonicSeriesParameter 

jetuk:
amplitude and phase were not updated correctly. This is pretty subtle, but I think to do with when numpy/cython does or does not make a copy of the underlying memoryview. Anyway I'm pretty sure it wasn't work right, but this makes it much more explicit in any case.


snorfalorpagus:
Can we get a test demonstrating that this fixes the issue?

jetuk:
Added the test. It would be interesting to know whether this is a change due to a change in cython or not.

jetuk:
So I did some testing of this. It's the difference between `np.array` and `np.asarray` in the properties. The former (as it was) returns a copy of the data, the later does not.

Issue #594, opened 2018-02-11
CachedParameter can be removed.


jetuk:
Since the component dependency tree was implemented `CachedParameter` is redundant. It should be removed.

jetuk:
Seems like it's only the definition in `_parameters.pxd` that remains.

PR #626, opened 2018-06-12
removed CachedParamter defition

laurenpetch:
I did a search and couldn't see it anywhere else so just deleted the definition (#594)


jetuk:
LGTM

PR #631, opened 2018-06-13
Add new HydroPowerTargetParameter

jetuk:
This parameter can be used to calculate flow given an energy target. See docstring.

Test included and modifications to existing hydropower recorders docstring to reference this new one. Added a new section to parameters API docs for hydropower.

### DIFF
--- a/docs/source/api/pywr.parameters.rst
+++ b/docs/source/api/pywr.parameters.rst
@@ -90,7 +90,7 @@ Control curve parameters
     control_curves.ControlCurveInterpolatedParameter
     control_curves.ControlCurveIndexParameter
 
-Hydro-power parameters
+Hydropower parameters
 ----------------------
 
 .. autosummary::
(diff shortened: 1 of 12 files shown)


docs/source/api/pywr.parameters.rst
```diff
@@ -90,6 +90,14 @@ Control curve parameters
     control_curves.ControlCurveInterpolatedParameter
     control_curves.ControlCurveIndexParameter
 
+Hydro-power parameters
```
snorfalorpagus: Usually without the hyphen.

pywr/parameters/_hydropower.pxd
```diff
@@ -0,0 +1,16 @@
+from .._core cimport Timestep, Scenario, ScenarioIndex, AbstractNode, Storage, AbstractStorage
+from ._parameters cimport Parameter, IndexParameter
+
+
+
+cdef class HydroPowerTargetParameter(Parameter):
```
snorfalorpagus: HydroPower vs Hydropower?

jetuk: Agreed should be the later. For consistency we will need to change the parameters that are merged already too.

PR #635, opened 2018-06-22
WIP: Refactor of several recorders

jetuk:
This started out as adding two new recorders: `TotalParameterRecorder` and `MeanParameterRecorder` to be equivalent of the total and mean flow recorders for a node. 

However, it made me realise that the standard `NumpArrayXXX` parameters should support arbitrary temporal aggregation and support the `.values()` and `.aggregated_value()` methods. Following that I refactored the existing recorders that have some sort of temporal aggregation (e.g. `FlowDurationCurveRecorder`) to use a consistent naming `temporal_agg_func` as opposed to recorder specific keywords. This is all then achieved through a common `Aggregator` function that abstracts the setting and application of "agg_func" properties to all the recorder subclasses.

Along the way I tried to improve some docstrings too.

Todo:
- [ ] `Aggregator` needs some tests of its own.
- [ ] More docstrings need completing. 
- [ ] Checking I've not missed anything obvious here.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -1,6 +1,7 @@
 import numpy as np
 cimport numpy as np
 import pandas as pd
+import warnings
 from past.builtins import basestring
 
 recorder_registry = {}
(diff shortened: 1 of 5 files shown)


jetuk:
I'm unlikely to spend any more time on the docstrings now. I've fixed a couple of things, but would be grateful for a review of this.

jetuk:
@Batch21 @snorfalorpagus either of you have time to review this? Once this is merged we can probably do a v0.6 release.

pywr/recorders/_recorders.pyx
```diff
@@ -40,29 +37,113 @@ _obj_direction_lookup = {
...
+                func = _agg_func_lookup[func.lower()]
+            elif callable(func):
+                self._user_func = func
+                func = AggFuncs.CUSTOM
+            else:
+                raise ValueError("Unrecognised aggregation function: \"{}\".".format(func))
+            self._func = func
+
+    cpdef double aggregate_1d(self, double[:] data, ignore_nan=False) except *:
+        """Compuate an aggreagted value across 1D array.
```
Batch21: typos

Batch21:
Looks good. One thing I not sure about is the naming of `temporal_agg_func`.  When it is applied in the flow duration recorders it is aggregating across percentiles not time, which is a little confusing. Could the name be more generic? It might be clearer to call it `agg_func` and rename the original `agg_func` to `scenario_agg_func`, though this would probably cause greater backward compatibility issues.  

Another question I have after looking at this is why `AggFuncs` and `_agg_func_lookup` need to be defined as global variables rather that within `Aggregator` I sure there is a good reason, I just can't work it out.

jetuk:
Ultimately this is end user territory, so if "temporal" is confusing for a percentile based output then we should be clearer. 

re `AggregatedRecorder`. That docstring is a mess. The `__init__` docstring should be removed (it's the wrong format anyway). I think the description at class level is OK, but it needs a the parameters section writing in the correct format.

Batch21:
temporal is less confusing if there is a `scenario_agg_func' arg. I think it just needs a line in the docstring explaining that the input into the temporal agg function may have already been modified by the recorder.

jetuk:
I've clarified the `AggregatedRecorder` docstring. 

I've also had second thoughts on `scenario_agg_func`. This weekend I've written a recorder that returns a value per year in `.values()` rather than per scenario. `agg_func` is still used in `aggregated_value()` as is, but in this case `scenario_agg_func` would make no sense.

jetuk:
I've tacked one more commit on to here. @Batch21 could you check it quickly and then I'm going to merge this if OK.

jetuk:
I've rebased this ready for merging.

Batch21:
Had a quick look through the additional commits and they look good

PR #641, opened 2018-06-30
WIP: Schemas the right way!

jetuk:
After the discussion in #611 this new PR branches from the approach using a `Schema` class on every node. However, it would now utilise custom fields to directly load nodes and components. This is opposed to the approach that simply used the schemas for validation.

This PR currently shows a proof of concept in the `tests/test_schema.py`. Here a Schema is created for a particular class and valid data is loaded in to an instance of that class. Note that crucially this does *not* use the `.load` classmethod to create the instance. 

@snorfalorpagus @Batch21 what do you think?

NB I've disabled using the schemas during `Model.load` for the time being.


jetuk:
So, after a little more work we've realised this is going to bump in to #380 (which #586 was also trying to fix). Circular loading will become a difficulty to do this "right". While I'm sure we can implement a marshmallow based loading scheme with the same limitations as we have right now it seems like we should solve this problem at the same time.

Some thoughts from @snorfalorpagus quoted below (formatting adjusted).

> - Circular dependency of Parameters is invalid - even if it was possible to create the objects, it wouldn't be possible to evaluate their value at run time.
>  - Therefore circular loading of parameters shouldn't be a problem for us, although we should test a useful error is raised
>  - Circular loading of parameters in combination with nodes is possible, and is valid because the state > of nodes is not dynamic (i.e. values like current storage never have any dependencies
>  - A silly example, two reservoirs, each of who's cost depends on the current storage of the other reservoir.
>  - This creates a circular load error in the current model.
> - Two solutions to this problem:
>    1. load all nodes before parameters, or 
>     2. use some kind of node reference/placeholder.
> - Placeholder is nice, but probably awkward - needs to have the correct type in Cython to be stored. Then need some way to resolve the reference at end of load.
> - Loading node first is OK, but means many parameters aren't set using the init method. Also, some non-parameter values must be set at init, e.g. the num-inputs value for compound nodes, as this can't be changed afterwards. So only do this for properties that accept a Parameter. As a general rule, anything that can be set to a parameter should be able to be changed after init, as Parameters themselves are a kind of deferred value.

I agree with this. It boils down to instancing nodes with all arguments besides parameter instances. Delaying the parameter creation until after nodes are loaded. Then assigning / updating the parameters to the correct attributes. This is the approach used currently in several `.load` methods to avoid common circular problems. 

I don't know how the above would fit it on the marshmallow framework. The `ParameterField` would need changing to not load (only validate?). The `@post_load` method would have to pass all data except parameters to `cls.__init__` before assigned the parameters by attribute to the new Node instance.

jetuk:
Quick update on this.

Majority of the implementation is complete for Nodes. However, there needs to be more comprehensive tests for each node to ensure the attributes between the Node instance and schema are actually setting the correct things.

There are some outstanding issues such as the concept of `max_flow` for piecewise links. The attribute is settable but does nothing because the node you instantiate is never directly part of a route. This is possibly a side issue.

Regarding implementing schema for components (parameters and recorders) there is an issue as follows:

```
Error compiling Cython file:
------------------------------------------------------------
...
    pass


cdef class Parameter(Component):

    class Schema(ParameterSchema):
   ^
------------------------------------------------------------

pywr/parameters/_parameters.pyx:24:4: class definition not allowed here

```

It is not possible for a Cython extension type to contain a nested class reference. Therefore we can't store the component schema's on the extension type in the same way. I propose the following approach. The parent `Component.__init__` uses a private method to assign a class object to `.Schema`. This can then be used in the same way as the `Node.Schema`.

Component authors must then implement this new method to return the appropriate class. 

```
cdef class Component:
    cdef public object Schema   # this is new 
    cdef object _name
    ...
```


```
cdef class Component:
    def __init__(self, model, name=None, comment=None):
        ...
        self.Schema = self._get_schema_klass()

    def _get_schema_klass(self):
        raise NotImplementedError()
````

Any thoughts @pywr/pushers

**Comment by Batch21** [comment]
this all looks good. The fact that marshmallow has removed the need for node load methods makes it use seem a lot cleaner than it previously did for me. 

Regarding the component schema issue, would the schema be define as a separate class in pyx or are they going to be put in .schema

**Comment by jetuk** [comment]
That's a good question. The choice is one of these two I think. It's mostly a namespace/style issue. Either way having to implement a private method is not great for component subclass authors. I can't see another way of doing this though.


```cython
cdef class MyParameter(Parameter):
    def __init__(self, model, name=None, comment=None):
        ...
        super().__init__(...)

    def _get_schema_klass(self):
        class Schema(ParameterSchema):
            value = fields.Number()
        return Schema
```

vs

```cython
class MyParameterSchema(ParameterSchema):
    value = fields.Number()

cdef class MyParameter(Parameter):
    def __init__(self, model, name=None, comment=None):
        ...
        super().__init__(...)

    def _get_schema_klass(self):
        return MyParameterSchema
```

Batch21:
I didn't realise the first method was possible. I think that method maybe better than the second as its slightly closer in structure to the way it done for the nodes - in that it keeps the schema class internal to the component class.

Would be nice if there was a cleaner way but this doesn't seem like too much extra overhead for people creating component subclasses

jetuk:
I'm assuming it is possible. Classes are just objects in Python. Though I haven't tested it.

jetuk:
Looking at this again. I think I found a prettier solution. Still requires you to define an external schema, but it uses the existing register classmethod to associate a schema with the parameter class. Then you can retrieve the schema through a `.Schema` property.

```python
class MyParameterSchema(ParameterSchema):
    value = fields.Number()

cdef class MyParameter(Parameter):
    def __init__(self, model, name=None, comment=None):
        ...
        super().__init__(...)

    @classmethod
    def register(cls, schema):
        parameter_registry[cls.__name__.lower()] = cls
        # add schema to registry
        parameter_schema_registry[cls.__name__.lower()] = schema

    @property
    def Scehma(self):
         return parameter_schema_registry[cls.__name__.lower()]

MyParameter.register(MyParameterSchema)
```

PS we should probably do some name collision checking here.

jetuk:
After discussion with @Batch21 yesterday we fleshed out the details of the final paragraph of the previous comment. Currently the JSON formats something like the following for any parameters that take a value or values input (constant, profile, etc.)

```json
"myparam": {
  "type": "constant",
  "url": "myfile.csv",
  "column": "A",
  "index": "thisone",
  "dayfirst": true,
  "random_pandas_kwarg": "value"
}
```
This has two issues,
  1. The schema is open ended with any non-specified keys being passed directly to the appropriate `pandas.read_xxx` function. This means validation of the schema is difficult.
  2. This pattern only allows one url for each parameter. While this is the case for many it's not that flexible. Think of an interpolation parameter that might need x and y arrays.  Notice also that the `value` key is replace and not required here. 

The proposed schema would be:

```json
"myparam": {
  "type": "constant",
   "value": {
    "url": "myfile.csv",
    "column": "A",
    "index": "thisone",
    "pandas_kwargs": {  "dayfirst": true,  "random_pandas_kwarg": "value" }
  }
}
```
This has the following advantages,
1. pandas keyword arguments are explicit.
2. The attribute (`value`) remains so it is explicit which attribute is being loaded.
3. The schema is mostly explicitly (only the  "pandas_kwargs" remains as an open ended dict).
4. It is more flexible. Something like this would be possible:

```json
"myparam": {
  "type": "interpolated",
   "x": {
    "url": "myfile.csv",
    "column": "A",
    "index": "thisone",
    "pandas_kwargs": {  "dayfirst": true,  "random_pandas_kwarg": "value" }
  },
   "y": {
    "url": "myfile.csv",
    "column": "B",
    "index": "thisone",
    "pandas_kwargs": {  "dayfirst": true,  "random_pandas_kwarg": "value" }
  }
}
```
This is obviously backward incompatible, but I think an improved approach. Any thoughts?

snorfalorpagus:
The current test failure is due to the "two_reservoirs.json" model setting `max_flow` on a Catchment instead of `flow`.

```
        {
            "name": "catchment1",
            "type": "catchment",
            "max_flow": "inflow"
},
```

This should be updated in the model.

jetuk:
I've added a custom error handler to the schemas. In the above example it would produce the following output. 

```
>       raise marshmallow.exceptions.ValidationError(msg)
E       marshmallow.exceptions.ValidationError: Error(s) occurred with the data for class <class 'pywr.domains.river.Catchment'>: {'name': 'catchment1', 'max_flow': 'inflow'}
E         Validation error messages: 
E           max_flow: ['Unknown field.']
```

I also fixed the bug in the JSON file.

jetuk:
@snorfalorpagus @Batch21 https://github.com/pywr/pywr/pull/641/commits/a0d9fb772a9b64849c2106747e03c25b81fe98e5 adds initial support for dataframe fields. What do you think?

A couple of the tests haven't been moved over to the new JSON schema that this requires.

Batch21:
I had a look through this again after your previous comment but I'd lost track a bit of the different approaches. I'll have another look but I think the current approach probably has the lowest overhead for new users. It just involves adding a nested schema class to each node/component and specifying the schema field for each of the attributes of the outer class?

jetuk:
I think so. I guess it's a bit more code than previously. 

Did you have any comment specifically on the change to the dataframe schema?

jetuk:
I've just merged `v1.6.0` (master) back in to here to keep this tracking.

jetuk:
While this was a useful exercise I don't think it's going to get merged. The thinking has gone in to `pywr-next`.

PR #649, opened 2018-08-25
Removed finally caluse in _cached property of BaseOptimisationWrapper

jetuk:
There are instances where the `cache` variable might not be assigned. Best to remove the finally clause entirely.


jetuk:
Thanks @Batch21

PR #657, opened 2018-09-08
Adding area attribute

avnish98:
I have added area attribute to Storage() class and added test in test_reservoir_surface_area() using hasattr()


jetuk:
Thanks for this.

PR #668, opened 2018-10-02
WIP: Node documentation

Batch21:
I started this back in January but forgot about it. I've just added in the river domain nodes. Should I also add the nodes from _core.pyx ?


Batch21:
Just noticed that the river domain nodes have already been added in a different location. I'll remove them from this pr

PR #677, opened 2018-11-01
Fixed the fdc deviation recorder to work with no scenario

igricalde:


jetuk:
We ought to add a regression test for this.

Issue #669, opened 2018-10-10
NumpyArrayAreaRecorder is missing


jetuk:
There are recorders for flow, volume and level but not one for area of `Storage` nodes.

jetuk:
Fixed in #684

PR #684, opened 2019-01-05
Add area recorder

Batch21:
some work on #669 

Currently the area stored by the recorder is calculated using reservoir volume from the previous timestep. I think this is because the `get_area` method of the storage node calls `get_value` on the area parameter which returns the area value calculated in `calc_values`, which is calculated before the reservoir volume is updated for the current timestep.

I’m not sure if this is desirable behavior or not. Changing the `get_area` method so that it calls `value` rather than `get_value` seems to work in terms of recording the area of the current timestep. Should I make this change?

### DIFF
--- a/docs/source/api/pywr.recorders.rst
+++ b/docs/source/api/pywr.recorders.rst
@@ -30,6 +30,7 @@ accessed following a model run before the model instances is destroyed.
    NumpyArrayNodeRecorder
    NumpyArrayStorageRecorder
    NumpyArrayLevelRecorder
+   NumpyArrayAreaRecorder
    NumpyArrayParameterRecorder
    NumpyArrayIndexParameterRecorder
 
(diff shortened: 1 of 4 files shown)


jetuk:
I guess it should be the same as the level recorder?

Batch21:
The level recorder, which I don't think has any tests, has the same behavior, which is different from the `NumpyArrayStorageRecorder` which record values for the current timestep.

jetuk:
I see what you mean. I can see how this could be inconsistent, but also that it's consistent with the values that were used from the parameters in this time-step. Now if you reconcile that with the volume saved at the same timestep it won't be consistent. 

For reference #128 added the level recorder.

jetuk:
I've done some work to this to tidy it up. There was a small refactor to make the area, level and volume recorders all share the same baseclass.

Issue #470, opened 2017-04-29
Storage max volume parameter is accessed before initialisation


snorfalorpagus:
The `setup`/`reset` method on nodes is called before the component tree is evaluated. See:

https://github.com/pywr/pywr/blob/0b12cadb34b7207737ba53b258168fb67cceb353/pywr/model.py#L607-L621

If the maximum volume of a `Storage` node is defined by a parameter, the current percent full is mis-calculated in the initial timestep, as the parameter is accessed during the `Storage.reset` method *before* the parameter has been evaluated.

I came across this while trying to implement #453, which also requires the maximum volume.

The solution to this seems to be to include the nodes in the component tree, which I think we have discussed as a possibility already.

snorfalorpagus:
I started work on trying to make `Node` into a subclass of `Component`. Most of the tests pass with only a few changes. However...

The difficult part will be maintaining the parameters relationship with the nodes. When a node property is set to a parameter it and edge needs to be added to the tree. When another parameter is set to the same property the edge must be removed, *but only if it's not also used by another property of the same node*.

A possible solution would be to add a tuple to the tree in the form: `G.add_edge((node, "max_flow"), parameter)`.

I wonder if the whole thing is complicated by some nodes needing to call setup before the parameters, and some after? Not sure if this is the case - just thinking aloud.

snorfalorpagus:
The cost of a reservoir may depend on it's volume (storage before parameter), but the volume is defined by the max volume (parameter before storage). This suggests it *does* need to go into the tree. I still get the feeling it's more complicated than that though.

jetuk:
> The cost of a reservoir may depend on it's volume (storage before parameter), but the volume is defined by the max volume (parameter before storage). 

I don't think this is a problem. The cost uses previous time-steps volume (and max volume) for its calculations. It's not until `current_pc` is updated at the end of this time-step that the correct max volume is needed. Which works as advertised.

snorfalorpagus:
We've agreed a work-around: in the reset method of Storage, check if the max_volume parameter is defined and if it has any children. If it does, raise an exception. If it doesn't, call calc_values immediately.

We should also add a test for this. Create a ConstantParameter = 50 and set it to the max_volume of a Storage. Set the initial volume to 25. Call setup/reset on the Storage, then check the _current_pc. It should be 0.5, but will often be 0.0 or NaN if the constant parameter was uninitialised when it was accessed.

snorfalorpagus:
Tag #488.

[...]

PR #690, opened 2019-02-02
A better workaround for #470

jetuk:
This change the reservoir volume initialisation in the case where
max_volume is parameter. Now both `initial_volume` and
`initial_volume_pc` must be specified as floats, and are used as
is. This removes the current whitelist of acceptable parameters.

It does however mean that if the user specifies initial conditions
that are not compatible with the actual max_volume (i.e.
initial_volume != initial_volume_pc/max_volume) then there is the
possibility that parameter values will be incorrect on the first
time-step. This changes effectively makes that the users
responsibility.


KevisPachos:
Hi @jetuk,
I would like to apply the new discounting functionality in a scheduling formulation. That would better show the value in delaying implementation of schemes compared to a static formulation. 
Would that require that this branch is reviewed and merged into master? If so, can I somehow help you on making this functionality available?
Thanks

jetuk:
Yes, probably. I haven't looked at this for a while. I'll merge master here to get it up to date for you.

KevisPachos:
Thank you @jetuk

jetuk:
> I'll merge master here to get it up to date for you.

Done.

Batch21:
Implementation looks good. Only comment is that the requirement to set both 'initial_volume` and `initial_volume_pc`  when `max_volume` is a parameter should probably be specified in the `Storage` docstring

jetuk:
Thanks @Batch21. I've updated the `Storage` class docstring to include a paramters section.

PR #692, opened 2019-02-20
Remove conda from CI

jetuk:
This PR is an attempt to remove the use of conda and its recipe from Pywr's CI config. The ultimate objective here is to get Pywr in to conda-forge with the recipe outside of this repository. The other motivation is to align the CI builds & testing with the builds (wheels) that ultimately get pushed to pypi. These wheels should also include the external libraries (GLPK) making installation using pip much easier in future. Currently releasing to pypi is a bit of a manual process (see pywr-deploy) at the moment, and doesn't embed the libraries. It also problematic that those builds are not tested by the current CI process. 

This current set of changes builds wheels for Windows and Linux. Linux is problematic because the [PEP513](https://www.python.org/dev/peps/pep-0513/) requires conforming to the `manylinux1` specification. This is based on CentOS 5 (old) and requires the wheels to be built in that environment - hence the use of docker in the travis config here. I have made a derivative of the `manylinux1` docker image that includes a newer version of GLPK. We will have to support and update that image - and also rebase it once `manylinux2010` is ready.

~~The ancient version environment for `manylinux1` means that there is no shipped version of `pytables` for Python 3.7 on Linux~~. Therefore this CI config us currently failing. I suggest we disable until it is fixed. There's a similar issue for `pygmo` on Windows with 3.7, but that's really an optional dependency so those tests are currently skipped.

- [ ] OSX?
- [x] Building documentation
- [x] Add lpsove.
- [x] Remove the conda recipe.
- [x] ~~Disable Py3.7 on Linux.~~ 3.7 now working on Linux
- [ ] (optional) add automatic deployment (to github and/or pypi) on tags.


jetuk:
We should continue with this now there is a pywr-feedstock for conda forge.

https://github.com/conda-forge/pywr-feedstock

jetuk:
I've added the documentation building back in here and also removed the recipe.

For the docs I've changed it to push to a sub-folder (either `master` or `v-$TRAVIS_TAG`) so we can have different versions up on github pages. Obviously this won't be properly tested until this PR is merged to master.

@snorfalorpagus would you mind reviewing this?

travis/build-wheels.sh
```diff
@@ -0,0 +1,46 @@
...
+# Note this script does not build the wheels with lpsolve support.
+set -e -x
+
+cd /io
+# Setup path to use PYBIN's binary folder
+export PATH=${PYBIN}:${PATH}
+
+# Install the build dependencies of the project. If some dependencies contain
+# compiled extensions and are not provided as pre-built wheel packages,
+# pip will build them from source using the MSVC compiler matching the
```
snorfalorpagus: Reference to MSVC in a bash script?

setup.py
```diff
@@ -171,4 +181,10 @@
 setup_kwargs['ext_modules'] = cythonize(extensions + extensions_optional,
                                         compiler_directives=compiler_directives, annotate=annotate,
                                         compile_time_env=compile_time_env)
+
+if os.environ.get('PACKAGE_DATA'):
```
snorfalorpagus: This would return `True` for something like `PACKAGE_DATA=0`.

.travis.yml
```diff
@@ -1,64 +1,45 @@
...
-      - os: linux
-        env: CONDA_PY=36 BUILD_DOC=1
-      - os: linux
-        env: CONDA_PY=37
-      - os: osx
-        env: CONDA_PY=36
-             MACOSX_DEPLOYMENT_TARGET=10.9
+env:
+  global:
+    - DOCKER_IMAGE=pywr/manylinux1_x86_64-glpk
```
snorfalorpagus: Where does the Dockerfile for this live?

jetuk: It is on my machine atm. Where do you think it should live? We have a `pywr-docker` repo. I guess it should go there.

snorfalorpagus: It could live in the `travis/` directory in this repo.

jetuk:
I've address the comments and added the docker files.

travis/manylinux1_x86_64-glpk/Dockerfile
```diff
@@ -0,0 +1,10 @@
+from quay.io/pypa/manylinux1_x86_64
+
+MAINTAINER tomo.bbe@gmail.com
```
snorfalorpagus: The MAINTAINER statement is deprecated. Instead you should use:

```
LABEL maintainer="tomo.bbe@gmail.com"
```

travis/manylinux1_x86_64-glpk/Dockerfile
```diff
@@ -0,0 +1,10 @@
+from quay.io/pypa/manylinux1_x86_64
```
snorfalorpagus: `FROM` should be upper case

travis/manylinux1_x86_64-glpk/build_lpsolve.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+LPS_VER="5.5.2.5"
+
+# Must download this manually due to old OpenSSL
+# wget https://sourceforge.net/projects/lpsolve/files/lpsolve/${LPS_VER}/lp_solve_${LPS_VER}_source.tar.gz
```
snorfalorpagus: I assume this is no longer true?

travis/manylinux1_x86_64-glpk/build_lpsolve.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# TODO add checksum
```
snorfalorpagus: TODO

travis/manylinux1_x86_64-glpk/build_lpsolve.sh
jetuk: Yes, you have to do this locally. `wget` is broken on this ancient image. I think it's due to sourceforge using a redirect or something. I left the command for reference purposes.

travis/manylinux1_x86_64-glpk/build_lpsolve.sh
jetuk: This will get complete once manylinux2010 is relased and wget is updated.

PR #695, opened 2019-02-22
fixed test warnings

matrosoe:
Fixed test warnings in aggregate node tests
Will work on other warnings

### DIFF
--- a/tests/test_parameters.py
+++ b/tests/test_parameters.py
@@ -67,9 +67,6 @@ class TestConstantParameter:
 
         np.testing.assert_allclose(p.get_double_lower_bounds(), np.array([np.pi/2]))
         np.testing.assert_allclose(p.get_double_upper_bounds(), np.array([2*np.pi]))
-        # Test deprecated API too.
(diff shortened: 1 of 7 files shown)


jetuk:
@matrosoe is this ready now?

matrosoe:
Think so!

jetuk:
I can't merge it until you remove the draft tag.

tests/test_parameters.py
```diff
@@ -67,9 +67,6 @@ def test_being_a_variable(self, simple_linear_model):
 
         np.testing.assert_allclose(p.get_double_lower_bounds(), np.array([np.pi/2]))
         np.testing.assert_allclose(p.get_double_upper_bounds(), np.array([2*np.pi]))
-        # Test deprecated API too.
-        np.testing.assert_allclose(p.lower_bounds(), np.array([np.pi/2]))
```
jetuk: I think we should probably remove the actual API functions on the `Parameter` object too.

PR #700, opened 2019-03-05
Add support for dataframe data embedded directly in to JSON.

jetuk:
A relatively small change that allows dataframe data to be directly encoded in the JSON document via a "data" key.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -1503,7 +1503,7 @@ def load_dataframe(model, data):
         name = table_ref
         df = model.tables[table_ref]
     else:
-        name = data.get('url')
+        name = data.get('url', None)
         df = read_dataframe(model, data)
 
     # if column is not specified, use the whole dataframe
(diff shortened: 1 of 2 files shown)


matrosoe:
Perhaps some tests to check non-time series dataframes should be added?

pywr/parameters/_parameters.pyx
```diff
@@ -1538,27 +1538,37 @@ def load_dataframe(model, data):
...
+        # Must be an embedded dataframe
+        df_data = data.pop('data', None)
+
+    if url is None and df_data is None:
+        raise ValueError('No data specified. Provide a "url" or "data" key.')
 
     # Check hashes if given before reading the data
     checksums = data.pop('checksum', {})
     for algo, hash in checksums.items():
         check_hash(url, hash, algorithm=algo)
```
jetuk: `url` could be `None` here. Move inside if block bellow.

jetuk: Fixed.

PR #705, opened 2019-03-28
Enable codecov based coverage reporting.

jetuk:


jetuk:
The error above is because we don't have a coverage report for the master commit this was based off. It should go away in future PR once this is merged.

PR #714, opened 2019-04-23
Add InterpolatedQuadratureParameter

jetuk:
A parameter that returns the interpolated quadrature of another parameter.


tests/test_parameters.py
```diff
@@ -951,6 +951,31 @@ def expected_func(timestep, scenario_index):
...
+        elif i < 11:
+            value = 25 + 4*(i - 5)**2 / 2 + (i - 5) * 10
+        else:
+            value = 25 + 50 + 50 + 28 / 2 + 2
+        return value
+
+    model.run()
+
+
+
```
snorfalorpagus: There should only be 2 blank lines between classes/functions.

jetuk:
I merge master in to here so it does the new CI correctly.

Batch21:
Gone through this and it looks good. Can't think of a use case, but is it worth making it possible to set the `a` arg of the quad function to a parameter?

jetuk:
> ... is it worth making it possible to set the a arg of the quad function to a parameter

Probably, for completeness. I guess this should also be interpolated.

jetuk:
@Batch21 I've addressed your comment. If you are happy could submit a review and approve?

PR #719, opened 2019-04-28
Added getter for Recorder.agg_func

### LINKED ISSUE #681: Recorder.agg_func is not readable
It should be possible to do something like this without the error.

```
    metrics_aggregated[rec.name] = {'agg_func': rec.agg_func, 'value': np.array(rec.aggregated_value())}
AttributeError: attribute 'agg_func' of 'pywr.recorders._recorders.Recorder' objects is not readable
```

snorfalorpagus:
Added a `__get__` method for `Recorder.agg_func` and the underlying `Aggregator`.

Closes #681.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -955,6 +955,7 @@ _agg_func_lookup = {
     "all": AggFuncs.ALL,
     "median": AggFuncs.MEDIAN,
 }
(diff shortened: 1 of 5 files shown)


jetuk:
LGTM.

We should add the same to `AggregatedParameter` and `AggregatedIndexParameter`.

snorfalorpagus:
> We should add the same to AggregatedParameter and AggregatedIndexParameter.

I'll make those changes in this branch.

PR #720, opened 2019-04-28
Refactor of named parameter/recorder loading

### LINKED ISSUE #674: Recorders loading recorders out of order is broken.
I think this has not been done before, but should be supported.

Using the `load_recorder` function from the `.load` method of another recorder does not correctly load the recorder if it is not already loaded. 

First this line should pop the data from the `._recorders_to_load` dictionary.
https://github.com/pywr/pywr/blob/60cf2b9ae4c9507355c6491ae7ddf937eb71410e/pywr/recorders/_recorders.pyx#L1364

However this does not working because that dictionary is currently being iterated. See here.

https://github.com/pywr/pywr/blob/60cf2b9ae4c9507355c6491ae7ddf937eb71410e/pywr/_model.pyx#L375

The solution is to implement the same approach as is used for loading parameters (a few lines below).

snorfalorpagus:
Refactor of the code used to load named parameters and recorders to use shared functions (as they are both components).

Closes #674.

### DIFF
--- a/pywr/_model.pyx
+++ b/pywr/_model.pyx
@@ -8,6 +8,7 @@ from packaging.version import parse as parse_version
 import warnings
 import inspect
 import time
+from functools import wraps
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


jetuk:
```
    @pytest.fixture()
    def two_reservoir_problem():
        filename = os.path.join(TEST_FOLDER, 'models', 'two_reservoir.json')
>       yield TwoReservoirWrapper(filename)
tests/test_platypus.py:21: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pywr/optimisation/platypus.py:16: in __init__
    m = self.model
pywr/optimisation/__init__.py:83: in model
    return self._cached.model
pywr/optimisation/__init__.py:70: in _cached
    model = self.make_model()
pywr/optimisation/__init__.py:102: in make_model
    m = Model.load(self.pywr_model_json)
pywr/_model.pyx:258: in pywr._model.Model.load
    return cls.loads(data, model, path, solver)
pywr/_model.pyx:192: in pywr._model.Model.loads
    return cls.load(data, model, path, solver, **kwargs)
pywr/_model.pyx:372: in pywr._model.Model.load
    load_components(model._recorders_to_load, load_recorder)
pywr/_model.pyx:893: in pywr._model.listify.wrapper
    return list(f(*args, **kwargs))
pywr/_model.pyx:369: in load_components
    component = load_component(model, component_data, name)
pywr/recorders/_recorders.pyx:1451: in pywr.recorders._recorders.load_recorder
    recorder = cls.load(model, data)
pywr/recorders/_recorders.pyx:286: in pywr.recorders._recorders.AggregatedRecorder.load
    recorders = [model.recorders[name] for name in recorder_names]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
>   raise KeyError("'{}'".format(key))
E   KeyError: "'deficit1'"
```

Test is failing to load the `two_reservoir.json` model in the platypus tests.

snorfalorpagus:
I didn't have platypus installed so was missing the failure that travis found. It looks like `two_reservoirs.json` is the only example of recursive recorder loading?

jetuk:
It must be based on this failure.

PR #721, opened 2019-04-28
Use flow instead of max_flow in two_reservoirs

### LINKED ISSUE #666: catchment flow in the MOEA example
In the two_reservoir.json model used in the MOEA examples, the inflow dataframe parameter is assigned to the `max_flow` attributes of both catchments, but this doesn't seem to work. I assume that it should be assigned to the `flow` attributes instead?

snorfalorpagus:
The existing code worked because of the costs in the model, but strictly the catchments should use `flow` instead of `max_flow`.

Closes #666.

### DIFF
--- a/examples/two_reservoir.json
+++ b/examples/two_reservoir.json
@@ -13,12 +13,12 @@
         {
             "name": "catchment1",
             "type": "catchment",
-            "max_flow": "inflow"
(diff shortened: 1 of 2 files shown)


jetuk:
We should do this to `two_reservoir.json` in `tests/models` too.

PR #723, opened 2019-04-29
Fix a bug with AggregatedRecorder not returning the instance on load.

jetuk:
- Includes associated test for future regression.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -292,6 +292,7 @@ cdef class AggregatedRecorder(Recorder):
         recorder_names = data.pop("recorders")
         recorders = [load_recorder(model, name) for name in recorder_names]
         rec = cls(model, recorders, **data)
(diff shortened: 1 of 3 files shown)


pywr/recorders/_recorders.pyx
```diff
@@ -292,6 +292,7 @@ cdef class AggregatedRecorder(Recorder):
         recorder_names = data.pop("recorders")
         recorders = [load_recorder(model, name) for name in recorder_names]
         rec = cls(model, recorders, **data)
+        return rec
```
snorfalorpagus: Good spot. Type annotations / linting might have caught this.

We could also (in another PR?) check if the load methods always return a `Component`?

jetuk: I'll move this to a separate issue.

jetuk: See #727

PR #724, opened 2019-04-30
Add flow parameter for tracking yesterday's flow.

jetuk:
- Includes a test and test model.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -1377,6 +1377,56 @@ cdef class DeficitParameter(Parameter):
 DeficitParameter.register()
 
 
+cdef class FlowParameter(Parameter):
+    """Parameter that provides the flow from a node from the previous time-step.
+
+    Parameters
+    ----------
+    model : pywr.model.Model
+    node : Node
+      The node that will have its flow tracked
+
+    Notes
+    -----
+    This parameter keeps track of the previous time step's flow on the given node. These
+    values can be used in calculations for the current timestep as though this was any
+    other parameter.
+    """
+    def __init__(self, model, node, *args, **kwargs):
+        super().__init__(model, *args, **kwargs)
+        self.node = node
+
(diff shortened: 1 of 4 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -1377,6 +1377,56 @@ cdef class DeficitParameter(Parameter):
...
+            num_comb = 1
+        self.__next_values = np.empty([num_comb], np.float64)
+
+    cpdef reset(self):
+        self.__values[...] = 0.0
+
+    cdef calc_values(self, Timestep timestep):
+        cdef int i
+        for i in range(self.__values.shape[0]):
+            print(i, self.__next_values[i], timestep)
```
snorfalorpagus: remove debug

pywr/parameters/_parameters.pyx
```diff
@@ -1377,6 +1377,56 @@ cdef class DeficitParameter(Parameter):
 DeficitParameter.register()
 
 
+cdef class FlowParameter(Parameter):
+    """Parameter that provides the flow from a node from the previous time-step.
```
snorfalorpagus: What is the value of this parameter on the initial timestep? Does it need an initial value?

jetuk: Good point. It is currently zero and not configurable.

pywr/parameters/_parameters.pyx
jetuk: Done

pywr/parameters/_parameters.pyx
```diff
@@ -1377,6 +1377,56 @@ cdef class DeficitParameter(Parameter):
...
+
+    Parameters
+    ----------
+    model : pywr.model.Model
+    node : Node
+      The node that will have its flow tracked
+
```
snorfalorpagus: Need to add `initial_value` to docstring.

PR #725, opened 2019-04-30
WIP: Radial basis function

jetuk:
Here is some initial work on an RBF based parameter. 

There are some things to do -
- [ ] Decide which values should / should not be exposed as "variables" for the optimisation wrappers.
- [ ] Write a better numerical test
- [ ] Decided whether to complete the simpler version that justs uses volume.

Feedback welcome.


jetuk:
Discussed how to specify which coefficients are variables today:

```json
{
        "rbf": {
            "type": "rbf",
            "y": {
                 "coefficients": [1, 2, 3],
                 "is_variable": true, 
                 "upper_bounds": 10.0,
                 "lower_bounds": 0.0
            },
            "nodes": {
                "supply1": [1, 0.5, 0]
            },
            "parameters": {
                "constant": [2, 4, 6]
            },
            "days_of_year": [1, 100, 200]
        }
}
```

pywr/parameters/rbf.py
```diff
@@ -0,0 +1,331 @@
...
+            args.append(x)
+
+        if days_of_year is not None:
+            # Normalise DoY to be between 0 and 1.
+            x = 2 * np.pi * (ts.dayofyear - 1) / 365
+            args.append(np.sin(x))
+            args.append(np.cos(x))
+
+        # Perform interpolation.
+        return self._rbf_func(*args)
```
jetuk: Add a (optional?) clipping to make this return a value >= 0?

```
return max(self._rbf_func(*args), 0)
```

jetuk:
I've merged master in to here to keep this up to date.

@Jmiguel17 if you're testing this let me know if it works well or not. It would be nice to get it finished, tidied up and merged. The docstrings on the parameters needs some work by the looks of it; see also my comments at the top of the discussion.

pywr/parameters/rbf.py
Jmiguel17: @jetuk, Before adding the clipping I tried to run the optimisation but I got this:
```
    m = Model.load(self.pywr_model_json)
  File "pywr\_model.pyx", line 258, in pywr._model.Model.load
  File "pywr\_model.pyx", line 192, in pywr._model.Model.loads
  File "pywr\_model.pyx", line 303, in pywr._model.Model.load
  File "pywr\_model.pyx", line 86, in pywr._model.Model.__init__
  File "c:\users\n65847jg\dropbox (the university of manchester)\doctorado\pywr\pywr\solvers\__init__.py", line 156, in <module>
    from .cython_lpsolve import CythonLPSolveSolver as cy_CythonLPSolveSolver
  File "pywr\recorders\_recorders.pxd", line 6, in init pywr.solvers.cython_lpsolve
ValueError: pywr.recorders._recorders.Aggregator size changed, may indicate binary incompatibility. Expected 56 from C header, got 40 from PyObject
```
Additionally, I changed the `timestep` freq to "D" and the error is:
```
  File "c:\users\n65847jg\dropbox (the university of manchester)\doctorado\pywr\pywr\optimisation\__init__.py", line 102, in make_model
    m = Model.load(self.pywr_model_json)
  File "pywr\_model.pyx", line 258, in pywr._model.Model.load
  File "pywr\_model.pyx", line 192, in pywr._model.Model.loads
  File "pywr\_model.pyx", line 300, in pywr._model.Model.load
ValueError: invalid literal for int() with base 10: 'D'
``` 
it is supposed having merged `master` the pandas' freq may be available. isn't it?

jetuk: Did you recompile?

Jmiguel17: Thanks. I had compiled but I did not `pull` the changes before the recompilation :flushed:

Jmiguel17:
@jetuk I ran the model with the clipping and is working. However, I would like to do some modifications in the way that the variables are stored. maybe this is related to your point of "Decide which values should / should not be exposed as "variables" for the optimisation wrappers". 

For example, the next are the results of one FE  
```
{"variables": [{"name": "rbf[d0]", "value": 0.5277444463941374, "upper_bounds": 1.0, "lower_bounds": 0.0}, 
			   {"name": "rbf[d1]", "value": 0.28299837732765387, "upper_bounds": 1.0, "lower_bounds": 0.0}, 
			   {"name": "rbf[d2]", "value": 0.16449953888733018, "upper_bounds": 1.0, "lower_bounds": 0.0}, 
			   {"name": "rbf[d3]", "value": 0.2752038767721551, "upper_bounds": 1.0, "lower_bounds": 0.0}, 
			   {"name": "rbf[d4]", "value": 173.2333631154337, "upper_bounds": 365.0, "lower_bounds": 1.0}, 
			   {"name": "rbf[d5]", "value": 355.5608323702431, "upper_bounds": 365.0, "lower_bounds": 1.0}], 
```
Here you can see that all variables take the name of the parameter `rbf`. However, I would like to differentiate the names of the `nodes`, `parameters` or `DoY` used in the interpolation when the results are stored. Maybe for the `Parameters` can be defined outside `rbf` itself and then passed to the `rbf` parameter as a list? but I don't know how can be done this for the node or nodes if there are more that one.

jetuk:
> Maybe for the Parameters can be defined outside rbf itself and then passed to the rbf parameter as a list? but I don't know how can be done this for the node or nodes if there are more that one.

This is an interesting idea. You could pass in a Parameter to be used as the value for each node, parameter, etc instead of the the numbers directly. It would make everything a lot more verbose I think. I'll mock up the JSON at some point; haven't got time now.

jetuk:
@Jmiguel17 are you still testing / using this? I think it needs finish or closing.

Jmiguel17:
Hi @jetuk, I have been testing this implementation in a couple of models and is working ok for me. However, so far I have used just the `RbfParameter` class. I have not used the `RbfVolumeParameter` class.

According to your points:

- [X] Decide which values should / should not be exposed as "variables" for the optimisation wrappers? 
I have been using as variables the `y` values and the `nodes`. I think this ok, I don't think we need to expose the `day_of_year` as a variable.
- [ ] Write a better numerical test?
I can help with this but we would have to discuss more this.
- [ ] Decided whether to complete the simpler version that justs uses volume?
tbh I have not checked the  `RbfVolumeParameter` class, I don't know what are the dif of this class and the `RbfParameter`.

jetuk:
A simpler version was introduced in #873. 

I think this needs some tidying up: docstrings completing, the second parameter is probably superseded by the RbfProfileParameter from #873. I need to do a proper review.

jetuk:
@Jmiguel17 is this likely to get tidied up, or should I close the MR?

PR #740, opened 2019-06-25
Interpolated Flow Parameter

MohammedBasheer:


pywr/parameters/parameters.py
```diff
@@ -130,6 +130,27 @@ def load(cls, model, data):
 InterpolatedVolumeParameter.register()
 
 
+class InterpolatedLevelParameter(AbstractInterpolatedParameter):
```
jetuk: > It interpolates a value based on node flow at previous timestep

I think this should be named different. It implies it interpolates level not flow at the moment. Perhaps `InterpolatedFlowParameter`

pywr/parameters/parameters.py
```diff
@@ -130,6 +130,27 @@ def load(cls, model, data):
...
+    """
+    Generic interpolation parameter for a non-storage node. It interpolates a value based on node flow at previous timestep
```
jetuk: Docstring should contain one sentence first, and then a paragraph with a more detailed explanation. We try to follow the numpy docstring guidance: https://numpydoc.readthedocs.io/en/latest/format.html

pywr/parameters/parameters.py
```diff
@@ -130,6 +130,27 @@ def load(cls, model, data):
...
+    """
+    def __init__(self, model, node, x, y, interp_kwargs=None, **kwargs):
+        super(InterpolatedLevelParameter, self).__init__(model, x, y, interp_kwargs, **kwargs)
```
jetuk: With Python 3.6+ we can simplify this to: `super().__init__(model, x, y, interp_kwargs, **kwargs)`

jetuk:
We ought to add a test for this in `tests/test_parameters.py` too.

pywr/parameters/parameters.py
MohammedBasheer: Yes, I agree with that

pywr/parameters/parameters.py
MohammedBasheer: Done

jetuk:
Thanks @MohammedBasheer . This is just missing a test. Do you think you can write one or should I?

MohammedBasheer:
> Thanks @MohammedBasheer . This is just missing a test. Do you think you can write one or should I?

MohammedBasheer:
I am not sure how and where to write a test. Would be great if you explained to me how to do it. 
Thanks

jetuk:
The tests are contained in the `tests` folder of the repository. The `test_parameters.py` file contains all the parameter tests and should be the place where a new test is added for this parameter. If you look for the tests related to the existing interpolation parameters you will see how those are implemented. We should do something similar for this parameter to ensure it functions as expected. This might require adding a new test model to the `tests/models/` directory.

To run the tests locally you can use pytest: `pytest tests` will run all of Pywr's tests. When you push a new commit to this branch the new test(s) will also be automatically rerun in the travis "checks" you can see here. 

Let me know if you need any more help.

MohammedBasheer:
I wrote a test for the parameter. I hope it is working fine now.

jetuk:
The test looks good. I'll try it out locally later today. Could you remove the `built.bat`, `build.txt` and `GIT_VERSION.txt` files from this PR? You'll need to use `git rm build.bat --cached` for example then make a commit to untrack them. Don't worry about all the extra commits, I'll squash them when this is merged.

It's not a bad idea to idea `build.bat`, but that should be in a separate PR - I think there's an issue about it somewhere.

jetuk:
Thanks @MohammedBasheer for this contribution.

MohammedBasheer:
> Thanks @MohammedBasheer for this contribution.

Thanks for taking the time to review it.

PR #741, opened 2019-06-25
Added a coefficient of determination recorder

MohammedBasheer:
In addition to adding the coefficient of determination as a recorder, I added calibration recorders that work for a storage node.

### DIFF
--- a/pywr/recorders/calibration.py
+++ b/pywr/recorders/calibration.py
@@ -12,7 +12,7 @@ Several different evaluation metrics are implemented in this module. Many are
 
 
 """
-from .recorders import NumpyArrayNodeRecorder
+from .recorders import NumpyArrayNodeRecorder, NumpyArrayStorageRecorder
 import numpy as np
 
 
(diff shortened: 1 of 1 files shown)


pywr/recorders/calibration.py
```diff
@@ -82,3 +82,84 @@ def values(self):
         return 1.0 - np.sum((obs-mod)**2, axis=0)/np.sum((obs-obs_mean)**2, axis=0)
 
 
+class CoefficientOfDeterminationNodeRecorder(AbstractComparisonNodeRecorder):
+    """ Recorder evaluates the Coefficient of Determination of model and observed """
+    def values(self):
+        mod = self.data        
+        obs = self._aligned_observed
+        n_items = len(mod)    
+        return ((n_items*np.sum(mod*obs, axis=0)-np.sum(mod, axis=0)*np.sum(obs, axis=0))/(np.sqrt((n_items*np.sum(mod**2, axis=0)-(np.sum(mod, axis=0))**2)*(n_items*np.sum(obs**2, axis=0)-(np.sum(obs, axis=0))**2))))**2
```
jetuk: Is it possible to split this calculation over multiple lines to improve readability?

pywr/recorders/calibration.py
```diff
@@ -82,3 +82,84 @@ def values(self):
...
+        self._aligned_observed = None
+
+    def setup(self):
+        super(AbstractComparisonStorageRecorder, self).setup()
+        # Align the observed data to the model
+        from pywr.parameters import align_and_resample_dataframe
+        self._aligned_observed = align_and_resample_dataframe(self.observed, self.model.timestepper.datetime_index)
+
+
+class RootMeanSquaredErrorStorageRecorder(AbstractComparisonStorageRecorder):
```
jetuk: This is great. I do wonder if we should make both sets of recorders use a shared set of functions to do the metric calculations? That might be more maintainable. I'll think about it.

pywr/recorders/calibration.py
```diff
@@ -82,3 +82,84 @@ def values(self):
...
+        obs_mean = np.mean(obs, axis=0)
+        return 1.0 - np.sum((obs-mod)**2, axis=0)/np.sum((obs-obs_mean)**2, axis=0)
+
+class CoefficientOfDeterminationStorageRecorder(AbstractComparisonStorageRecorder):
+    """ Recorder evaluates the Coefficient of Determination of model and observed """
+    def values(self):
+        mod = self.data        
+        obs = self._aligned_observed
+        n_items = len(mod)    
+        return ((n_items*np.sum(mod*obs, axis=0)-np.sum(mod, axis=0)*np.sum(obs, axis=0))/(np.sqrt((n_items*np.sum(mod**2, axis=0)-(np.sum(mod, axis=0))**2)*(n_items*np.sum(obs**2, axis=0)-(np.sum(obs, axis=0))**2))))**2
```
jetuk: Add a newline at the end of the file.

jetuk: Also see previous comment on making this line more readable.

PR #745, opened 2019-07-05
use metaclass kwarg for NodeMeta inheritance 

Batch21:


jetuk:
LGTM. Test failure on travis is due to new version of pandas removing `sheetname` keyword from `read_excel`.

PR #747, opened 2019-07-05
Fix pandas keyword argument for read_excel().

jetuk:


jetuk:
@Batch21 if you review this, we can merge this and it should fix the tests in #745.

PR #749, opened 2019-07-17
Fix loading PiecewiseLink with parameters from JSON.

### LINKED ISSUE #748: Updating PiecewiseLink values every timestep
I am trying to utilize the piecewise type for a node. When I try to update the values per timestep 
 the PiecewiseLink does not pass the values from a class. Instead the kwargs hold the names of the class.

```
<= JSON File =>
{
     "name": "Donnells PH [node]",
      "type": "PiecewiseLink",
      "max_flow": [flow_class_val_1, flow_class_val_2, flow_class_val_3, flow_class_val_4],
      "cost": [cost_class_val_1, cost_class_val_2, cost_class_val_3, cost_class_val_4]
},

<= In _init_ piecewise =>
# Would print out "flow_class_val_1" instead of a numerical value
costs = kwargs.pop('cost')
```

jetuk:
The load method on PiecewiseLink was broken when given parameters. It was also a little ambiguous regarding the use of 'max_flow' or 'max_flows', and 'cost' or 'costs'. This is now made explicit to use the singular case. That is consistent with the init methods of the class. Plural keys are no longer supported.

Add a test for loading parameters on to PiecewiseLink sublinks.

Fixes #748

### DIFF
--- a/pywr/domains/river.py
+++ b/pywr/domains/river.py
@@ -146,6 +146,14 @@ class RiverSplit(MultiSplitLink):
 
         super(RiverSplit, self).__init__(*args, **kwargs)
 
+    @classmethod
+    def load(cls, data, model):
+        max_flow = load_parameter(model, data.pop('max_flow', None))
+        cost = load_parameter(model, data.pop('cost', 0.0))
+
+        del(data["type"])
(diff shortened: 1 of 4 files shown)


jetuk:
@Batch21 thanks.

PR #750, opened 2019-07-18
Allow use of parameters as values in ControlCurveInterpolatedParameter

jetuk:
This copies the approach of ControlCurveParameter and allows the user
to provide either values as an iterable of floats or parameters as
an iterable of Parameter instances. Each parameter's value is used
during the calculation in the same way as the current list of values.

Tests extended to check this behaviour and support loading from JSON.

### DIFF
--- a/pywr/parameters/_control_curves.pyx
+++ b/pywr/parameters/_control_curves.pyx
@@ -246,19 +246,23 @@ cdef class ControlCurveInterpolatedParameter(BaseControlCurveParameter):
         # return the interpolated value for the current level.
         cdef double current_pc = node._current_pc[scenario_index.global_id]
         cdef double weight
-        cdef double[:] values
+        cdef double[:] values  # y values to interpolate between in this time-step
 
         if self.parameters is not None:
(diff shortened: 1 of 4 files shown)


pywr/parameters/_control_curves.pyx
```diff
@@ -205,14 +208,27 @@ cdef class ControlCurveInterpolatedParameter(BaseControlCurveParameter):
...
+                raise ValueError('Length of values should be two more than the number of '
+                                 'control curves ({}).'.format(nvalues))
+            self.values = np.asarray(values)
+
+        elif parameters is not None:
+            if len(parameters) != nvalues:
+                raise ValueError('Length of parameters should be two more than the number of '
+                                 'control curves ({}).'.format(nvalues))
+            self.parameters = list(parameters)
+            for p in self.parameters:
```
matrosoe: Add comment explaining what this is for.

jetuk: I've added this comment. I also added it to the `ControlCurveParameter` init method too.

Issue #754, opened 2019-08-06
Question about the use of Blender link


BaptisteFrancois:
Hi, 
I would like to split the outflow of a link to different nodes. I noticed that Blender link class was made for that, although I am having hard time to figure out how to use it within a json file.

My understanding of the issue #234 is the Blender link is likely to be removed at some point. The use of an AggregatedNode is mentioned but as far as I understand the aggreagted node, they cannot be linked to a network.

Do you have any guidance to provide on how to implement the splitting?
The ration between the two nodes would be 75% and 25%.
Thank you.
Baptiste

snorfalorpagus:
For now you should not use the `Blender` node. I don't think it's constraint is respected by the solver at all. Instead you must use the `AggregatedNode`. There isn't any documentation for this yet so the best I can suggest is to read the tests.

https://github.com/pywr/pywr/blob/f8de20b28d59fc5437fa658a0f649018a413c315/tests/test_agg_constraints.py

For example, the test below checks that the aggregated node "agg" is correctly constraining the flow via nodes "A" and "B" to be equal. This has the effect of limiting the flow in "A" to 40, the same as "B", despite "Z" demanding 100.

As you mention, the aggregated node is not "connected" to the network in the conventional sense. The constraint it applies is however respected by the solver. In fact, it would be an error to attempt to connect it.

```python
def test_aggregated_node_two_factors(model):
    """Nodes constrained by a fixed ratio between flows (2 nodes)"""
    A = Input(model, "A")
    B = Input(model, "B", max_flow=40.0)
    Z = Output(model, "Z", max_flow=100, cost=-10)

    agg = AggregatedNode(model, "agg", [A, B])
    agg.factors = [0.5, 0.5]
    assert_allclose(agg.factors, [0.5, 0.5])

    A.connect(Z)
    B.connect(Z)

    model.run()

    assert_allclose(agg.flow, 80.0)
    assert_allclose(A.flow, 40.0)
    assert_allclose(B.flow, 40.0)
```

jetuk:
@BaptisteFrancois are you happy with this response? `Blender` has now been removed from the code based entirely.

BaptisteFrancois:
@jetuk @snorfalorpagus thank you both for your answer. 
I have found a work around that works well (that I can describe if you want) but your solution is definitely more elegant and certainly more efficient as well. 
[...]

PR #756, opened 2019-08-06
Added some docs for AggregatedNode

snorfalorpagus:
In response to #754.

### DIFF
--- a/docs/source/cookbook/aggregated_node.rst
+++ b/docs/source/cookbook/aggregated_node.rst
@@ -1,0 +1,89 @@
+Aggregated nodes
+----------------
+
+The `AggregatedNode` is a special node that allows constraints to be added to a model that apply to a group of (possibly disconnected) nodes, enforcing either a fixed ratio or fixed maximum.
+
+Constraining flow using a ratio
+===============================
+
+Aggregated nodes allows a constraint to be added that ensures the flow via two or more nodes conforms to a specific ratio. This is useful, for example, when modelling a blending constraint between multiple sources of water.
+
+In the example below the aggregated node "D" constrains the flow in nodes "A" and "B" to be equal (0.5 + 0.5 == 1). Due to the constraint the solution is flow from A = 40, B = 40, despite C demanding 100.
+
+Note that the aggregated node "D" is not connected to any nodes via an edge. The constraint is applied regardless.
+
+.. code-block:: javascript
+
+    "nodes": [
+        {
+            "name": "A",
+            "type": "input"
+        },
+        {
+            "name": "B",
+            "type": "input",
+            "max_flow": 40.0
+        },
+        {
+            "name": "C",
+            "type": "output",
+            "max_flow": 100.0,
+            "cost": -10.0
+        },
+        {
+            "name": "D",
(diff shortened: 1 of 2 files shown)


docs/source/cookbook/aggregated_node.rst
```diff
@@ -0,0 +1,89 @@
+Aggregated nodes
+----------------
+
+The `AggregatedNode` is a special node that allows constraints to be added to a model that apply to a group of (possibly disconnected) nodes, enforcing either a fixed ratio or fixed maximum.
```
jetuk: `.... enforcing either a fixed ratio of flow and/or a total maximum flow.`

Is this more accurate? I assume the solver supports minimum flow too?

docs/source/cookbook/aggregated_node.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Constraining flow using a ratio
+===============================
+
+Aggregated nodes allows a constraint to be added that ensures the flow via two or more nodes conforms to a specific ratio. This is useful, for example, when modelling a blending constraint between multiple sources of water.
+
+In the example below the aggregated node "D" constrains the flow in nodes "A" and "B" to be equal (0.5 + 0.5 == 1). Due to the constraint the solution is flow from A = 40, B = 40, despite C demanding 100.
```
jetuk: The `0.5 + 0.5 == 1` is not a requirement? It simply uses relative weights? One could also do this with factors of 50 and 50?

docs/source/cookbook/aggregated_node.rst
```diff
@@ -0,0 +1,89 @@
...
+        {
+            "name": "B",
+            "type": "input",
+            "max_flow": 40.0
+        },
+        {
+            "name": "D",
+            "type": "aggregated",
+            "nodes": ["A", "B"],
+            "max_flow": 60.0
```
jetuk: I guess we shoud also show an example that uses the "flow_weights" key.

jetuk: This can also be a parameter.

docs/source/cookbook/aggregated_node.rst
```diff
@@ -0,0 +1,89 @@
...
+            "name": "C",
+            "type": "output",
+            "max_flow": 100.0,
+            "cost": -10.0
+        },
+        {
+            "name": "D",
+            "type": "aggregated",
+            "nodes": ["A", "B"],
+            "factors": [0.5, 0.5]
```
jetuk: We should comment that these ratios must (currently) be constant for an entire simulation.

Batch21:
I was just playing around with aggregated nodes and noticed that if you try and add two nodes from the same route to an aggregated node it crashes Python with this c error:

```
glp_set_mat_row: i = 8; ind[2] = 1; duplicate column indices not allowed
Error detected in file ..\src\api\prob1.c at line 779
```

If you try and do this within a notebook, there is no indication about what went wrong. The kernel just cashes.

I know you probably shouldn't be doing this (or need to) but it might be worth putting a explicit warning in the doc string about it.

Issue #758, opened 2019-08-21
Panic when two nodes in the same route added to AggregatedNode


snorfalorpagus:
I was just playing around with aggregated nodes and noticed that if you try and add two nodes from the same route to an aggregated node it crashes Python with this c error:

```
glp_set_mat_row: i = 8; ind[2] = 1; duplicate column indices not allowed
Error detected in file ..\src\api\prob1.c at line 779
```

If you try and do this within a notebook, there is no indication about what went wrong. The kernel just cashes.

I know you probably shouldn't be doing this (or need to) but it might be worth putting a explicit warning in the doc string about it.

_Originally posted by @Batch21 in https://github.com/pywr/pywr/pull/756#issuecomment-523084921_

snorfalorpagus:
There is no way to catch this kind of error from GLPK so we need to check for this condition in `setup()`.

jetuk:
Does `glp_error_hook` help with catching these errors?

snorfalorpagus:
`glp_error_hook` sounds like something we should investigate (#759). I also think we should be catching these errors specifically though.

jetuk:
> I also think we should be catching these errors specifically though.

Agreed.

jetuk:
@Batch21 what happens when you do the same test with the `glpk-edge` solver?

Batch21:
There is no error when using `glpk-edge` For reference, I'm testing this with the aggreagted1.json model from the tests folder with an extra link node added:

``` 
"nodes": [
        {
            "name": "A",
            "type": "input",
            "max_flow": 50
        },
        {
            "name": "B",
            "type": "input",
            "max_flow": 50
        },
        {
            "name": "C",
            "type": "input",
            "max_flow": 50
        },
        {
            "name": "X",
            "type": "link",
            "max_flow": 50
        },
        {
            "name": "Z",
            "type": "output",
            "max_flow": 100,
            "cost": -100
        },
        {
            "name": "agg",
            "type": "AggregatedNode",
            "nodes": ["A", "B", "C", "X"],
            "factors": [1.0, 1.0, 1.0, 1.0],
            "flow_weights": [1.0, 1.0, 1.0, 1.0],
            "max_flow": 30.0,
            "min_flow": 5.0
        }
    ],
    "edges": [
        ["A", "X"],
        ["X", "Z"],
        ["B", "Z"],
        ["C", "Z"]
    ]

```

jetuk:
> There is no error when using `glpk-edge`

[...]

Issue #760, opened 2019-08-25
Refactor GLPK solvers


jetuk:
We have two LP implementations using GLPK. There is a lot of replicated code in both implementations. A refactor that puts the boiler plate (e.g. basis management, init, etc.) into a shared base class would improve maintainability and perhaps open up possibilities for making more LP implementations (e.g. MILP #702).

PR #762, opened 2019-08-27
WIP: Recover from GLPK errors using glp_error_hook/setjmp

### LINKED ISSUE #759: Use glp_error_hook to catch unhandled errors
Following #758, we could use `glp_error_hook` to catch unhandled errors in GLPK which would prevent the program exiting and instead raise an exception in Python.

> 6.1.13    glp_error_hook—install  hook  to  intercept  abnormaltermination
>Synopsis
>`void glp_error_hook(void (*func)(void *info), void *info);`
>Description
>The routine `glp_error_hook` installs a user-defined hook routine to inter-cept abnormal termination.The parameterfuncspecifies the user-defined hook routine.  It is calledfrom  the  routineglp_errorbefore  the  latter  calls  the  abort  function  toabnormally terminate the application program because of fatal error.  
>The parameter info is a transit pointer,  specified in the corresponding call tothe routine `glp_error_hook`;  it may be used to pass some information tothe hook routine.
>To uninstall the hook routine the parametersfuncandinfoshould bespecified as `NULL`.
> Usage note
> If the hook routine returns,  the application program is abnormally terminated.  To prevent abnormal termnation the hook routine may perform aglobal jump using the standard function `longjmp`, in which case the application program must call the routine `glp_free_env`.

Example usage below. We could put the setjmp in `solve()` before the call to GLPK.

https://www.tutorialspoint.com/c_standard_library/c_function_longjmp.htm

```c
#include <stdio.h>
#include <stdlib.h>
#include <setjmp.h>

int main () {
   int val;
   jmp_buf env_buffer;

   /* save calling environment for longjmp */
   val = setjmp( env_buffer );
   
   if( val != 0 ) {
      printf("Returned from a longjmp() with value = %s\n", val);
      exit(0);
   }
   printf("Jump function call\n");
   jmpfunction( env_buffer );
   
   return(0);
}

void jmpfunction(jmp_buf env_buf) {
   longjmp(env_buf, "tutorialspoint.com");
}
```

It's probably worth doing some reading to see if there are any issues with this in Cython.

snorfalorpagus:
So far I've only done this for the non-edge GLPK solver, in the setup() method. I guess it should be used in solve() also. It seems to catch the error raised in #758.

Closes #759.

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -1,5 +1,6 @@
 from libc.stdlib cimport malloc, free
 from libc.math cimport abs
+from libc.setjmp cimport setjmp, jmp_buf
 from cython.view cimport array as cvarray
 import numpy as np
 cimport numpy as np
@@ -104,6 +105,13 @@ cdef class CythonGLPKSolver:
         cdef int status
         cdef cross_domain_row
         cdef int n, num
+        cdef jmp_buf env_buffer
+        cdef int jmpval
+
+        glp_error_hook(error_hook, &env_buffer)
+        jmpval = setjmp(env_buffer)
+        if jmpval != 0:
+            raise GLPKError("An unexpected error was raised by GLPK")
 
         self.all_nodes = list(sorted(model.graph.nodes(), key=lambda n: n.fully_qualified_name))
         if not self.all_nodes:
@@ -587,7 +595,10 @@ cdef class CythonGLPKSolver:
         # Set the basis for this scenario
         self._set_basis(scenario_index.global_id)
         # attempt to solve the linear programme
+
         simplex_ret = simplex(self.prob, self.smcp)
+
+
         status = glp_get_status(self.prob)
         if (status != GLP_OPT or simplex_ret != 0) and self.retry_solve:
             # try creating a new basis and resolving
--- a/pywr/solvers/glpk.pxi
+++ b/pywr/solvers/glpk.pxi
@@ -1,4 +1,5 @@
 from libc.float cimport DBL_MAX
+from libc.setjmp cimport longjmp, jmp_buf
 
 status_string = [
     None,
@@ -44,6 +45,14 @@ cdef int term_hook(void *info, const char *s):
     else:
         print(message)
     return 1
+
+
+class GLPKError(Exception):
+    pass
+
+
+cdef void error_hook(void *info):
+    longjmp((<jmp_buf*>info)[0], <int>1)
 
 
 cdef inline int constraint_type(double a, double b):
@@ -153,6 +162,7 @@ cdef extern from "glpk.h":
     int glp_get_status(glp_prob *P)
     int glp_term_out(int flag)
     void glp_term_hook(int (*func)(void *info, const char *s), void *info)
+    void glp_error_hook(void (*func)(void *info), void *info)
 
     double glp_get_row_prim(glp_prob *P, int i)
     double glp_get_col_prim(glp_prob *P, int j)
--- a/tests/models/aggregated1.json
+++ b/tests/models/aggregated1.json
@@ -20,6 +20,16 @@
             "max_flow": 50
         },
         {
+            "name": "C",
+            "type": "input",
+            "max_flow": 50
+        },
+        {
+            "name": "X",
+            "type": "link",
+            "max_flow": 50
+        },
+        {
             "name": "Z",
             "type": "output",
             "max_flow": 100,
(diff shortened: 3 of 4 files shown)


pywr/solvers/glpk.pxi
```diff
@@ -46,6 +47,14 @@ cdef int term_hook(void *info, const char *s):
     return 1
 
 
+class GLPKError(Exception):
+    pass
+
+
+cdef void error_hook(void *info):
+    longjmp((<jmp_buf*>info)[0], <int>1)
```
jetuk: GLPK example calls `glp_free_env` here.

pywr/solvers/cython_glpk.pyx
```diff
@@ -104,6 +105,13 @@ cdef class CythonGLPKSolver:
         cdef int status
         cdef cross_domain_row
         cdef int n, num
+        cdef jmp_buf env_buffer
+        cdef int jmpval
+
+        glp_error_hook(error_hook, &env_buffer)
+        jmpval = setjmp(env_buffer)
+        if jmpval != 0:
+            raise GLPKError("An unexpected error was raised by GLPK")
```
jetuk: Do we need to call `glp_error_hook(NULL, NULL)` when we're finished? Maybe in `__dealloc__`?

pywr/solvers/cython_glpk.pyx
```diff
@@ -587,7 +595,10 @@ cdef class CythonGLPKSolver:
         # Set the basis for this scenario
         self._set_basis(scenario_index.global_id)
         # attempt to solve the linear programme
+
         simplex_ret = simplex(self.prob, self.smcp)
+
+
```
jetuk: Extra line.

jetuk:
This is cool.

> I guess it should be used in solve() also.

Why wouldn't it work in solve()? The hook is still there during solve?

I added some minor comments after looking at this example: https://en.wikibooks.org/wiki/GLPK/Error_handling

In that example they use a simple struct to pass some text to the error hook function. I am not sure we need to do that? However, it might be good if we can raise a different error message for the different parts of our setup & solve. E.g. "An unexpected error was raised by GLPK when setting up the aggregated nodes." 

I did wonder if you could create a context manager for the life-cycle of the hook, but this might be pretty slow during solve. This is probably unnecessary here though. 

```python
with GlpkErrorManager("An unexpected error was raised by GLPK"): 
    # do glp_xxx stuff.
```

Also we should add a test for this. Add it reinforces the need for #760 so this can be used by both solvers.

pywr/solvers/glpk.pxi
snorfalorpagus: Good catch. It's quite explicit that this needs to be done (before longjmp!)

snorfalorpagus:
Any idea how we should test this? These kind of errors shouldn't normally be possible to trigger.

pywr/solvers/cython_glpk.pyx
snorfalorpagus: It occurs to me that this error hook must be global, rather than linked to a specific problem. What does it do when there are two models?

jetuk:
I think an easy one (besides the example from #758) is passing NaN as a max_flow. I think this causes one of these types of errors.

pywr/solvers/cython_glpk.pyx
jetuk: Indeed. Seems we have a choice. A global hook for all GLPK solvers, etc. Or add and remove the hook as needed, c.f. my comment about a context manager.

jetuk:
I was having a play with this today. Passing a NaN does create a catchable error. So this is a valid test to add. 

Calling `glp_free_env` (as recommended by the example) invalidates all the pointers. This includes the  `glp_prob` reference. Currently this is deleted using `glp_delete_prob` in `__dealloc__`. However, if the environment has been freed (i.e. there's been an internal GLPK error) the pointer is invalid and a seg fault is raised in the call to `glp_delete_prob`. 

I am not sure how to deal with this. In the case where the programme has multiple GLPK pointers (i.e. solvers) all of them will be invalidated. There does not appear to be a GLPK API method to determine if the environment has been freed. Therefore, the only choice is some global variable in our code to flag if this happens?

jetuk:
I've added the NaN test here.

jetuk:
> Therefore, the only choice is some global variable in our code to flag if this happens?

After further looking at this it does indeed seem that we must keep track globally of all our pointers to implement something like this. That is possible with some module level variables in Cython, but because we have two GLPK solvers it would make sense to solve #760 first.

jetuk:
#822 has caused conflicts with this, but should enable it to work!

Issue #764, opened 2019-09-04
Add templated parameter definitions


jetuk:
One of the issues we have with more complex Pywr models is the definition of custom parameters and recorders. There is a grey area as to whether these class definitions are part of a model's definition or just part of the infrastructure / library to run a model. Pywr provides many of these classes out of the box, but it is often the case that a specific model has a specific rule, licence or calculation that is only going to be used in that model. In this case there is a strong argument that these definitions should be considered as part of the model's definition.

Therefore, we should consider adding definition of custom parameters to the Pywr JSON schema. Broadly there are two approaches to this that I want to set out:

**Arbitrary Python file**

One could extend the JSON schema to support evaluating python source files. These could be added to the existing includes section:

```json
{
  "includes": ["my_parameters.py"]
}
```

In this case Pywr would simple `eval()` or `import` the contents of `my_parameters.py`. Any class definitions in there would be available as is. 

**A templated approach**

The downside with the above is that users would need to be fairly competent Python programmers to write a full Parameter class definition including all the boiler plate. An alternative approach is to provide a template for a parameter (or recorder) class (perhaps using `jinja2`). The user would then provide only the code for specific parts of this templates. For example:

```python
class {{class_name}}(Parameter):
    def __init__(self, model,
                 {% for arg in arguments %}
                 arg,
                 {% endfor %}
                 **kwargs):
        super().__init__(model, **kwargs)
        # Perhaps provide some basic type manipulation here?
        # e.g. cast to array using np.asarray()
        {% for arg in arguments %}
        self.{{arg}} = {{arg}}
        {% endfor %}

    def setup(self):
        super().setup()
        {{ setup_code }}

    def reset(self):
        {{ reset_code }}

    def value(self, ts, si):
        {{ value_code }}

    def before(self):
        {{ before_code }}

    def after(self):
        {{ after_code }}
{{class_name}}.register()
```

A JSON schema would then define the variables for these sections:

```python
{
    "class_name": "my_new_parameter",
    "arguments": [
        "mean",
        "amplitudes",
        "phase"
    ],
    "value_code": """
    value = self.mean 
    for a, p in zip(self.amplitudes, self.phase):
        value += a*np.cos(2*np.pi*timestep.dayofyear/365 + p)       
    return value
    """
}
```
And a list of these definitions provided as a top level key in the main JSON schema. 

** Security **

[...]

PR #765, opened 2019-09-05
Initial implementation of including python modules.

jetuk:
This approach simply detects the file extension of filenames referenced in the existing "includes" section. For .py files runpy is used to execute the code.

Begins to solve #764.

### DIFF
--- a/pywr/_model.pyx
+++ b/pywr/_model.pyx
@@ -207,29 +207,45 @@ class Model(object):
         """
         if "includes" in data:
             for filename in data["includes"]:
+                _, ext = os.path.splitext(filename)
                 if path is not None:
                     filename = os.path.join(os.path.dirname(path), filename)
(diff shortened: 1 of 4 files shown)


jetuk:
After reading a bit more about `run_path` and `importlib` I think this is probably the preferred approach. Using `importlib` would require modifying `sys.path`. `run_path` does this automatically and in a non-permanent manner. It does use `exec` underneath, but seems like a simple and clean approach for us to use.

jetuk:
> I had a go at using it for custom cython parameters and I just needed to import them into the .py file that was in the includes list. Would this be your approach for this?

I think this might be a bit flaky depending on where the files are. However, I don't have a better solution really. Generally I think the preferred approach is to have any extra things installed and imported in to the environment you are running Pywr in. 

I'll merge this as is.

PR #767, opened 2019-09-09
Fix CSVRecorder volume bug

### LINKED ISSUE #766: CSV recorder saves netflow not volume for Storage nodes.
This bug was introduced in #530 with the change from `Node` to `AbstractNode` checking. All nodes inherit from `AbstractNode` so the volumes of storage nodes are never saved. 

Easy fix is to swap the order of the type checking and add a regression test.

jetuk:
Fix #766 and tidy-up some old Python 2 stuff.

### DIFF
--- a/tests/test_recorders.py
+++ b/tests/test_recorders.py
@@ -555,23 +555,15 @@ def test_csv_recorder(simple_storage_model, tmpdir, complib):
     model.run()
 
     import csv
-
-    if sys.version_info.major >= 3:
-        kwargs = {"encoding": "utf-8"}
-        mode = "rt"
-    else:
-        kwargs = {}
-        mode = "r"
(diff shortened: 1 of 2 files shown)


tests/test_recorders.py
```diff
@@ -531,24 +531,21 @@ def test_concatenated_dataframes(simple_storage_model):
...
-
-    """
-    model = simple_linear_model
+    model = simple_storage_model
     otpt = model.nodes['Output']
-    model.nodes['Input'].max_flow = 10.0
     otpt.cost = -2.0
 
     # Rename output to a unicode character to check encoding to files
     if sys.version_info.major >= 3:
```
Batch21: Check not needed now python 2 support has been dropped?

jetuk: Could spot. I'll remove this too.

PR #788, opened 2019-10-28
pop comment from data passed to pandas.read_xxxxx

rheinheimer:
This prevents passing a "comment" argument to Pandas when reading from an external source.


jetuk:
Thanks for this. Happy to merge.

PR #789, opened 2019-10-28
Add year and ordinal threshold params

jetuk:
These two parameters allow for a comparison against the current year and ordinal day of the simulation respectively. They are useful for triggering things at specific dates in the simulation.

Also 
 - adds support for testing index values from the assertion recorder.
 - removes an old `basestring` usage in `_thresholds.pyx`


jetuk:
I am not sure what's up with appveyor here. I've not replicated this error on my Windows machine.

jetuk:
Looks like a pytables issue. See #790

jetuk:
I'm going to merge this despite the appveyor failures. I've tested locally on Windows, and we know that the failure is to do with something else.

PR #790, opened 2019-10-29
Depending on tables<3.6.1 due to bug in package in Windows.

jetuk:
This is a test for appveyor. I suspect a problem with tables-3.6.1, but need to confirm.


jetuk:
This has now resolved itself. So this change is not needed.

PR #796, opened 2019-11-10
Add Python 3.8 to CI builds.

jetuk:


jetuk:
Looks like this is missing `pygmo` dependency for 3.8. There are no builds for it yet.

snorfalorpagus:
Added an issue to the pygmo tracker for this: https://github.com/esa/pagmo2/issues/379

We could make this an optional dependency for the tests using `pytest.mark.skipif`.

jetuk:
> We could make this an optional dependency for the tests using `pytest.mark.skipif`.

I think this is the best solution given it is an entirely optional dependency for us.

snorfalorpagus:
It looks like we already do this with

```python
pygmo = pytest.importorskip("pygmo")
```

However the issue is it never gets that far, because this fails:

```
pip install platypus-opt inspyred pygmo
```

I think the solution in this case is to allow the install to fail:

```
pip install platypus-opt inspyred pygmo || true
```

snorfalorpagus:
The appveyor failure is actually something completely different:

```
    C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -Ic:\python38-x64\include -Ic:\python38-x64\include "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\winrt" /Tcwinpty/cywinpty.c /Fobuild\temp.win-amd64-3.8\Release\winpty/cywinpty.obj
    cywinpty.c
    winpty/cywinpty.c(641): fatal error C1083: Cannot open include file: 'winpty.h': No such file or directory
    error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\x86_amd64\\cl.exe' failed with exit status 2
```

snorfalorpagus:
In the very short term we could flag these jobs as allowed to fail and merge this.

jetuk:
My guess is that appveyor is trying to build pygmo from source and is failing.

I'll make the change to the build script to allow the install failures to occur.

jetuk:
@snorfalorpagus any guess on this issue? I've tried to replicate this locally, but struggling with cython versioning issues with the dependencies. Looks like 3.8 has worked on travis fine.

jetuk:
I think this is good to go now?

Issue #794, opened 2019-11-06
scenario parameters for daily and weekly profiles


Batch21:
Pywr has a `ScenarioMonthlyProfileParameter` but not corresponding parameters for daily and weekly profiles.

I've just created a daily version. Is it worth adding this and a weekly version to pywr?

jetuk:
Hmm I thought we got rid of this! I don't see why these wouldn't be useful additions.

You might be interested in the `ScenarioWrapperParameter`, but that is unlikely to be very efficient.

Batch21:
got rid of the `ScenarioMonthlyProfileParameter` ?

I had a look at the `ScenarioWrapperParameter` but, like you say, it didn't seem efficient for 50+ profiles

jetuk:
Yeah, I was mistaken. Must have been something else we removed!

jetuk:
These have been merged.

PR #802, opened 2019-11-16
daily and weekly scenario profiles

Batch21:
As discussed in #794

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -636,7 +636,7 @@ MonthlyProfileParameter.register()
 
 
 cdef class ScenarioMonthlyProfileParameter(Parameter):
-    """ Parameter which provides a monthly profile per scenario
+    """ Parameter that provides a monthly profile per scenario
 
     Behaviour is the same as `MonthlyProfileParameter` except a different
     profile is returned for each ensemble in a given scenario.
@@ -671,7 +671,8 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
 ScenarioMonthlyProfileParameter.register()
 
 cdef class ScenarioWeeklyProfileParameter(Parameter):
-    """
+    """Parameter that provides a weekly profile per scenario
+
     This parameter provides a repeating annual profile with a weekly resolution. A
     different profile is returned for each member of a given scenario
     
@@ -680,12 +681,13 @@ cdef class ScenarioWeeklyProfileParameter(Parameter):
     scenario: Scenario
         Scenario object over which different profiles should be provided.
     values : iterable, array
-        length of 1st dimension should equal the number of members in the scenario object 
+        Length of 1st dimension should equal the number of members in the scenario object 
         and the length of the second dimension should be 52
 
     """
     def __init__(self, model, Scenario scenario, values, *args, **kwargs):
         super().__init__(model, *args, **kwargs)
+        values = np.array(values)
         if values.ndim != 2:
             raise ValueError("Factors must be two dimensional.")
         if scenario._size != values.shape[0]:
@@ -703,7 +705,7 @@ cdef class ScenarioWeeklyProfileParameter(Parameter):
 
     cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
         cdef int i = ts.dayofyear - 1
-        if not is_leap_year(<int>(ts.datetime.year)):
+        if not is_leap_year(ts.year):
             if i > 58: # 28th Feb
                 i += 1
         cdef Py_ssize_t week
@@ -717,7 +719,8 @@ cdef class ScenarioWeeklyProfileParameter(Parameter):
 ScenarioWeeklyProfileParameter.register()
 
 cdef class ScenarioDailyProfileParameter(Parameter):
-    """
+    """Parameter which provides a daily profile per scenario
+
     This parameter provides a repeating annual profile wiatth a daily resolution. A
     different profile is returned for each member of a given scenario
 
@@ -726,12 +729,13 @@ cdef class ScenarioDailyProfileParameter(Parameter):
     scenario: Scenario
         Scenario object over which different profiles should be provided
     values : iterable, array
-        length of 1st dimension should equal the number of members in the scenario object 
+        Length of 1st dimension should equal the number of members in the scenario object 
         and the length of the second dimension should be 366
 
     """
     def __init__(self, model, Scenario scenario, values, *args, **kwargs):
         super().__init__(model, *args, **kwargs)
+        values = np.array(values)
         if values.ndim != 2:
             raise ValueError("Factors must be two dimensional.")
         if scenario._size != values.shape[0]:
@@ -749,7 +753,7 @@ cdef class ScenarioDailyProfileParameter(Parameter):
 
     cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
         cdef int i = ts.dayofyear - 1
-        if not is_leap_year(<int>(ts.year)):
+        if not is_leap_year(ts.year):
             if i > 58: # 28th Feb
                 i += 1
         return self._values[scenario_index._indices[self._scenario_index], i]
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -719,9 +719,9 @@ cdef class ScenarioWeeklyProfileParameter(Parameter):
 ScenarioWeeklyProfileParameter.register()
 
 cdef class ScenarioDailyProfileParameter(Parameter):
-    """Parameter which provides a daily profile per scenario
+    """Parameter which provides a daily profile per scenario.
 
-    This parameter provides a repeating annual profile wiatth a daily resolution. A
+    This parameter provides a repeating annual profile with a daily resolution. A
     different profile is returned for each member of a given scenario
 
     Parameters
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -670,6 +670,45 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
 
 ScenarioMonthlyProfileParameter.register()
 
+cdef class ScenarioDailyProfileParameter(Parameter):
+    """ An annual profile consisting of daily values that varys by scenario.
+
+    This parameter provides a repeating annual profile with a daily resolution. A total of 366 values
+    must be provided. These values are coerced to a `numpy.array` internally.
+
+    Parameters
+    ----------
+    values : iterable, array
+        The 366 values that represent the daily profile.
+
+    """
+    def __init__(self, model, Scenario scenario, values, *args, **kwargs):
+        super(ScenarioDailyProfileParameter, self).__init__(model, *args, **kwargs)
+        #v = np.squeeze(np.array(values))
+        if values.ndim != 2:
+            raise ValueError("Factors must be two dimensional.")
+        if scenario._size != values.shape[0]:
+            raise ValueError("First dimension of factors must be the same size as scenario.")
+        if values.shape[1] != 366:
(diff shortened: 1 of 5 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
 
 ScenarioMonthlyProfileParameter.register()
 
+cdef class ScenarioWeeklyProfileParameter(Parameter):
+    """
+    This parameter provides a repeating annual profile with a weekly resolution. A
```
jetuk: Could you separate the docstring formatting into a single initial sentence on the first line, and then a following paragraph.

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+    scenario: Scenario
+        Scenario object over which different profiles should be provided.
+    values : iterable, array
+        length of 1st dimension should equal the number of members in the scenario object 
+        and the length of the second dimension should be 52
+
+    """
+    def __init__(self, model, Scenario scenario, values, *args, **kwargs):
+        super().__init__(model, *args, **kwargs)
+        if values.ndim != 2:
```
jetuk: I wonder if we should do `values = np.array(values)` first? This would allow you to init with a list of list for example. As it is you must init with a numpy array.

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+            raise ValueError("Factors must be two dimensional.")
+        if scenario._size != values.shape[0]:
+            raise ValueError("First dimension of factors must be the same size as scenario.")
+        if values.shape[1] != 52:
+            raise ValueError("52 values must be given for a weekly profile.")
+        self._values = values
+        self._scenario = scenario
+
+    cpdef setup(self):
+        super(ScenarioWeeklyProfileParameter, self).setup()
```
jetuk: Remove arguments to `super`

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+        # This setup must find out the index of self._scenario in the model
+        # so that it can return the correct value in value()
+        self._scenario_index = self.model.scenarios.get_scenario_index(self._scenario)
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int i = ts.dayofyear - 1
+        if not is_leap_year(<int>(ts.datetime.year)):
```
jetuk: I think we can use `ts.year` and drop the int casting.

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+            week = 51
+        else:
+            week = i // 7
+        return self._values[scenario_index._indices[self._scenario_index], week]
+
+ScenarioWeeklyProfileParameter.register()
+
+cdef class ScenarioDailyProfileParameter(Parameter):
+    """
+    This parameter provides a repeating annual profile wiatth a daily resolution. A
```
jetuk: Fix docstring formatting as above.

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+    """
+    This parameter provides a repeating annual profile with a weekly resolution. A
+    different profile is returned for each member of a given scenario
+    
+    Parameters
+    ----------
+    scenario: Scenario
+        Scenario object over which different profiles should be provided.
+    values : iterable, array
+        length of 1st dimension should equal the number of members in the scenario object
```
jetuk: Sentences start with a capital

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+    """
+    This parameter provides a repeating annual profile wiatth a daily resolution. A
+    different profile is returned for each member of a given scenario
+
+    Parameters
+    ----------
+    scenario: Scenario
+        Scenario object over which different profiles should be provided
+    values : iterable, array
+        length of 1st dimension should equal the number of members in the scenario object
```
jetuk: Sentences start with a capital

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+
+    cpdef setup(self):
+        super(ScenarioDailyProfileParameter, self).setup()
+        # This setup must find out the index of self._scenario in the model
+        # so that it can return the correct value in value()
+        self._scenario_index = self.model.scenarios.get_scenario_index(self._scenario)
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int i = ts.dayofyear - 1
+        if not is_leap_year(<int>(ts.year)):
```
jetuk: Is the cast to `int` needed here?

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
     model.run()
     assert_allclose(inpt.flow, 365)
 
+def test_scenario_daily_profile(simple_linear_model):
+
+    model = simple_linear_model
+    scenario = Scenario(model, 'A', 2)
+    values =  np.array([np.arange(366, dtype=np.float64), np.arange(366, dtype=np.float64)])
```
jetuk: Shouldn't we test with different values in each scenario?

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
...
+
+def test_scenario_daily_profile_leap_day(model):
+    """Test behaviour of daily profile parameter for leap years
+    """
+    inpt = Input(model, "input")
+    otpt = Output(model, "otpt", max_flow=None, cost=-999)
+    inpt.connect(otpt)
+
+    scenario = Scenario(model, 'A', 2)
+    values = np.array([np.arange(366, dtype=np.float64), np.arange(366, dtype=np.float64)])
```
jetuk: Shouldn't we test with different values in each scenario?

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
     model.run()
     assert_allclose(inpt.flow, 365)
 
+def test_scenario_daily_profile(simple_linear_model):
```
jetuk: Perhaps organise the two `ScenarioDailyProfileParameter` tests into a class for readability.

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
...
+        for i in range(scenario.size):
+            si = ScenarioIndex(i, np.array([i], dtype=np.int32))
+            np.testing.assert_allclose(p.value(ts, si), values[i, iday])
+
+def test_scenario_weekly_profile(simple_linear_model):
+
+    model = simple_linear_model
+    scenario = Scenario(model, 'A', 2)
+
+    v = np.arange(52, dtype=np.float64)
```
jetuk: Shouldn't we test with different values in each scenario?

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
...
+    values = np.array([v, v])
+    r = np.array([v for _ in range(0, 7)]).T.flatten()
+    results = np.array([r, r])
```
jetuk: Perhaps call this `expected_results` for clarity

pywr/parameters/_parameters.pyx
```diff
@@ -670,6 +670,92 @@ cdef class ScenarioMonthlyProfileParameter(Parameter):
...
+        self._scenario = scenario
+
+    cpdef setup(self):
+        super(ScenarioDailyProfileParameter, self).setup()
+        # This setup must find out the index of self._scenario in the model
+        # so that it can return the correct value in value()
+        self._scenario_index = self.model.scenarios.get_scenario_index(self._scenario)
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int i = ts.dayofyear - 1
```
jetuk: `ts.dayofyear` comes from Pandas I think. I assume it does not do this leap year calculation as we want to. I wonder if we should just make `ts.dayofyear` do this instead of repeating it in the parameters?

jetuk: Similarly the week number could be calculated on the `Timestep` as well. This way everything can use the same calculations?

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
...
+
+    model = simple_linear_model
+    scenario = Scenario(model, 'A', 2)
+    values =  np.array([np.arange(366, dtype=np.float64), np.arange(366, dtype=np.float64)])
+
+    p = ScenarioDailyProfileParameter(model, scenario, values)
+
+    model.setup()
+    # Iterate in time
+    for ts in model.timestepper:
```
jetuk: Did the assertion recorder and decorator not work here?

tests/test_parameters.py
```diff
@@ -324,6 +325,72 @@ def test_daily_profile_leap_day(model):
...
+    v = np.arange(52, dtype=np.float64)
+    values = np.array([v, v])
+    r = np.array([v for _ in range(0, 7)]).T.flatten()
+    results = np.array([r, r])
+
+    param = ScenarioWeeklyProfileParameter(model, scenario, values)
+
+    model.setup()
+    # Iterate in time
+    for ts in model.timestepper:
```
jetuk: Assertion recorder?

pywr/parameters/_parameters.pyx
Batch21: I tried before but got this error:

`RuntimeError: super(): __class__ cell not found`

Looks like it is caused by the problem identified in this open cython issue - https://github.com/cython/cython/issues/2912 

As `__class__` in `setup` does not refer to the object class it is not possible to use `super` with no args.

What should we do here? It does look a bit inconsistent using args for `super` in `setup` when they are not used in `__init__`

pywr/parameters/_parameters.pyx
Batch21: so add them as methods/properties of the `Timestep` object?

pywr/parameters/_parameters.pyx
jetuk: Yeah, OK. I forgot I was reading Cython!

pywr/parameters/_parameters.pyx
jetuk: Yes. I suggest doing this in a separate PR though. Maybe raise an issue about it from this discussion.

Batch21:
Think I covered all the review comments.

The parameters will load from json - they use the load method inherited from the base parameter as this deals with the scenario and values args. I've used the load method for one of the parameters in one of the existing tests to confirm this but I could add a separate tests that loads from a json.

tests/test_parameters.py
```diff
@@ -324,6 +325,90 @@ def test_daily_profile_leap_day(model):
...
+        week = min(day // 7, 51)
+        value = week + 1
+        if scenario_index.global_id == 1:
+            value *= 2
+        return value
+
+    model.setup()
+    model.run()
+
+def test_scenario_daily_profile_leap_day(model):
```
jetuk: This test is repeated above?

jetuk:
Test failure was due to coverage version issue that was fixed in master.

PR #803, opened 2019-11-17
Remove warning filters in core.py #797

### LINKED ISSUE #797: Remove warning filters in core.py
There are some warning filters in `pywr.core` that we should probably remove. They are masking some deprecation warnings in our tests at the moment. I think this kind of filtering should be in application code not library code (i.e. not in pywr).

https://github.com/pywr/pywr/blob/2b886178a4ff87cd11210d0fc8a3ca3844d18dc8/pywr/core.py#L2

m7142yosuke:
Fixes #797


jetuk:
Thanks for this contribution @m7142yosuke

Issue #808, opened 2019-12-19
setup.py argument logic is un-intuitive


knoxsp:
In setup.py, the logic for turning on and off glpk and lpsolve is not intuitive:
In order to turn off building with lpsolve, you must use the argument `--with-glpk`

A preferrable solution would be to explicitly turn off solvers instead of turning on the other ones, for example using `--no-lpsolve`

jetuk:
Yeah, I agree with this.

You can use the `PYWR_BUILD_GLPK` environment variable also, but I think that suffers from the same logic where specifying GLPK means lpsolve is not built.

jetuk:
Closed in #811

PR #811, opened 2019-12-21
Refactor of setup.py

### LINKED ISSUE #809: setup.py requires cython, numpy and packaging to be preinstalled
Before running `setup.py` you must first manually install numpy, cython and packaging. This isn't very user friendly. There should be some way around this.

snorfalorpagus:
Some refactoring of `setup.py`. It is no longer required to pre-install cython, numpy and packaging - they should be installed automatically.

Closes #809 and #808

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -81,7 +81,7 @@ build_script:
   - cmd: copy %GLPK_HOME%\COPYING pywr\.libs\licenses\COPYING
 
   # build pywr and create a wheel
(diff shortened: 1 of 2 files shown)


jetuk:
lpsolve on Windows used to work? What happened?

snorfalorpagus:
lpsolve probably does still work on Windows, but it's never been tested on Appveyor. If you look at the existing Appveyor config it builds `--with-glpk` which implicitly disables lpsolve. Perhaps another PR could add testing of lpsolve on Windows.

snorfalorpagus:
I've added lpsolve to appveyor.

PR #812, opened 2019-12-23
Refactor of GLPK solvers

### LINKED ISSUE #760: Refactor GLPK solvers

snorfalorpagus:
Refactor of common code in the GLPK solvers into a shared parent class.

Closes #760.

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -59,6 +59,7 @@ cdef class CythonGLPKSolver(GLPKSolver):
 
     def __init__(self, use_presolve=False, time_limit=None, iteration_limit=None, message_level='error',
                  save_routes_flows=False, retry_solve=False):
+        super().__init__(time_limit, iteration_limit, message_level)
         self.stats = None
         self.is_first_solve = True
         self.has_presolved = False
@@ -66,16 +67,6 @@ cdef class CythonGLPKSolver(GLPKSolver):
         self.save_routes_flows = save_routes_flows
         self.retry_solve = retry_solve
         self.basis_manager = BasisManager()
-
-        # Set solver options
-        glp_init_smcp(&self.smcp)
-        self.smcp.msg_lev = message_levels[message_level]
-        if time_limit is not None:
-            self.smcp.tm_lim = time_limit
-        if iteration_limit is not None:
-            self.smcp.it_lim = iteration_limit
-
-        glp_term_hook(term_hook, NULL)
 
     def setup(self, model):
         cdef Node input
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -15,8 +15,6 @@ import time
 
 include "glpk.pxi"
 
-inf = float('inf')
-
 cdef class AbstractNodeData:
     cdef public int id
     cdef public bint is_link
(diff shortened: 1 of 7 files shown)


pywr/solvers/cython_glpk_edge.pyx
```diff
@@ -6,24 +6,21 @@ cimport numpy as np
...
 cimport cython
 
+from .glpk cimport GLPKSolver, BasisManager, simplex, set_obj_coef, set_row_bnds, set_col_bnds, set_mat_row
+
 from pywr._core import BaseInput, BaseOutput, BaseLink
 from pywr._core cimport *
 from pywr.core import ModelStructureError
 import time
 
 include "glpk.pxi"
```
jetuk: Is there a reason to keep `glpk.pxi` at all? Why can't all the function definitions be in the `glpk.pxd`?

pywr/solvers/cython_glpk.pyx
```diff
@@ -577,15 +518,8 @@ cdef class CythonGLPKSolver:
...
-        # Apply presolve if required
-        if self.use_presolve and not self.has_presolved:
-            self.smcp.presolve = GLP_ON
-            self.has_presolved = True
-        else:
-            self.smcp.presolve = GLP_OFF
-
         # Set the basis for this scenario
-        self._set_basis(scenario_index.global_id)
+        self.basis_manager.set_basis(self.prob, self.is_first_solve, scenario_index.global_id)
```
jetuk: Can we abstract the whole section on solving the model and error handling (522-549 here) into the base class?

pywr/solvers/cython_glpk.pyx
```diff
@@ -395,48 +372,12 @@ cdef class CythonGLPKSolver:
...
-            for i in range(nrows):
-                glp_set_row_stat(self.prob, i+1, self.row_stat[global_id, i])
-            for i in range(ncols):
-                glp_set_col_stat(self.prob, i+1, self.col_stat[global_id, i])
-
     def reset(self):
         # Resetting this triggers a crashing of a new basis in each scenario
         self.is_first_solve = True
 
     cpdef object solve(self, model):
```
jetuk: Can this be done in the base class? Looks the same between the two implementations?

pywr/solvers/cython_glpk.pyx
```diff
@@ -395,48 +372,12 @@ cdef class CythonGLPKSolver:
...
-            # otherwise we restore basis from previous solve of this scenario
-            nrows = glp_get_num_rows(self.prob)
-            ncols = glp_get_num_cols(self.prob)
-
-            for i in range(nrows):
-                glp_set_row_stat(self.prob, i+1, self.row_stat[global_id, i])
-            for i in range(ncols):
-                glp_set_col_stat(self.prob, i+1, self.col_stat[global_id, i])
-
     def reset(self):
```
jetuk: Can this be done in the base class? Looks the same between the two implementations?

snorfalorpagus:
The choice to leave the `reset` and `solve` methods in the child classes was a deliberate one. Neither of those functions are very long, so the duplication isn't too painful. These methods are identical in the two existing solvers but I can imagine others where they are different. For example, a solver than handles multiple scenarios in the same LP. The choice to keep `glpk.pxi` separate is for the same reason - perhaps someone would write a solver that works completely differently but still wants to use GLPK. Also it needs to be in a pxi rather than pxd because there are a couple of helper functions in there still.

PR #813, opened 2019-12-23
Added __contains__ method to NamedIterator and NodeIterator

### LINKED ISSUE #786: support 'mynode' in self.model.nodes
At some point I need to conditionally check if a node is actually in my model, using:
```python
if 'mynode' in self.model.nodes:
    n = self.model.nodes['mynode']
    # do something
else:
    # do something else
```
This doesn't seem to work - even if `mynode` is known to be a node, `'mynode' in self.model.nodes` returns false. I recommend supporting this, as a low priority nice-to-have feature request to file away. In the meantime, `try-except` works fine.

snorfalorpagus:
It's now possible to test for node and component membership using their names:

```python
assert "River Thames" in model.nodes
assert "Demand" in model.parameters
```

If a reference to the actual node is required, using try-except or `.get` is preferable to avoid searching twice.

```python
node = model.nodes.get("River Thames")
if node:
    print(f"max_flow: {node.max_flow}")
else:
    print("Not found")
```

```python
try:
    node = model.nodes["River Thames"]
except KeyError:
    print("Not found")
else:
    print(f"max_flow: {node.max_flow}")
```

Closes #786.

### DIFF
--- a/docs/source/json.rst
+++ b/docs/source/json.rst
@@ -192,6 +192,43 @@ Instead of defining the data inline using the ``"values"`` property, external da
         }
     ]}
 
+
+Loading a JSON document
+~~~~~~~~~~~~~~~~~~~~~~~
+
+A Pywr JSON document can be loaded into a `Model` instance by using the `Model.load` class-method:
+
+.. code-block:: python
+    from pywr.model import Model
+    my_model = model.load('/path/to/my_model.json')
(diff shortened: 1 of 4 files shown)


jetuk:
@snorfalorpagus LGTM, but I added a section to the tutorial based on your comment at the top of this PR.

PR #822, opened 2020-01-11
Merge all GLPK solvers into one module.

### LINKED ISSUE #819: glp memory allocation error
After installing the latest changes on master I got the following error when trying to run a  model:

```
glp_free: memory allocation error
Error detected in file ..\src\env\alloc.c at line 72
```

This was originally with python 3.7 and glpk 4.65.  Also tried with a new Python 3.8 env and got the same error.

not sure if I'm missing something with the updates to the solvers and setup.py

jetuk:
This commit merges all the GLPK related code into a single extension module. This addresses an issue (#819) found when static linking to GLPK used. The GLPK library relies on some global  memory for its environment. When static linking is used this global memory is created for each extension module. This breaks the ability to share GLPK problem pointers (`*glp_prob`) between extension modules.

Rather than require dynamic linking this commit merges all GLPK related code to a single module. Therefore static linking can be used safely.

Fixes #819.

@Batch21 can you check if this works with your setup? Seems to for me.


Batch21:
checked the changes in a python 3.8 environment and the tests work for both the standard and edge glpk solvers

PR #824, opened 2020-01-15
Add .pxd files to package_data in setup.py

### LINKED ISSUE #823: Cython definitions are missing from package_data
We do not currently package the `.pxd` files. This makes it impossible to build other Cython modules against a version of Pywr installed using a wheel or sdist. General advice on packaging Cython packages is below. We also don't follow the advice of packaging the `.c` as well as the `.pyx`. 

https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules

jetuk:
Fixes #823.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,6 +121,10 @@ def long_description():
 def package_data():
     pkg_data = {
         "pywr.notebook": ["*.js", "*.css"],
+        'pywr': ['*.pxd'],
+        'pywr.parameters': ['*.pxd'],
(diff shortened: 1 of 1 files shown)


jetuk:
> Only query is whether there would be a situation where the conditional starting on line 129 overwrites the pywr entry of `pkg_data`?

Good spot. I'm not entirely sure why that conditional exists... I'll investigate.

PR #834, opened 2020-01-29
D3 update

Batch21:
Work to update the `draw_graph` function to use d3.v5 and generally improve its functionality


Batch21:
@jetuk as mentioned a couple of weeks ago this is now ready for review.

The main changes are to update to d3 to v5 and to add x/y coordinates to the node attribute table when this is displayed. d3 is now loaded by require.js in the javascript script rather than by using the ipython  Javascript object as this would not correctly load d3 v5 from the CDN.

One consequence of using require.js, is that the `draw_graph` function will now definitely not work when running it in a notebook in jupyter lab, which does not expose require.js as a global variable. However, I could not get the previous version of the function to run in jupyter lab either as jupyter lab deals with js evaluation differently to traditional notebooks (You need to install a plugin, which involves installing node.js first, which isn't really worth the effort). I can create a separate issue about this

jetuk:
This looks good to me. 

I have noticed another bug when testing this: #856

pywr/notebook/draw_graph.js
```diff
@@ -161,49 +150,53 @@ function tick() {
...
+                })
+                .enter()
+                .append('td')
+                .text(function (d) { return d.value; });
+
+        }).on("mouseout", function(){
+            d3.select(".table-tooltip").transition().delay(2000).remove();
+        });
+    {% endif %}
+})
```
jetuk: Add an empty new line.

pywr/notebook/__init__.py
```diff
@@ -255,7 +255,7 @@ def _draw_graph(model, width=500, height=400, labels=False, attributes=False, cs
             labels=labels,
             attributes=attributes,
             css=css.replace("\n","")
-        ),
-        lib="http://d3js.org/d3.v3.min.js",
+        )
     )
-    return js
+
+    return js
```
jetuk: Add an empty new line.

PR #836, opened 2020-01-30
Added UniformDrawdownProfileParameter

jetuk:
This commit adds a new parameter that returns a recurring profile that draws down from 1.0 to 0.0 over the course of a year. The day on which to start the drawdown is configurable with the same keywords
as to `AnnualVirtualStorage`.

Basic tests included. Also added a more complex `AnnualVirtualStorage` test model that uses the new `UniformDrawdownProfileParameter` as a control curve in a dynamic cost calculation.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -747,13 +747,13 @@ cdef class ScenarioWeeklyProfileParameter(Parameter):
 
     This parameter provides a repeating annual profile with a weekly resolution. A
     different profile is returned for each member of a given scenario
-    
+
     Parameters
     ----------
     scenario: Scenario
         Scenario object over which different profiles should be provided.
     values : iterable, array
-        Length of 1st dimension should equal the number of members in the scenario object 
+        Length of 1st dimension should equal the number of members in the scenario object
         and the length of the second dimension should be 52
 
     """
@@ -801,7 +801,7 @@ cdef class ScenarioDailyProfileParameter(Parameter):
     scenario: Scenario
         Scenario object over which different profiles should be provided
     values : iterable, array
-        Length of 1st dimension should equal the number of members in the scenario object 
+        Length of 1st dimension should equal the number of members in the scenario object
         and the length of the second dimension should be 366
 
     """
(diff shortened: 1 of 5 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -833,6 +833,76 @@ cdef class ScenarioDailyProfileParameter(Parameter):
 ScenarioDailyProfileParameter.register()
 
 
+cdef class UniformDrawdownProfileParameter(Parameter):
+    """Parameter which provides a uniformly reducing value from one to zero.
+
+     This parameter is intended to be used with an `AnnualVirtualStorage` node to provide a profile
+     that represents perfect average utilisation of the annual volume. It returns a value one the
+     reset day, and subsequently reduces by 1/366 every day afterward.
```
Batch21: Should this be : 'It returns a value of 1 on the ....'

pywr/parameters/_parameters.pyx
```diff
@@ -833,6 +833,76 @@ cdef class ScenarioDailyProfileParameter(Parameter):
...
+        # Note that this is zero-based
+        self._reset_idoy = pandas.Period(year=2016, month=self.reset_month, day=self.reset_day, freq='D').dayofyear - 1
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
+        cdef int current_idoy = ts.dayofyear - 1
+        cdef int total_days_in_period
+        cdef int days_into_period
+        cdef int year = ts.year
+
+        if not is_leap_year(ts.year):
```
Batch21: use typed `year`  variable from above?

pywr/parameters/_parameters.pyx
```diff
@@ -833,6 +833,76 @@ cdef class ScenarioDailyProfileParameter(Parameter):
...
+        if days_into_period < 0:
+            # We're not past the reset day yet; use the previous year
+            year -= 1
+
+        if self._reset_idoy > 59:
+            # Reset occurs after February therefore next year's February might be a leap year?
+            year += 1
+
+        # Determine the number of days in the period based on whether there is a leap year or not in the current period
+        if is_leap_year(year):
```
Batch21: This is already check above. Could the contents of this conditional be shifted there?

jetuk: `year` could be modified by this point by the intervening checks on where we are relative to the possible leap day.

Batch21: yeah, good point!

pywr/parameters/_parameters.pyx
jetuk: `ts.year` should be a typed attribute because `ts` is typed.

jetuk: In fact I've confirmed by looking at the Cython `--annotate` output.

pywr/parameters/_parameters.pyx
jetuk: Fixed

PR #837, opened 2020-02-02
AnnualTotalFlowRecorder can now take a Factor. Test Added.

JackJohnson93:
AnnualTotalFlowRecorder can now take a Factor (as a dictionary of (node, factor) pairs. Test added and passing.

### DIFF
--- a/pywr/recorders/_recorders.pxd
+++ b/pywr/recorders/_recorders.pxd
@@ -147,7 +147,7 @@ cdef class AnnualTotalFlowRecorder(Recorder):
     cdef int _current_year
     cdef int _start_year
     cdef Aggregator _temporal_aggregator
-    cdef double [:] factors
+    cdef double [:] _factors
 
 cdef class AnnualCountIndexParameterRecorder(IndexParameterRecorder):
     cdef public int threshold
--- a/pywr/recorders/_recorders.pxd
+++ b/pywr/recorders/_recorders.pxd
@@ -147,7 +147,7 @@ cdef class AnnualTotalFlowRecorder(Recorder):
     cdef int _current_year
     cdef int _start_year
     cdef Aggregator _temporal_aggregator
-    cdef dict factors
+    cdef double [:] factors
 
 cdef class AnnualCountIndexParameterRecorder(IndexParameterRecorder):
     cdef public int threshold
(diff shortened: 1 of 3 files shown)


pywr/recorders/_recorders.pyx
```diff
@@ -1660,16 +1660,21 @@ cdef class AnnualTotalFlowRecorder(Recorder):
...
+    A list of factors can be provided to scale the total flow (e.g. for calculating operational costs).
+
     Parameters
     ----------
     model : `pywr.core.Model`
     name : str
         The name of the recorder
     nodes : list
         List of `pywr.core.Node` instances to record
+    factors (OPTIONAL) : dictionary
```
jetuk: This is good. However, I would make it a list where each entry corresponds to a node in the `nodes` list. Accessing this will be much faster than the dictionary.

pywr/recorders/_recorders.pyx
```diff
@@ -1660,16 +1660,21 @@ cdef class AnnualTotalFlowRecorder(Recorder):
...
+        Dictionary of (node, factor) for each factor to apply to each node
     """
     def __init__(self, model, str name, list nodes, *args, **kwargs):
         temporal_agg_func = kwargs.pop('temporal_agg_func', 'sum')
+        self.factors = kwargs.pop('factors', {})
```
jetuk: It will require you to handle no factors being supplied. In that case I guess it should default a list of 1.0s.

pywr/recorders/_recorders.pxd
```diff
@@ -147,6 +147,7 @@ cdef class AnnualTotalFlowRecorder(Recorder):
     cdef int _current_year
     cdef int _start_year
     cdef Aggregator _temporal_aggregator
+    cdef dict factors
```
jetuk: `cdef double[:] factors`

tests/test_recorders.py
```diff
@@ -1277,22 +1277,40 @@ def test_annual_count_index_threshold_recorder(simple_storage_model, params):
...
+        model.nodes['Input'].max_flow = 10.0
+        otpt.cost = -2
+        rec = AnnualTotalFlowRecorder(model, 'Total Flow', [otpt])
 
-    model.run()
+        model.run()
+
+        assert_allclose(3650.0, rec.data, atol=1e-7)
+
+    def test_annual_total_flow_recorder_factored(self, simple_linear_model):
```
jetuk: Thanks for the test.

pywr/recorders/_recorders.pyx
jetuk: This should also  be `factors : list, optional` as per numpy style guide : https://numpydoc.readthedocs.io/en/latest/format.html

PR #838, opened 2020-02-05
Optional factor added to NumpyArrayNode Recorder, default=1.0. Test a…

JackJohnson93:
…dded and passing.

Questions:
- Am I right to cdef factor as public double? Wasn't sure on the public part but copied elsewhere.
- Tried adding the Property method, wasn't sure if required? Seemed to work as planned in the test, i.e. allowing my_rec.factor = X

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -463,6 +463,8 @@ cdef class NumpyArrayNodeRecorder(NodeRecorder):
         Aggregation function used over time when computing a value per scenario. This can be used
         to return, for example, the median flow over a simulation. For aggregation over scenarios
         see the `agg_func` keyword argument.
+    factor: float (default=1.0)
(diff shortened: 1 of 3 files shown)


pywr/recorders/_recorders.pyx
```diff
@@ -492,13 +496,17 @@ cdef class NumpyArrayNodeRecorder(NodeRecorder):
...
         for i in range(self._data.shape[1]):
-            self._data[ts.index, i] = self._node._flow[i]
+            self._data[ts.index, i] = self._node._flow[i]*self.factor
         return 0
 
     property data:
         def __get__(self, ):
             return np.array(self._data)
 
+    property factor:
```
jetuk: As discussed this property is unnecessary, and the reason the build is failing. It has the same name (`factor`) as the extension type attribute (in the .pxd). Making factor a `cdef public float` is sufficient to allow access from Python space in this case.

PR #851, opened 2020-03-02
Paper examples

jetuk:
This is a branch I've had for a while. It has some updates the examples and the analytical test to match those described in the paper. 

It also looks like I added a couple of features:
1. Timings for `Model.before` and `Model.after` to the results object.
2. A helper static method for the TablesRecorder that generates dataframes.


jetuk:
Closing in favour of #852.

PR #852, opened 2020-03-02
Paper examples

jetuk:
This is a branch I've had for a while. It has some updates to the examples and the analytical test to match those described in the paper. 

It also looks like I added a couple of features:
1. Timings for `Model.before` and `Model.after` to the results object.
2. A helper static method for the TablesRecorder that generates dataframes.


examples/thames/thames.py
```diff
@@ -0,0 +1,310 @@
...
+    values.append(stats_df.loc['time_taken'][0] - sum(values))
+    values = np.array(values) / sum(values)
+    labels.append('Other')
+    explode.append(0.0)
+
+    fig, (ax1, ax2) = plt.subplots(figsize=(12, 4), ncols=2, sharey='row',
+                                   gridspec_kw={'width_ratios': [2, 1]})
+
+    print(values, labels)
+    ax1.pie(values, explode=explode, labels=labels, autopct='%1.1f%%', startangle=90)
```
Batch21: When I ran this, 'time taken after' took 88% of the total time, which meant the labels for all the other components ended up on top of each other

examples/two_reservoir/two_reservoir_moea.py
```diff
@@ -43,19 +43,14 @@ def platypus_main(harmonic=False):
     wrapper = PlatypusWrapper(get_model_data(harmonic=harmonic))
 
     with platypus.ProcessPoolEvaluator() as evaluator:
-        algorithm = platypus.NSGAII(wrapper.problem, population_size=50, evaluator=evaluator)
+        algorithm = platypus.NSGAIII(wrapper.problem, population_size=50, evaluator=evaluator, divisions_outer=12)
```
Batch21: Any particular reason for switching the algorithm?

pywr/recorders/recorders.py
```diff
@@ -518,4 +519,16 @@ def finish(self):
         self._arrays = {}
         self._routes_flow_array = None
 
+    @staticmethod
+    def generate_dataframes(h5file, time='/time', scenarios='/scenarios'):
```
Batch21: Is there an argument that this should be a normal method? Would be quite nice to be able to use this on instances of the recorder directly

examples/thames/thames.py
jetuk: I think this is because of the `TablesRecorder`

examples/two_reservoir/two_reservoir_moea.py
jetuk: No.

pywr/recorders/recorders.py
jetuk: It was written as a static method so you don't have to make an instance of the recorder, and therefore model, when post-processing the results. I've realised this function doesn't set the column index when reading the data which is kind of annoying when there are multiple scenarios.

Batch21: yeah, I didn't notice that the scenario func arg is not used. Looks like you will also get an error if you use it on a file with no time node. Maybe just use the default Pandas range index in these cases.

jetuk: To be honest it needs some tests!

jetuk: @Batch21 I've added a test and improved the function. Can you check?

Batch21: looks good

PR #857, opened 2020-03-09
Add `ControlCurvePiecewiseInterpolatedParameter`

jetuk:
A new parameter the interpolates between multiple control curves in a piecewise fashion. A pair of values is given for each "band" between the storage bounds and the control curves. This pair is used to set linear interpolation based on the current storage.

### DIFF
--- a/pywr/parameters/_control_curves.pyx
+++ b/pywr/parameters/_control_curves.pyx
@@ -10,7 +10,7 @@ cdef extern from "numpy/npy_math.h":
 
 cdef class PiecewiseLinearControlCurve(Parameter):
     """Piecewise function composed of two linear curves
-    
+
     Parameters
     ----------
     model : Model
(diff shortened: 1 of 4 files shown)


pywr/parameters/_control_curves.pyx
```diff
@@ -306,6 +306,87 @@ cdef class ControlCurveInterpolatedParameter(BaseControlCurveParameter):
...
 ControlCurveInterpolatedParameter.register()
 
+
+cdef class ControlCurvePiecewiseInterpolatedParameter(BaseControlCurveParameter):
+    """A control curve Parameter that interpolates between two or more pairs of values.
+
+    Return values are linearly interpolated between a pair of values depending on the current
+    storage. The first pair is used between maximum and the first control curve, the next pair
+    between the first control curve and second control curve, and so on until the last pair is
+    used between the last control curve the minimum value. The first value in each pair is the
```
Batch21: ' and the minimum value'

jetuk:
Thanks for the review. I'll fix the docstring and merge.

> Is the plan to depreciated `PiecewiseLinearControlCurve` ?

The only thing wrong with it is the inconsistent entry of the values. This one is a more general version of `PiecewiseLinearControlCurve`. Do you think it should be deprecated? It would reduce maintenance / confusion by not having two things doing the same job. Some models would need to be move over.

Batch21:
Could you add a depreciation warning that it will be removed completely in couple of versions time (or in v2)?

jetuk:
Yes, we could do that.

jetuk:
@Batch21 comment addressed, and deprecation warning added. Could you approve (or not)?

PR #861, opened 2020-03-16
Fix issue with using AbstractStorage in control curve parameters.

jetuk:
Removed local type (cdef Storage) in favour of using attribute directly.


jetuk:
> why would you want to use an AbstractStorage here

It allows you to use it with any of the subclasses. E.g. `AggregatedStorage`

PR #863, opened 2020-03-17
registering ArrayIndexedScenarioParameter

Jmiguel17:
This parameter `ArrayIndexedScenarioParameter` was not registered.


jetuk:
Thanks for fixing this.

PR #864, opened 2020-03-19
Api constantscenarioparameter

Jmiguel17:
Hi @jetuk, I just implemented the methods `set_double_varibles` and `get_double_variables`. I was thinking to add `get_double_lower_bounds` and `get_double_upper_bounds` but this parameter does not have these properties (lower and upper bounds). I can add these other methods, however, do you think this parameter must have these properties? What was the initial thought (the use) with this parameter? What other methods do you think this parameter must have?

I am creating this MR to start iterating on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+*.eggs/
 # PyInstaller
 # Usually these files are written by a python script from a template
 # before PyInstaller builds the exe, so as to inject date/other infos into it.
@@ -66,3 +67,5 @@ venv/
 
 pywr_glpk_debug.lp
 pywr_glpk_debug.mps
+
+build.bat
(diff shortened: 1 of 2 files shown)


jetuk:
I think we ought to implement the whole thing so that then you can use this parameter in optimisation if you would like (which requires the bounds). Ideally we want to add a test equivalent to the one for `ConstantParameter`. That test is here:

https://github.com/pywr/pywr/blob/f560f8aeb784e4ac16ee2d3f2b3a5516bbe92ba1/tests/test_parameters.py#L58

The current `ConstantScenarioParameter` test is here. Try to make it similar to the test class for the regular constant (i.e. a class with two test functions).

https://github.com/pywr/pywr/blob/f560f8aeb784e4ac16ee2d3f2b3a5516bbe92ba1/tests/test_parameters.py#L134

Let me know if none of that makes sense.

jetuk:
@Jmiguel17 have you made any more progress on this?

Jmiguel17:
Hi @jetuk during this week I will finish this.

jetuk:
@Jmiguel17 are you still working on this?

Jmiguel17:
Sorry, I have been busy and have not allocated time to write the test. But I tested the implementation to couple two `pywr` models using `pynsim` and is working. I have some deadlines to meet this month :worried:, let me I pass them and I finish the test :+1:.

PR #866, opened 2020-03-31
Add count_nonzero aggregation function.

jetuk:
This is useful with, for example, the annual count recorders to get a sum of the years with nonzero values (i.e. a count of the number of years with failures).


jetuk:
Thanks @Batch21. I added a docstring to the `Aggregator` class as it is used all over. This is also added to the API docs.

PR #867, opened 2020-04-14
Add PywrRandomGenerator as a Platypus generator.

jetuk:
This generator seeds the initial population with a `Solution` that corresponds to the model's current configuration.


jetuk:
@Batch21 and I discussed this. Commented it would be good to support generating random models that are perturbations of the initial config as well.

PR #868, opened 2020-04-14
New notebook graph functionality 

Batch21:
I’ve moved the notebook `draw_graph` function into a class and added a method to the class that will read the graph and save the positions of any node moved by a user into a new model json file. This hopefully deals with the frustration of spending ages moving nodes around to make a model graph readable only to lose this work when the notebook is reloaded. I’ve also added a method that will save the d3 graph into an html file, which should make it easier to share.

Still needs some work but any initial comments @pywr/pushers ?

### DIFF
--- a/pywr/notebook/__init__.py
+++ b/pywr/notebook/__init__.py
@@ -51,11 +51,11 @@ class PywrSchematic:
         """
 
         if isinstance(model, Model):
-            self.graph = self.pywr_model_to_d3_json(model, attributes)
+            self.graph = pywr_model_to_d3_json(model, attributes)
             # TODO update when schema branch is merged
             self.json = None
         else:
-            self.graph = self.pywr_json_to_d3_json(model, attributes)
+            self.graph = pywr_json_to_d3_json(model, attributes)
             with open(model) as d:
                 self.json = json.load(d)
 
--- a/pywr/notebook/__init__.py
+++ b/pywr/notebook/__init__.py
@@ -99,7 +99,7 @@ class PywrSchematic:
             node positions to a csv file.
         """
 
-        if self.json is None:
+        if self.json is None and filetype == "json":
             warnings.warn("Node positions cannot be saved to json if PywrSchematic object has been instantiated using"
                           "a pywr model object. Please use a json file path or model dict instead", stacklevel=2)
         else:
--- a/pywr/notebook/__init__.py
+++ b/pywr/notebook/__init__.py
@@ -69,196 +69,6 @@ class PywrSchematic:
         else:
             self.css = css
 
-    def pywr_model_to_d3_json(self, model, attributes=False):
-        """
-        Convert a Pywr graph to a structure d3 can display
-
-        Parameters
-        ----------
-        model : `pywr.core.Model`
-        attributes: bool (default=False)
-            If True, attribute data for each node is extract
-        """
-        nodes = []
-        node_names = []
-        for node in model.graph.nodes():
-            if node.parent is None and node.virtual is False:
-                nodes.append(node)
-                node_names.append(node.name)
-
-        edges = []
-        for edge in model.graph.edges():
-            node_source, node_target = edge
-
-            # where a link is to/from a subnode, display a link to the parent instead
-            if node_source.parent is not None:
-                node_source = node_source.parent
-            if node_target.parent is not None:
-                node_target = node_target.parent
-
-            if node_source is node_target:
-                # link is between two subnodes
-                continue
-
-            index_source = node_names.index(node_source.name)
-            index_target = node_names.index(node_target.name)
-            edges.append({'source': index_source, 'target': index_target})
-
-        json_nodes = []
-        for n, node in enumerate(nodes):
-            node_dict = {"name": node.name}
-            classes = []
-            cls = node.__class__
-            classes.append(cls)
-            while True:
-                for base in cls.__bases__:
-                    if issubclass(base, Node) and base is not Node:
-                        classes.append(base)
-                if classes[-1] is cls:
-                    break
-                else:
-                    cls = classes[-1]
-            classes = classes[::-1]
-            node_dict["clss"] = [cls.__name__.lower() for cls in classes]
-            try:
-                node_dict["position"] = node.position["schematic"]
-            except KeyError:
-                pass
-
-            if attributes:
-                node_dict["attributes"] = self.get_node_attr(node)
-
-            json_nodes.append(node_dict)
-
-        graph = {
-            "nodes": json_nodes,
-            "links": edges}
-
-        return graph
-
-    def get_node_attr(self, node):
-        """
-        Returns a dictionary that contains node attributes as strings
-
-        Parameters
-        ----------
-        node : a pywr node object
-        """
-        attrs = inspect.getmembers(node, lambda a:not(inspect.isroutine(a)))
-        attribute_data = []
-        for att in attrs:
-
-            attr_name, attr_val = att
-            if attr_name.startswith("_"):
-                continue
-            attr_type = type(attr_val).__name__
-
-            attrs_to_skip = ["component_attrs", "components", "color", "model",  "input", "output",
-                            "inputs", "outputs", "sub_domain", "sub_output", "sublinks", "visible",
-                            "fully_qualified_name", "allow_isolated"]
-            if not attr_val or attr_name.lower() in attrs_to_skip:
-                continue
-
-            if isinstance(attr_val, Component):
-                attr_val = attr_val.name
-                if not attr_val:
-                    attr_val = attr_type
-                else:
-                    attr_val = attr_val + " - " + attr_type
-
-            if isinstance(attr_val, list):
-                new_vals = []
-                for val in attr_val:
-                    val_name = str(val)
-                    val_name = val_name.replace("[", "").replace("]", "")
-                    new_vals.append(val_name)
-                attr_val = "".join(new_vals)
-            else:
-                attr_val = str(attr_val)
-
-            attribute_data.append({"attribute": attr_name, "value": attr_val})
-
-        return attribute_data
-
-    def pywr_json_to_d3_json(self, model, attributes=False):
-        """
-        Converts a JSON file or a JSON-derived dict into structure that d3js can use.
-
-        Parameters
-        ----------
-        model : dict or str
-            str inputs should be a path to a json file containing the model.
-        """
-
-        if isinstance(model, str):
-            with open(model) as d:
-                model = json.load(d)
-
-        nodes = [node["name"] for node in model["nodes"]]
-
-        edges = []
-        for edge in model["edges"]:
-            sourceindex = nodes.index(edge[0])
-            targetindex = nodes.index(edge[1])
-            edges.append({'source': sourceindex, 'target': targetindex})
-
-        nodes = []
-        node_classes = self.create_node_class_trees()
-
-        for node in model["nodes"]:
-            json_node = {'name': node["name"], 'clss': node_classes[node["type"].lower()]}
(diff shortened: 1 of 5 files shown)


jetuk:
Tests failing because `template.html` is not packaged. It needs adding to the package date in `setup.py`.

jetuk:
@Batch21 if you fix the tests I'll give this a review.

Batch21:
@jetuk I've fixed the tests, which also involved moving the functions used to create the d3 graph data back out of the new schematic class. I'm having second thoughts about whether the `draw_graph` func should be in the class as it will break existing notebooks where it is used. Could have it duplicated as a function with a depreciation warning?

jetuk:
> involved moving the functions used to create the d3 graph data back out of the new schematic class.

Why does it involve doing this?

pywr/notebook/__init__.py
```diff
@@ -17,8 +16,110 @@
...
+
+
+class PywrSchematic:
+
+    def __init__(self, model, width=500, height=400, labels=False, attributes=False, css=None):
+        """This object contains methods that allow the graph of a pywr model network to be displayed in a jupyter
+        notebook or saved to an html file.
+
+        It also contains a method to save the node positions of a notebook graph back to a pywr model json file. Note
+        that this method currently odoes not work if the object has be instantiated using a model object.
```
jetuk, line 34: `odoes` typo

pywr/notebook/__init__.py
```diff
@@ -17,8 +16,110 @@
...
+            If True, each graph node is labelled with its name. If false, the node names are displayed
+            during mouseover events
+        attributes : bool
+            If True, a table of node attributes is displayed during mouseover events
+        css : string
+            Stylesheet data to use instead of default
+        """
+
+        if isinstance(model, Model):
+            self.graph = self.pywr_model_to_d3_json(model, attributes)
```
jetuk, line 54: Call to a method that doesn't exist? Should be the function instead?

pywr/notebook/__init__.py
```diff
@@ -17,8 +16,110 @@
...
+            # TODO update when schema branch is merged
+            self.json = None
+        else:
+            self.graph = self.pywr_json_to_d3_json(model, attributes)
```
jetuk, line 58: Call to a method that doesn't exist? Should be the function instead?

pywr/notebook/__init__.py
```diff
@@ -17,8 +16,110 @@
...
+                model_data=json.dumps(self.json),
+                height=self.height,
+                width=self.width,
+                filename=json.dumps(filename)
+            )))
+
+    def to_html(self, filename, title="Model Schematic"):
+        """Save an HTML file of schematic"""
+
+        # TODO add option to get node position from graph that has already been drawn in a notebook
```
jetuk, line 102: What would this look like? Is it hard?

pywr/notebook/__init__.py
```diff
@@ -106,8 +207,8 @@ def get_node_attr(node):
         attr_type = type(attr_val).__name__
 
         attrs_to_skip = ["component_attrs", "components", "color", "model",  "input", "output",
-                         "inputs", "outputs", "sub_domain", "sub_output", "sublinks", "visible",
-                         "fully_qualified_name", "allow_isolated"]
+                        "inputs", "outputs", "sub_domain", "sub_output", "sublinks", "visible",
```
jetuk, line 210: Why delete these spaces? PEP8 violation!

pywr/notebook/__init__.py
```diff
@@ -17,8 +16,110 @@
...
+        labels : bool
+            If True, each graph node is labelled with its name. If false, the node names are displayed
+            during mouseover events
+        attributes : bool
+            If True, a table of node attributes is displayed during mouseover events
+        css : string
+            Stylesheet data to use instead of default
+        """
+
+        if isinstance(model, Model):
```
jetuk, line 53: You could instead have a classmethod for each of these functions. E.g. `from_json` and `from_model`. Or just make them static methods. I think most notebooks use `draw_graph` which has been broken here anyway?

pywr/notebook/__init__.py
```diff
@@ -212,50 +313,3 @@ def create_node_class_trees():
         clss = [cls.__name__.lower() for cls in classes[::-1]]
         node_class_trees[name] = clss
     return node_class_trees
-
-
-def draw_graph(model, width=500, height=400, labels=False, attributes=False, css=None):
```
jetuk, line 217: We could replace this with a function that creates a schematic instance and calls `draw_graph`?

pywr/notebook/draw_graph.js
```diff
@@ -136,6 +137,12 @@ require(["d3"], function(d3) {
 
     function tick() {
 
+        node.attr("transform", function(d) {
```
jetuk, line 140: Add a comment that this is bounding the node positions to stay inside the box. 

An alternative to this is to retain the coordinates as is, but use a `viewBox` on the SVG element to view the box containing the model?

pywr/notebook/save_graph.js
```diff
@@ -0,0 +1,53 @@
...
+require(["d3"], function(d3) {
+
+    const model_data = {{ model_data }};
+
+    const width = {{ width }},
+        height = {{ height }};
+
+    const nodes = d3.select(".pywr_schematic").selectAll(".node").data();
+
+    const node_data_obj = {}
```
jetuk, line 14: ;

pywr/notebook/save_graph.js
```diff
@@ -0,0 +1,53 @@
...
+    const posX = d3.scaleLinear()
+                    .range([-100, 100])
+                    .domain([0, width]);
+    const posY = d3.scaleLinear() 
+                   .range([100, -100])
+                   .domain([0, height]);
+    
+    for (let i = 0; i < model_data.nodes.length; i++){
+        let node_name = model_data.nodes[i].name;
+        let node_data = node_data_obj[node_name]
```
jetuk, line 29: ;

pywr/notebook/__init__.py
```diff
@@ -17,8 +16,110 @@
...
+    def to_html(self, filename, title="Model Schematic"):
+        """Save an HTML file of schematic"""
+
+        # TODO add option to get node position from graph that has already been drawn in a notebook
+
+        js = draw_graph_template.render(
+            graph=self.graph,
+            width=self.width,
+            height=self.height,
+            element=json.dumps(".schematic"),
```
jetuk, line 108: Why do you need to `json.dumps` this string?

pywr/notebook/save_graph.js
```diff
@@ -0,0 +1,53 @@
...
+        if (node_data.fixed){
```
jetuk, line 30: Worth having an option to save all node positions even if not fixed?

pywr/notebook/save_graph.js
```diff
@@ -0,0 +1,53 @@
...
+            model_data["nodes"][i]["position"] = {"schematic": [posX(node_data.x), posY(node_data.y)]}
```
jetuk, line 31: This will nuke any existing position information on the node (e.g. "geographic" coordinates.)

pywr/notebook/save_graph.js
```diff
@@ -0,0 +1,53 @@
+
+
+require.config({paths: {d3: 'https://d3js.org/d3.v5.min'}});
+  
+require(["d3"], function(d3) {
```
jetuk, line 5: Add a comment / docstring

jetuk:
@Batch21 could you look in to #870 and merge if OK, then rebase this against the new master?

jetuk:
@Batch21 is this ready now?

pywr/notebook/__init__.py
Batch21: Without it, the string does not get properly rendered with `"` in the js file

Batch21:
@jetuk still need to respond to a few of the comment and tidy up

pywr/notebook/save_graph.js
Batch21: Not sure how to format this as the function is an anonymous one passed as a callback to require. The alternative could be to add some more comments to the code within the function?

pywr/notebook/__init__.py
Batch21: Done

pywr/notebook/__init__.py
Batch21: Do you mean converting the `to_d3` functions into static or class methods rather than standalone functions?

pywr/notebook/__init__.py
Batch21: I can't think of a way of doing it without making the existing `draw_graph.js` script even messier - might be better to have a separate script.

I think I might leave this and see if the method get used at all

Batch21:
@jetuk  - I've addressed most of your comments and asked for clarification on a few

pywr/notebook/__init__.py
```diff
@@ -82,38 +82,46 @@ def draw_graph(self):
...
 
-    def save_positions(self, filename="node_positions.csv"):
-        """Save a copy of the model json with update schematic positions"""
+        Parameters
+        ----------
+        filename: str
+            The name of the file to save the output data to.
+        save_unfixed: bool
+            If True, then all node position are saved to output file. If False, only nodes who have had their position
+            fixed in the d3 grpah have their positions saved.
```
jetuk, line 96: `grpah`

pywr/notebook/__init__.py
```diff
@@ -82,38 +82,46 @@ def draw_graph(self):
...
+        filetype: str
+            Should be either 'json' to save the model data with updated node positions to a json file or 'csv' to save
+            node positions to a csv file.
+        """
 
-        if self.json is None:
-            warnings.warn("Node positions cannot be saved to json if PywrSchematic object has been instantiated using"
-                          "a pywr model object. Please use a json file path or Python dict instead", stacklevel=2)
+        if self.json is None and filetype == "json":
+            warnings.warn("Node positions cannot be saved to json if PywrSchematic object has been instantiated using "
```
jetuk, line 103: `json` -> `JSON` in docstring

pywr/notebook/save_graph.js
```diff
@@ -4,54 +4,59 @@ require.config({paths: {d3: 'https://d3js.org/d3.v5.min'}});
...
                 node_data.fixed,
                 posX(node_data.x),
                 posY(node_data.y),
-            ]
-            output_data.push(position_data.join(","))
+            ];
+            output_data.push(position_data.join(","));
         }
-        download(filename, output_data.join("\n"));
+        // TODO append error message if filetype not recognised
```
jetuk, line 52: Yes, I guess this should raise an error, but what is the JS equivalent of that?

pywr/notebook/save_graph.js
```diff
@@ -4,54 +4,59 @@ require.config({paths: {d3: 'https://d3js.org/d3.v5.min'}});
...
     }
 
-    
+    if (filetype == "json"){
+        download(filename, JSON.stringify(model_data));
```
jetuk, line 56: Why did you move these calls to download to a separate if-else-if statement and not keep them in the above blocks?

pywr/notebook/save_graph.js
Batch21: I added a check for this in python instead

pywr/notebook/save_graph.js
Batch21: The check above is in a for loop which is why it is redone here

jetuk:
Thanks @Batch21 looks good, I'll merge this now. Hopefully we can give it a bit of road test before the next release. It would be good to think what to do with this good in the long term. We don't test it very well at the moment.

PR #869, opened 2020-04-18
Rework the Recorder constraint functionality to be more flexible.

jetuk:
Upper and lower bounds are now specified on `Recorders` instead of by a binary `is_constraint` flag. This allows the optimisation wrappers to handle equality, single bounded and double bounded constraints. It also allows the user to specify the constraint threshold. The previous implementation relied on the default setting of the respective optimisation wrappers (which weren't consistent).

In summary this allows:
 - One Recorder for both upper and lower bounds. Previously would have required two recorders.
 - The constraint thresholds to be defined on the Recorder itself instead of using a separate threshold recorder.
- A single Recorder can be an objective and a bounded constraint. This has generally been useful for bounding objectives (see example below).
- Makes the platypus and pygmo wrappers handle user defined constraints. Currently platypus defaults to `<= 0` and pygmo to `== 0`. This means that running the same model with the two wrappers would result in different treatment of the constraints. That should now be solved.

Issues:
 - It breaks the JSON format by removing `is_constraint` and replacing it with `constraint_upper_bounds` and `constraint_lower_bounds`. 

Constrained objective example that minimise bounded between 2 and 10. Before this would require creating two additional recorders to compute the difference between the objective and constraint bounds.
```json
{
  "recorders": {
    "my_constrained_objective": {
      "type": "...",
      "is_objective": "min",
      "constraint_upper_bounds": 10.0,
      "constraint_lower_bounds": 2.0
    }
  }
}
```

@matrosoe @Batch21 Could you both have a look at this as you have both used the optimisation stuff quite a lot.


jetuk:
One other thought on this. There's no function here to determine whether the constraint bounds are violated; this is handled by the platypus or pygmo at the moment. However, it might be useful to have a function that simply returns whether the individual recorder is violated or the entire model is in violation of these constraints (and/or the magnitude of the violation). What do you think?

Jmiguel17:
This is very useful. However, I have a question, why does it declare `"objective": "min"` in the constraint? is this not confusing? Actually, I gave a look at the implementation and I didn't find any reference where do you use the `"objective"` argument.

jetuk:
@Jmiguel17 this should have been `"is_objective"` (I've edited the description now, thanks for spotting). It's the existing functionality for defining an objective and is unchanged in this PR. The declaration of an objective here is entirely optional, but this change allows you to have one recorder return a value for an objective and also constrain the range of that objective. 

You can still define a constraint on its own with `"is_objective": false` (or omitted), or an objective on its own with no constraint bounds.

jetuk:
> One other thought on this. There's no function here to determine whether the constraint bounds are violated; this is handled by the platypus or pygmo at the moment. However, it might be useful to have a function that simply returns whether the individual recorder is violated or the entire model is in violation of these constraints (and/or the magnitude of the violation). What do you think?

I've now added methods to compute recorder level constraint violation and model level feasibility.

pywr/_model.pyx
```diff
@@ -140,6 +140,13 @@ class Model(object):
     def objectives(self):
         return NamedIterator(n for n in self.recorders if n.is_objective)
 
+    def is_feasible(self):
```
Batch21, line 143: worth calling this in `model.check`?

pywr/optimisation/pygmo.py
```diff
@@ -23,11 +23,29 @@ def fitness(self, solution):
             value = r.aggregated_value()
             objectives.append(sign*value)
 
-        constraints = [r.aggregated_value() for r in self.model_constraints]
+        # Return separate lists for equality and inequality constraints.
+        # pygmo requires that inequality constraints are all of the form g(x) <= 0
+        # Therefore these a converted to this form from their respective bounds.
```
Batch21, line 28: 'There these **are** ...'

pywr/optimisation/platypus.py
```diff
@@ -56,8 +72,25 @@ def _make_variables(self, variables):
...
+                self.problem.constraints[ic] = platypus.Constraint('==', value=c.constraint_lower_bounds)
+                ic += 1
+            elif c.is_lower_bounded_constraint:
+                self.problem.constraints[ic] = platypus.Constraint('>=', value=c.constraint_lower_bounds)
+                ic += 1
+            elif c.is_upper_bounded_constraint:
+                self.problem.constraints[ic] = platypus.Constraint('<=', value=c.constraint_upper_bounds)
+                ic += 1
+            else:
+                raise RuntimeError(f'The bounds if constraint "{c.name}" could not be identified correctly.')
```
Batch21, line 93: '... **of** constraint'

pywr/recorders/_recorders.pyx
```diff
@@ -201,20 +201,30 @@ cdef class Recorder(Component):
...
     is_objective : {None, 'maximize', 'maximise', 'max', 'minimize', 'minimise', 'min'}
         Flag to denote the direction, if any, of optimisation undertaken with this recorder.
-    is_constraint : bool (default=False)
-        Flag to denote whether this recorder is to be used as a constraint during optimisation.
     epsilon : float (default=1.0)
         Epsilon distance used by some optimisation algorithms.
+    constraint_lower_bounds, constraint_upper_bounds : double (default=None)
+        The value(s) to use for lower and upper bound definitions. These values determine whether the recorder
+        instance is marked as a constraint. Either bound can be `None` (the default) to disable the respective
+        bound. If both bounds are `None` then the `is_constraint` property will return `False`. The lower bound must
```
Batch21, line 209: might be worth mentioning here that you set both constraints to same value for an equality constraint

pywr/recorders/_recorders.pyx
```diff
@@ -240,6 +250,78 @@ cdef class Recorder(Component):
...
+    property constraint_upper_bounds:
+        def __set__(self, value):
+            if value is None:
+                self._constraint_upper_bounds = -np.inf
+            else:
+                if self.constraint_lower_bounds is not None and value < self.constraint_lower_bounds:
+                    raise ValueError('Upper bounds can not be smaller than the lower bounds.')
+                self._constraint_upper_bounds = value
+        def __get__(self):
+            if np.isneginf(self._constraint_upper_bounds):
```
Batch21, line 276: should be `np.isinf`

jetuk:
> I noticed that there are no changes for ispyred. I'm not that familiar with this package - does it not use constraints?

It doesn't have constraints built-in. The tutorial suggests punishing the objective function(s) if constraints are violated. However, the current wrapper just ignores them.  We should probably add a warning or raise an error if there are constraints when using the inspyred wrapper.

pywr/_model.pyx
jetuk: The `model.check()` function was intended as sanity check of the model before simulation. You can't do this feasibility check before the simulation.

pywr/recorders/_recorders.pyx
jetuk: Done.

jetuk:
@Batch21 I've addressed your comments.

pywr/recorders/_recorders.pyx
jetuk: Nice catch. The internal representation should have been `np.inf` as well.

pywr/_model.pyx
Batch21: ok

jetuk: I'll add a comment to the docstring.

Batch21:
> We should probably add a warning or raise an error if there are constraints when using the inspyred wrapper.

Is this going to be a separate issue?

jetuk:
> Is this going to be a separate issue?

Yeah, this change makes it no worse :)

jetuk:
See https://github.com/pywr/pywr/issues/871

jetuk:
Thanks @Batch21. Just going to see if @matrosoe has any comments.

jetuk:
Thanks both. I'll merge this.

PR #870, opened 2020-04-22
Add IPython to install dependencies.

jetuk:
This should reactivate the `tests/test_notebook.py` module.


jetuk:
Hmm that didn't work. Something else must going on!

PR #873, opened 2020-04-30
Add RbfProfileParameter.

jetuk:
See docstring of new parameter for information.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -912,7 +912,7 @@ cdef class RbfProfileParameter(Parameter):
     The daily profile is computed during model `reset` using a radial basis function with
     day-of-year as the independent variables. The days of the year are defined by the user
     alongside the values to use on each of those days for the interpolation. The first
-    day of the years should always be one, and its value is repeated as the 365th value.
+    day of the years should always be one, and its value is repeated as the 366th value.
     In addition the second and penultimate values are mirrored to encourage a consistent
     gradient to appear across the boundary. The RBF calculations are undertaken using
     the `scipy.interpolate.Rbf` object, please refer to Scipy's documentation for more
@@ -943,10 +943,23 @@ cdef class RbfProfileParameter(Parameter):
         self.double_size = len(values)
         self.integer_size = 0
         self._values = np.array(values)
-        self._days_of_year = np.array(days_of_year, dtype=np.int32)
+        self.days_of_year = days_of_year
         self._lower_bounds = np.ones(self.double_size)*lower_bounds
         self._upper_bounds = np.ones(self.double_size)*upper_bounds
         self.rbf_kwargs = rbf_kwargs if rbf_kwargs is not None else {}
+
+    property days_of_year:
+        def __get__(self):
+            return np.array(self._days_of_year)
+        def __set__(self, values):
+            values = np.array(values, dtype=np.int32)
+            if values[0] != 1:
+                raise ValueError('The first day of the years must be 1.')
+            if len(values) < 3:
+                raise ValueError('At least days of the year are required.')
+            if np.any(0 > values > 365):
+                raise ValueError('Days of the years should be between 1 and 365 inclusive.')
+            self._days_of_year = values
 
     cpdef reset(self):
         Parameter.reset(self)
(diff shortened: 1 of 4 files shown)


jetuk:
Note that this is a more specific form of what's in #725 .

pywr/parameters/_parameters.pyx
```diff
@@ -905,6 +906,109 @@ cdef class UniformDrawdownProfileParameter(Parameter):
...
+        self._interpolate()
+
+    cpdef _interpolate(self):
+        cdef int i
+
+        days_of_year = list(self._days_of_year)
+        # Append day 365 to the list and mirror the penultimate and second DOY at the start and end
+        # of the list respectively. This helps ensure the gradient is roughly the same across the boundary
+        # between days 365 and 0.
+        days_of_year = [days_of_year[-1]-365+1] + list(days_of_year) + [365, 365+days_of_year[1]-1]
```
Batch21, line 965: If one of the input 'days_of_year' is 365 then adding another one here will cause an error. probably needs a check/warning in the docstring/init

pywr/parameters/_parameters.pyx
```diff
@@ -905,6 +906,109 @@ cdef class UniformDrawdownProfileParameter(Parameter):
...
+    In addition the second and penultimate values are mirrored to encourage a consistent
+    gradient to appear across the boundary. The RBF calculations are undertaken using
+    the `scipy.interpolate.Rbf` object, please refer to Scipy's documentation for more
+    information.
+
+    Parameters
+    ----------
+    days_of_year : iterable, integer
+        The days of the year at which the interpolation values are defined. The first
+        value should be one.
```
Batch21, line 925: Worth raising an exception if the first value is not 1?

pywr/parameters/_parameters.pyx
```diff
@@ -905,6 +906,109 @@ cdef class UniformDrawdownProfileParameter(Parameter):
...
+    """
+    def __init__(self, model, days_of_year, values, lower_bounds=0.0, upper_bounds=np.inf, rbf_kwargs=None, **kwargs):
+        super(RbfProfileParameter, self).__init__(model, **kwargs)
+
+        if len(days_of_year) != len(values):
+            raise ValueError(f"The length of values ({len(values)}) must equal the length of "
+                             f"`days_of_year` ({len(days_of_year)}).")
+
+        self.double_size = len(values)
+        self.integer_size = 0
```
Batch21, line 944: Already set to 0 in `Parameter`?

pywr/parameters/_parameters.pyx
jetuk: I think this is incorrect actually. It should be repeating day 1 as value 366 - I'm assuming zero based indexing here.

pywr/parameters/_parameters.pyx
jetuk: Good idea.

pywr/parameters/_parameters.pyx
jetuk: Just being explicit :)

pywr/parameters/_parameters.pyx
Batch21: yeah ok, forgot about the zero indexing!

jetuk: This should be fixed now. I've added some checking of the days of the year.

pywr/parameters/_parameters.pyx
jetuk: Done.

jetuk: Actually, I guess this could be implemented without this being a requirement?

jetuk:
@Batch21 I've updated this. The interpolation 1-based indexing should make sense, and there are input checks on the `days_of_year`. However, I just thought whether the requirement for the first day to be 1 is really needed?

Batch21:
because it assumes the users wants the boundary to be as smooth as possible? Could just be advice in the docstring rather than a hard constraint

jetuk:
I think you could retain the mirroring (for the boundary effect) without requiring the first day to be 1.

jetuk:
It does make it much more controlled at the boundary, so perhaps a good idea to keep it as is.

Batch21:
I'm happy for it to be kept. Had a quick play around, and even with the mirroring you don't have to move the index of the first value that far away from 1 to get quite a jump at the boundary

jetuk:
OK. So shall we merge this as is?

Issue #632, opened 2018-06-14
You shouldn't be able to set is_variable = True if it's not supported.


jetuk:
Taken from #230 

 >>   Ideally you shouldn't be able to set is_variable = True if it's not supported.
>
> I think that can be achieved by making is_variable a property on Parameter. We'd need a class variable that declares if Parameter can be a variable, which causes the __set__ method to raise an exception if needed.
>
> At first I thought that sounded like two variables doing the same thing, but one would describe if a Parameter can be a variable and the other would describe if it actually is.

An alternative is to have the property check `double_size + integer_size > 0`.

jetuk:
The problem with the second approach is that the size could be dynamic and is generally not initialised until `__init__`. Therefore, the check would fail in `super()` before size is initialised if `is_variable` is passed as a keyword argument.

PR #874, opened 2020-04-30
Add OffsetParameter

jetuk:
I also fixed a bug that meant the `NegativeMinParameter` was not being tested and the `NegativeMaxParameter` was not being tested properly.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -1578,7 +1578,7 @@ NegativeMinParameter.register()
 
 
 cdef class OffsetParameter(Parameter):
-    """Parameter that takes offsets another `Parameter` by a constant value.
+    """Parameter that offsets another `Parameter` by a constant value.
 
     This class is a more efficient version of `AggregatedParameter` where
     a single `Parameter` is offset by a constant value.
@@ -1586,9 +1586,9 @@ cdef class OffsetParameter(Parameter):
     Parameters
     ----------
     parameter : `Parameter`
-        The parameter to to compare with the float.
+        The parameter to compare with the float.
     offset : float (default=0.0)
-        The offset to add to apply to value returned by `parameter`.
+        The offset to apply to the value returned by `parameter`.
     lower_bounds : float (default=0.0)
         The lower bounds of the offset when used during optimisation.
     upper_bounds : float (default=np.inf)
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -1577,6 +1577,58 @@ cdef class NegativeMinParameter(MinParameter):
 NegativeMinParameter.register()
 
 
+cdef class OffsetParameter(Parameter):
+    """Parameter that takes offsets another `Parameter` by a constant value.
+
+    This class is a more efficient version of `AggregatedParameter` where
+    a single `Parameter` is offset by a constant value.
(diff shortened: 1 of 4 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -1577,6 +1577,58 @@ cdef class NegativeMinParameter(MinParameter):
 NegativeMinParameter.register()
 
 
+cdef class OffsetParameter(Parameter):
+    """Parameter that takes offsets another `Parameter` by a constant value.
```
Batch21, line 1581: Delete 'takes'

pywr/parameters/_parameters.pyx
```diff
@@ -1577,6 +1577,58 @@ cdef class NegativeMinParameter(MinParameter):
...
+
+    This class is a more efficient version of `AggregatedParameter` where
+    a single `Parameter` is offset by a constant value.
+
+    Parameters
+    ----------
+    parameter : `Parameter`
+        The parameter to to compare with the float.
```
Batch21, line 1589: delete extra 'to'

pywr/parameters/_parameters.pyx
```diff
@@ -1577,6 +1577,58 @@ cdef class NegativeMinParameter(MinParameter):
...
+    offset : float (default=0.0)
+        The offset to add to apply to value returned by `parameter`.
```
Batch21, line 1591: delete 'to add'

pywr/parameters/_parameters.pyx
```diff
@@ -1577,6 +1577,58 @@ cdef class NegativeMinParameter(MinParameter):
...
+        The upper bounds of the offset when used during optimisation.
+    """
+    def __init__(self, model, parameter, offset=0.0, lower_bounds=0.0, upper_bounds=np.inf, *args, **kwargs):
+        super(OffsetParameter, self).__init__(model, *args, **kwargs)
+        self.parameter = parameter
+        self.children.add(parameter)
+        self.offset = offset
+        self.double_size = 1
+        self._lower_bounds = np.ones(self.double_size) * lower_bounds
+        self._upper_bounds = np.ones(self.double_size) * upper_bounds
```
Batch21, line 1604: There are a few other parameters that need need to have these optimisation attributes setup in the same way (e.g. `MaxParameter`).

Could the parameters that have another parameter and a single float as input args all inherit from a single base class that defines these attributes and associated methods?

jetuk: I'm not sure how you would do this cleanly (e.g. this offset uses a different internal name). I think there could be a common base class for all of `MaxParameter`, `MinParameter`, `NegativeMaxParameter` and `NegativeMinParameter` to add this support. Probably one for another issue. It is also related to #632. What do you think?

jetuk: We should probably have a `MinMaxParameter` too.

jetuk:
@Batch21 I've fixed the docstring typos (please check) and replied to the other comment. Thanks.

pywr/parameters/_parameters.pyx
Batch21: I agree it is a separate issue related to #632. Looks like there are quite a few parameters that people potentially might want to use as optimisation variables but are not currently setup with the new attributes.

PR #877, opened 2020-05-02
Expose run statistics to optimisation wrappers

jetuk:
Log the run time and speed during optimisation, and also save the `ModelResult` on to the wrapper so it can be inspected if required.

### DIFF
--- a/pywr/optimisation/platypus.py
+++ b/pywr/optimisation/platypus.py
@@ -106,7 +106,7 @@ class PlatypusWrapper(BaseOptimisationWrapper):
                 ints = np.round(np.array(x[-var.integer_size:])).astype(np.int32)
                 var.set_integer_variables(ints)
 
-        run_stats = self.model.run()
+        self.run_stats = self.model.run()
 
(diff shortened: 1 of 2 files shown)


jetuk:
The failure here is just appveyor being a bit terrible.

pywr/optimisation/platypus.py
```diff
@@ -106,7 +106,7 @@ def evaluate(self, solution):
                 ints = np.round(np.array(x[-var.integer_size:])).astype(np.int32)
                 var.set_integer_variables(ints)
 
-        run_stats = self.model.run()
+        self.run_stats = self.model.run()
```
Batch21, line 109: Won't this attribute get continuously overwritten during the optimisation?

jetuk: Yes, but this allows you to access the run statistics if you subclass the wrapper. E.g. if you want to save them somewhere.

Batch21: ah ok

PR #879, opened 2020-05-04
Change the signature of recorder aggregation functions to catch exceptions.

jetuk:
Update the Cython function signature to `except *` so that exception checking is always done. This is helpful if there are errors during the computation of metrics during optimisation.

### DIFF
--- a/pywr/recorders/_recorders.pxd
+++ b/pywr/recorders/_recorders.pxd
@@ -17,8 +17,8 @@ cdef class Recorder(Component):
     cdef public double epsilon
     cdef public bint ignore_nan
     cdef public Aggregator _scenario_aggregator
-    cpdef double aggregated_value(self) except? -1
-    cpdef double[:] values(self)
+    cpdef double aggregated_value(self) except *
+    cpdef double[:] values(self) except *
 
 cdef class AggregatedRecorder(Recorder):
     cdef object recorder_agg_func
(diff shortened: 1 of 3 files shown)


Batch21:
Not sure I fully understand the Cython documentation for exceptions. For `except *` it states:

> If you have a function returning void that needs to propagate errors, you will have to use this form, since there isn’t any return value to test. Otherwise there is little use for this form

But the functions where you've added it are returning values. I think you said this change is because `except? -1` wasn't working correctly?

jetuk:
The `except? -1` form means that if the cdef function returns `-1` Cython will check if an exception occurred. It will then propagate that exception in Python space. The `except *` form means a check for an exception is done after every call to the function. The docs are saying if your function returns no value then you have to use the `except *` form because there is no return value to trigger the check.

Currently this means if an exception occurs inside, for example, `.aggregated_value()` then unless that function also returns -1 (which they don't) the exception is lost and program continues. Changing to `except *` will mean any exception in that function will be propagated to Python space. Basically for a slightly loss in performance we make debugging errors in these functions much easier (and also reduce the risk of simulations appearing to run fine but are hiding errors).

jetuk:
Note we could make this change to other functions. For example the `Parameter.value()` method. However, this would need testing for its performance impact. I'm not actually sure what happens if you run a parameter like this:

```python
class ExceptionParameter(Parameter):
    def value(self, ts, si):
        raise ValueError('A bad thing happened!')
```

And would it be different if this was Cythonised parameter?

Batch21:
thanks for the info. It seems like the change is worth the slight performance hit for the aggregated functions.

For other functions, such as `Parameter.value()`, it would be nice if you could run the model in a debug mode that would switch on the better error handling. Can't really see a nice way of implementing this thought.

jetuk:
It might be worth benchmarking the impact on `.value`. That would bea separate issue I guess. Shall we merge this as is?

PR #882, opened 2020-05-08
Add optional dependencies for docs and dev.

### LINKED ISSUE #755: Add documentation dependencies to setup.py
There are a few dependencies required to build the documentation that aren't mentioned in `setup.py`. These should be added as "extras". I suggest a "docs" and "dev" (includes docs) extras.

* sphinx
* sphinx_rtd_theme
* numpydoc

jetuk:
There was an existing test extra which I renamed to dev.

Fixes #755.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,9 @@ def setup_package():
 
     # Extra optional dependencies
     docs_extras = ["sphinx", "sphinx_rtd_theme", "numpydoc"]
-    dev_extras = docs_extras + ["pytest"]
+    test_extras = ["pytest"]
+    dev_extras = docs_extras + test_extras
+    opt_extras = ["platypus-opt", "pygmo"]
 
     metadata = dict(
         name="pywr",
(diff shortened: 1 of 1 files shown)


jetuk:
The failure here is appveyor being rubbish (again).

@Batch21 could you review?

setup.py
```diff
@@ -44,9 +48,12 @@ def finalize_options(self):
         author="Joshua Arnott",
         author_email="josh@snorfalorpagus.net",
         url="https://github.com/pywr/pywr",
-        setup_requires=["setuptools>=18.0", "setuptools_scm", "cython", "numpy",],
+        setup_requires=["setuptools>=18.0", "setuptools_scm", "cython", "numpy"],
         install_requires=["pandas", "networkx", "scipy", "tables", "xlrd", "packaging", "matplotlib", "jinja2", "ipython"],
-        extras_require={"test": ["pytest"]},
```
Batch21, line 49: setuptools docs says this:

> if a project ends up not needing any other packages to support a feature, it should keep an empty requirements list for that feature in its extras_require argument, so that packages depending on that feature don’t break (due to an invalid feature name

But maybe not necessary here? I'm not aware of any other packages that depend on pywr

jetuk: Because I've renamed it? I was in two minds about this. We could still call it "test" if you prefer.

jetuk: We should probably have one like this too: `"optimisation": ["platypus-opt", "pygmo"]`

Batch21: On the keys, I think its just best practice to retain keys so that it doesn't break any downstream packages. Probably also clearer as `test`

I agree with having an optimisation entry.

jetuk: I've kept "dev", put "test" back and added "optimisation".

PR #885, opened 2020-05-09
skip virtual nodes when creating d3 graph data from model json

Batch21:
This makes it consistent with how the d3 data is created from a model object

### DIFF
--- a/tests/test_notebook.py
+++ b/tests/test_notebook.py
@@ -46,8 +46,10 @@ def test_from_json(from_json):
 
 
 def test_d3_data():
-    """ Check that the graph data returned by the pywr_json_to_d3_json and pywr_model_to_d3_json functions contains the
-    same nodes and number on links. The data won't match exactly due to differences in node ordering.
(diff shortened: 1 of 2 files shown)


jetuk:
Should we have a test like this?

```python
assert pywr_model_to_d3_json(Model.load(data)) == pywr_json_to_d3_json(data)
```

Batch21:
Couldn't add a test that does exactly that, as the resulting node ordering is different for each method, but I've added a test that does some checks that the data is the same from a graphing point of view.

tests/test_notebook.py
```diff
@@ -43,3 +43,18 @@ def test_from_json(from_json):
     demand_max_flow = get_node_attribute(demand, "max_flow")
 
     assert demand_max_flow["value"] == "demand_max_flow - AggregatedParameter"
+
+
+def test_d3_data():
+
```
jetuk, line 49: Just add a quick description for this. Comment that it only checks nodes are the same and number of links are the same.

Issue #2, opened 2015-01-21
Implement seasonal and rolling licences


snorfalorpagus:
Implement a licence class to be handled by the solver that supports:
- [x] Daily licence
- [x] Annual licence
- [x] Seasonal licence
- [x] Rolling licence
- [x] Licence groups (done via aggregated nodes)
- [x] JSON support

snorfalorpagus:
Daily and annual licenses implemented in 5047294a67845992e9674490605b752c28ecb139. Solver applies licenses as constraints in 5029478d999b8b50e21269f7a7de98bbaca195ce. Still need to implement other license types, including:

* Group licenses

snorfalorpagus:
Remove timestep specific code from DailyLicense. TimestepLicense?

snorfalorpagus:
I've begun to make the changes required for licenses to work with non-daily timesteps, and for the use of seasonal and rolling licenses in ddaca50.

However, to continue a decision needs to be made: should licenses have access to `model.timestamp` (currently they do not) or should `License.commit` be called with `timestamp` as an argument?

jetuk:
I vote for something on commit. This way in theory it could vary. However
does it need timestamp and timestep?

I've begun to make the changes required for licenses to work with non-daily
timesteps, and for the use of seasonal and rolling licenses in ddaca50
<https://github.com/snorfalorpagus/pywr/commit/ddaca50377427431a353a810f08c906850a3d0ad>
.

However, to continue a decision needs to be made: should licenses have
access to model.timestamp (currently they do not) or should License.commit
be called with timestamp as an argument?

—
Reply to this email directly or view it on GitHub
<https://github.com/snorfalorpagus/pywr/issues/2#issuecomment-112209932>.

snorfalorpagus:
Are you thinking of a variable timestep? I hadn't really considered that.

jetuk:
It's common when solving differential equations in time. They often have
some sort of error check to reduce the timestep if it thinks the one it has
just done is too large. That's probably not needed here, but you could
imagine specifying weekly or daily during different periods of the year, or
if storage was low?

Perhaps not the highest priority.

snorfalorpagus:
Currently licences that have a state (e.g. annual licences) are not scenario-aware.

snorfalorpagus:
As of 491f764bbe3e4a4b1692c4d1d545596eba040abb the licence classes have been refactored. They are now scenario-aware.

XML support is now broken (I think).

[...]

PR #891, opened 2020-05-22
Initial commit of RollingVirtualStorage node.

jetuk:
This is a basic implementation of a subclass of `VirtualStorage` that keeps a memory of past flow that is re-applied to the store after a number days/timesteps.

The tests need improving to ensure it works with more complicated use cases. Comments welcome.

Partially resolves issue #2.


pywr/_core.pyx
```diff
@@ -962,18 +962,23 @@ cdef class Storage(AbstractStorage):
...
-            # Ensure variable maximum volume is taken in to account
+            # Apply any storage adjustment if given
+            if adjustment is not None:
+                self._volume[i] += adjustment[i]
 
+            # Ensure variable maximum volume is taken in to account
             mxv = self.get_max_volume(si)
             mnv = self.get_min_volume(si)
+            # Ensure volume stays within bounds.
+            self._volume[i] = min(mxv, max(mnv, self._volume[i]))
```
jetuk, line 981: Note to self. This is probably a bad idea from as mass balance point of view. At least make it optional / configurable.

pywr/_core.pyx
```diff
@@ -1080,4 +1085,38 @@ cdef class VirtualStorage(Storage):
...
+        assert adjustment is None
+
+        # Update the storage volumes by applying an adjustment
+        VirtualStorage.after(self, ts, adjustment=self._memory[self._memory_pointer, :])
+
+        # Store today's flow in the memory and increment the memory pointer
+        for i, si in enumerate(self.model.scenarios.combinations):
+            # Flow is negative in VirtualStorage to remove from the store, save the +ve number here for
+            # returning to the store later
+            self._memory[self._memory_pointer, i] = -self._flow[i]
```
Batch21, line 1121: Think this needs to be: 

`self._memory[self._memory_pointer, i] = -self._flow[i] * ts.day`

Otherwise the node does not work correctly if the model timestepper  delta > 1.  Probably worth having a test to check this

pywr/nodes.py
```diff
@@ -477,6 +477,92 @@ def load(cls, data, model):
...
+        The maximum volume of the storage.
+    initial_volume: float
+        The initial storage volume.
+    timesteps : int
+        The number of timesteps to apply to the rolling storage over.
+    days : int
+        The number of days to apply the rolling storage over. Specifying a number of days (instead of a number
+        of timesteps) is only valid with models running a timestep of daily frequency.
+    cost: float or parameter
+        The cost of flow into/outfrom the storage.
```
Batch21, line 506: Add space between 'out' and 'from'

pywr/nodes.py
```diff
@@ -477,6 +477,92 @@ def load(cls, data, model):
...
+
+        if factors is None:
+            self.factors = [1.0 for i in range(len(nodes))]
+        else:
+            self.factors = factors
+
+    def check(self):
+        super().check()
+        if self.cost not in (0.0, None):
+            raise NotImplementedError("VirtualStorage does not currently support a non-zero cost.")
```
Batch21, line 544: RollingVirtualStorage not VirtualStorage

pywr/nodes.py
```diff
@@ -477,6 +477,92 @@ def load(cls, data, model):
...
+
+    def setup(self, model):
+        if self.days is not None and self.days > 0:
+            try:
+                self.timesteps = self.days // self.model.timestepper.delta
+            except TypeError:
+                raise TypeError('A rolling period defined as a number of days is only valid with daily time-steps.')
+        if self.timesteps < 1:
+            raise ValueError('The number of time-steps for a RollingVirtualStorage node must be greater than one.')
```
Batch21, line 553: This check is also done in the setup of the cython class

Batch21 (changes requested):
I've test this on a couple of relatively complex models and it seems to work well expect in the case when the model timestep is greater than daily

pywr/_core.pyx
jetuk: @Batch21 did we discuss / resolve my previous concerns about this line?

Batch21: yeah, I think 40aefe00589d2bde48bf3b1e797f67be0e483370 was the result of that discussion

jetuk:
@Batch21 I've addressed you comments and also merged master back in here. This should now pass all the tests.

pywr/nodes.py
jetuk: I think that's fine.

Batch21 (approved):
changes look good

PR #892, opened 2020-05-28
Add ability to provide other solutions in PywrRandomGenerator.

### LINKED ISSUE #887: Injected Population to Platypus
Hi @jetuk 

I was looking at the `PywrRandomGenerator` class to see a way to inject a complete set of non-dominated solutions into `Platypus` to continue an algorithm search from the last non-dominated solutions I got from a previous search. However, I noticed that using this `pywr` class I can inject only one non-dominated solution (the solution there is in the Json file). I would like to ask you two questions:

1) Do you know if it is ok just inject one solution to continue the search? or I would need to inject a complete set of non-dominated solutions equal to the pop-size?
2) If you recommend me to inject the complete set of non-dominated solutions. Could you share some ideas about how this could be efficiently implemented in `pywr`? I could dedicate time to do this implementation but it would be helpful to get some tips from you.

Thanks

https://github.com/pywr/pywr/blob/e7b4e35c5eef08442e1f390c4a5b8270cb0ae926/pywr/optimisation/platypus.py#L135

jetuk:
This extends the ability of PywrRandomGenerator to return alternative initial solutions instead of just the current model configuration. User's provide a list of dictionaries containing variable values for each of the variable Parameters.

Fixes #887.

### DIFF
--- a/pywr/optimisation/platypus.py
+++ b/pywr/optimisation/platypus.py
@@ -135,33 +135,62 @@ class PlatypusWrapper(BaseOptimisationWrapper):
 class PywrRandomGenerator(platypus.RandomGenerator):
     """A Platypus Generator that injects the current setup of the Pywr model into the population.
 
-    The first Solution returned from the generate method is taken from the wrapper (i.e. the Pywr
-    model being wrapped) as the current values of the variable Parameters. This allows the population
-    to be seeded with the current model configuration, which is often a initial solution.
+    When use_current is true the first Solution returned from the generate method is taken from the wrapper
+    (i.e. the Pywr model being wrapped) as the current values of the variable Parameters. This allows the population
+    to be seeded with the current model configuration, which is often an initial solution. Additional solutions
+    can be provided in as an iterable of solutions. These can come from an alternative source such as previous
+    optimisation.
 
     Parameters
     ==========
     wrapper : PlatypusWrapper
         Wrapper from which to grab the current model and decision variables.
+    use_current: Bool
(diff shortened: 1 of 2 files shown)


pywr/optimisation/platypus.py
```diff
@@ -135,33 +135,62 @@ def evaluate(self, solution):
 class PywrRandomGenerator(platypus.RandomGenerator):
     """A Platypus Generator that injects the current setup of the Pywr model into the population.
```
Batch21, line 136: Need to update this line as you now use the class without injecting the current setup

tests/test_platypus.py
```diff
@@ -80,11 +80,38 @@ def test_platypus_nsgaii_process_pool(two_reservoir_problem):
...
+    def test_other_solutions(self, two_reservoir_problem, use_current):
+        """Test PywrRandomGenerator inserts other solutions into the population. """
+
+        # Create some alternative initial solutions
+        solutions = [
+            {
+                'control_curve': {'doubles': [1] * 12}
+            },
+            {
+                'control_curve': {'doubles': [2] * 12}
```
Batch21, line 103: Might be worth extending this so it tests setting integer variables

jetuk: I just looked at doing this. We currently have no parameters with `integer_size > 0`. So we'll have to park this.

pywr/optimisation/platypus.py
jetuk: Done.

PR #893, opened 2020-06-04
Reset RollingMeanFlowNodeRecorder's memory position on Model.reset()

jetuk:
I noticed this while looking at something else. Small fix to make sure the internal memory is reset if setup is not recalled between runs.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -1438,10 +1438,10 @@ cdef class RollingMeanFlowNodeRecorder(NodeRecorder):
         else:
             self.days = 0
         self._data = None
+        self.position = 0
 
     cpdef setup(self):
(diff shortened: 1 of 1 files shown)


pywr/recorders/_recorders.pyx
```diff
@@ -1438,10 +1438,10 @@ cdef class RollingMeanFlowNodeRecorder(NodeRecorder):
         else:
             self.days = 0
         self._data = None
+        self.position = 0
 
     cpdef setup(self):
         super(RollingMeanFlowNodeRecorder, self).setup()
-        self.position = 0
```
Batch21, line 1444: `RollingWindowParameterRecorder` initialises `position` in `setup`. I don't thinks this makes any real difference but it might be good to have them consistent?

jetuk: It does it in `reset` too?

PR #895, opened 2020-06-07
skip AggregatedNode and AggregatedStorage when creating d3 data

Batch21:
This makes `pywr_json_to_d3_json` and `pywr_model_to_d3_json` consistent in how they deal with virtual nodes.

It also fixes an error that occurred if you tried to add attributes to the graph using `pywr_json_to_d3_json` when the model contained AggregatedNode and/or AggregatedStorage nodes

### DIFF
--- a/pywr/notebook/__init__.py
+++ b/pywr/notebook/__init__.py
@@ -287,10 +287,10 @@ def pywr_json_to_d3_json(model, attributes=False):
 
     nodes = []
     node_classes = create_node_class_trees()
-
     for node in model["nodes"]:
 
-        if node["type"].lower() in ["annualvirtualstorage", "virtualstorage"]:
+        if node["type"].lower() in ["annualvirtualstorage", "virtualstorage", "aggregatednode", "aggregatedstorage"]:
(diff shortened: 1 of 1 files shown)


pywr/notebook/__init__.py
```diff
@@ -287,10 +287,10 @@ def pywr_json_to_d3_json(model, attributes=False):
 
     nodes = []
     node_classes = create_node_class_trees()
-
     for node in model["nodes"]:
 
-        if node["type"].lower() in ["annualvirtualstorage", "virtualstorage"]:
+        if node["type"].lower() in ["annualvirtualstorage", "virtualstorage", "aggregatednode", "aggregatedstorage"]:
```
jetuk, line 292: Should we be looking up the class here and testing for `allow_isolated`?

Batch21: I had a look at this but I don't think you can test the value of `allow_isolated` or `virtual` without creating an instance of the node. In this case, the function could just load the full model and call `pywr_model_to_d3_json`

Issue #791, opened 2019-10-29
Adding discount rate support to temporal aggregation.


jetuk:
@matrosoe and I discussed how we could add discount rates to the recorder aggregation functions today. It would be good to have support for the following:

1. A generic approach in the `Aggregator` class so that one can specify discounted aggregation as an alternative `agg_func` in any recorder that supports it.
2. The ability to supply time vary discount rates. 
3. Per recorder discount rates.
4. Specify the base year of discounting (might no coincide with the model's start year).

The main challenge in implementing this is that the existing `Aggregator` class does not have access to the model or the time-steps associated with the data it is given. When performing a discounted aggregation this would be a function of the years in the simulation. Therefore the aggregator needs to be aware of the time-steps associated with the data arrays. Two approaches to this:

1. We pass the time data (`pandas.DatetimeIndex`) to the aggregation functions. Those functions can use (or not) that datetime index during aggregation. This has the downside that every call to the aggregation functions would need to be modified to pass this new argument. We would also have to make sure that the index object is not recreated for every recorder. 
2. We give the aggregator a reference to the model instance which it can use (or not) to get the datetime index if it needs to. This makes the `Aggregator` class more complex, but is potential more efficient by only getting the datetime index when necessary. 

Finally, we noted,

1. That discounted aggregation could just be a special case of weighted sum (or dotproduct). The weights could be given by the users, or calculated from the discount rates.
2. If the `Aggregator` has access to the datetime index then we can conceivably make some of the pandas resampling aggregation functions available.

@Batch21 any thoughts on this?

jetuk:
Tagging #182

KevisPachos:
Hi @jetuk , I am using your scheduled formulation (attached as a doc file since json is not supported) with three options (reservoir, desalination and dm) to further test the discount factor developed in #901. I noticed that the way options are currently discounted is not exactly how I want it. In the attached xls file, I am comparing the way I would like to discount annualised capital costs and the way it is currently being done. The difference lies in that currently, an averaged discount rate is applied to an averaged annualised cost. Instead I would like to apply the discount factor in each year based on the annualised cost of that year (reflecting if the option is active from the beginning of the year, or could be half way through, or not at all). Would that be possible with the current discount rate implementation or do I need to add support? This does not require time vary discount rates, the rate can stay fixed at, for instance, 3,5% for all years. It is a matter of calculating the total discounted costs of an option by summing the discounted cost in each year that the cost occurs instead of using average values. Thanks

[...]

PR #901, opened 2020-06-18
Adding discounting capability 

KevisPachos:
This functionality is added to address issue 791 (https://github.com/pywr/pywr/issues/791).

A new parameter "DiscountFactorParameter" returns the current discount factor based on the discount rate (which is assumed to be fixed) and a base year. We then product this with other parameters to calculate discounted operational and capital costs.


tests/test_discount.py
```diff
@@ -0,0 +1,31 @@
+from pywr.parameters import Parameter, load_parameter
+import datetime
+
+class DiscountFactorParameter(Parameter):
```
jetuk, line 4: This parameter looks good, but it should be placed in `pywr/parameters/parameters.py` so its included in the library. Placed here it's in a test module. Instead this module should contain a test function that loads the `discount.json` model you added, and then test that the parameter loads correctly. Another test could test that the outputs of the parameter are the expected values. See `tests/test_parameters.py` for how this is done for some other parameters.

tests/test_discount.py
```diff
@@ -0,0 +1,31 @@
...
+    """Parameter that returns the current discount factor based on discount rate and a base year.
+
+    Parameters
+    ----------
+    discount_rate : discount rate fixed across planning period
+    base_year : base year in 'yyyy' format
```
jetuk, line 10: I think your code wants this as an `int`.
```
base_year : int
    Discounting base year (i.e. the year with a discount factor equal to 1.0).
```

tests/test_discount.py
```diff
@@ -0,0 +1,31 @@
+from pywr.parameters import Parameter, load_parameter
+import datetime
+
+class DiscountFactorParameter(Parameter):
+    """Parameter that returns the current discount factor based on discount rate and a base year.
+
+    Parameters
+    ----------
+    discount_rate : discount rate fixed across planning period
```
jetuk, line 9: Doc should state whether this is a % (0 - 100) or factor (0 - 1). The code is using the latter which is fine.

tests/test_discount.py
```diff
@@ -0,0 +1,31 @@
...
+        self.base_year = base_year
+
+    def value(self, discount_rate, base_year):
+        ct = self.model.timestepper.current #[0:4]
+        cy = ct.year
+        cyi = cy - self.base_year
+        return 1/pow((1+self.discount_rate), cyi) # return discount factor of current year
+
+    @classmethod
+    def load(cls, model, data):
```
jetuk, line 26: This could be simplified:

```python
    @classmethod
    def load(cls, model, data):
        return cls(model, **data)
```

The current code doesn't support  these attributes being parameters. The above will mean you can provide an non-parameter constant values only. If you want to support discount rate being a parameter then it needs a bit more treatment in `__init__` and `value`. I would say start with this, and add support for this later.

jetuk:
@KevisPachos thanks for this. I've left some comments on here. The parameter is largely OK, but needs to be moved inside to the `pywr.parameters.parameters` module, and ideally a test writing that loads, runs and checks your test model does what you think it does.

One other thing to think about is how you might want to use this parameter in conjunction with another parameters / recorders.

tests/test_discount.py
KevisPachos: Thanks James, I have moved the discount parameter to `pywr/parameters/parameters.py` and created a test in `tests/test_parameters.py` to load the `discount.json` model. I have not added yet a test that checks that the outputs of the parameter are the expected values. Should I create a list of expected values for the discount factor in each year based on a 3.5% discount factor and check that the outputs have the same values?

tests/test_discount.py
KevisPachos: Thanks, I have updated the description.

tests/test_discount.py
KevisPachos: Thanks, I added that to the doc.

tests/test_discount.py
KevisPachos: Thanks James, I am not sure the changes I made allows the discount rate being a parameter. I added `**kwargs` to  `__init__` and `value` and `**data` to the `return` function. 
Did you suggest that I removed `discount_rate` and  `base_year` from the `return` function and just use `**data`? If so, I'm not sure how to do this but I could have a go.

pywr/parameters/parameters.py
```diff
@@ -320,3 +320,39 @@ def __setitem__(self, key, value):
             value = ConstantParameter(value)
         dict.__setitem__(self, key, value)
 
+
+class DiscountFactorParameter(Parameter):
+    """Parameter that returns the current discount factor based on discount rate and a base year.
+
+    Parameters
+    ----------
+    discount_rate : 0 - 1 factor
```
jetuk, line 329: This should say `discount_rate : float`

pywr/parameters/parameters.py
```diff
@@ -320,3 +320,39 @@ def __setitem__(self, key, value):
...
+
+    @classmethod
+    def load(cls, model, data):
+        discount_rate = load_parameter(model, data.pop("value"))
+        base_year = load_parameter(model, data.pop("base_year"))
+        return cls(model, discount_rate, base_year, **data)
+
+DiscountFactorParameter.register()
+
+
```
jetuk, line 355: Leave only one blank line at the end of the file.

tests/models/discount.json
```diff
@@ -0,0 +1,63 @@
+{
+    "metadata": {
+        "title": "tutorial",
```
jetuk, line 3: Change the title to `DiscountFactorParameter test model`

tests/models/discount.json
```diff
@@ -0,0 +1,63 @@
...
+        ["supply", "demand"]
+    ],
+    "parameters": {
+        "discount_factor": {
+            "type": "DiscountFactorParameter",
+            "base_year": 2015,
+            "value": 0.035
+        }
+    },
+    "recorders": {
```
jetuk, line 35: Test model doesn't really need these recorders.

tests/test_parameters.py
```diff
@@ -1786,3 +1786,12 @@ def test_variable_api(self, simple_linear_model):
...
         p.set_double_variables(new_values)
         np.testing.assert_allclose(p.get_double_variables(), new_values)
+
+class TestDiscountFactorParameter:
+    def test_discount_json(self):
+        """ Test loading a DiscountFactorParameter from JSON. """
+        model = load_model("discount.json")
+        # run model for period 2015-2020, with base year 2015 and discount rate of 0.035 (3.5%)
+        model.run()
+        # To add: test that the outputs of the parameter are the expected values.
```
jetuk, line 1796: This needs to be done. You can look at the other tests for one that that uses the `@assert_rec` decorator. This allows you to write a simple function to return the value you expect the parameter to return.

Something like this:

```python
class TestDiscountFactorParameter:
    def test_discount_json(self):
        """ Test loading a DiscountFactorParameter from JSON. """
        model = load_model("discount.json")
        # run model for period 2015-2020, with base year 2015 and discount rate of 0.035 (3.5%)
        p = model.parameters['discount_factor']
        @assert_rec(model, p)
        def expected_func(timestep, scenario_index):
            year = timestep.year
            return 1/pow(1.035, year - 2015) 

        model.run()
```

KevisPachos: Thank you James, I have implemented this test now. How can I run the tests to make sure the test works as intended? I tried running `py.test tests` but I don't think that worked. I tried using wrong values so that it would fail the test but I did not get an error.

jetuk: `pytest tests` is the correct command.

jetuk: The test works for me. You can do `pytest tests/test_parameters.py -vv` if you want it to print the name of each test and whether it passed or failed. Or you can do `pytest tests -k discount` to filter for tests containing the word `discount`. 

And you can force a test failure with `assert False`

pywr/parameters/parameters.py
```diff
@@ -320,3 +320,34 @@ def __setitem__(self, key, value):
...
 
+
+class DiscountFactorParameter(Parameter):
+    """Parameter that returns the current discount factor based on discount rate and a base year.
+
+    Parameters
+    ----------
+    discount_rate : float
+                Discount rate (expressed as 0 - 1) used calculate discount factor for each year.
+        base_year : int
```
jetuk, line 331: There's an indenting error here.

jetuk:
This is working now. Does it do what you want? I.e. have you used it in conjunction with other stuff yet? If you want to you could now Cythonise it by moving to to `_parameters.pyx` and adding some `cdef` statements. This will make it much faster if you intend to use it a lot.

tests/test_parameters.py
KevisPachos: Thanks James. So did you get a PASSED for the discount test? When I run `pytest tests/test_parameters.py -vv` I got FAILED for the `TestDiscountFactorParameter::test_discount_json`.

Here is the full message (also attached as screenshot). Since we have not cythonised the parameter yet, why do we get a KeyError in file `_parameters.pyx`?

```
>   @classmethod
E   KeyError: 'discountfactorparameter'

pywr/parameters/_parameters.pyx:1679: KeyError

During handling of the above exception, another exception occurred:

self = <test_parameters.TestDiscountFactorParameter object at 0x11f2837f0>

    def test_discount_json(self):
        """ Test loading a DiscountFactorParameter from JSON. """
>       model = load_model("discount.json")

tests/test_parameters.py:1794: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/helpers.py:17: in load_model
    model = Model.loads(data, path=path, solver=solver)
pywr/_model.pyx:189: in pywr._model.Model.loads
    ???
pywr/_model.pyx:386: in pywr._model.Model.load
    return components_data
pywr/_model.pyx:916: in pywr._model.listify.wrapper
    self.version = version
pywr/_model.pyx:382: in load_components
    def collect_components(data, key):
pywr/parameters/_parameters.pyx:1715: in pywr.parameters._parameters.load_parameter
    super(OffsetParameter, self).__init__(model, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   parameter = load_parameter(model, data.pop("parameter"))
E   TypeError: Unknown parameter type: "DiscountFactorParameter"

pywr/parameters/_parameters.pyx:1681: TypeError
```

Screenshot 2020-06-24 at 11 42 02

pywr/parameters/parameters.py
KevisPachos: Thanks, I fixed that should appear in the next commit.

tests/test_discount.py
jetuk: Sorry I missed replying to this. The main question is whether you want to support the discount rate varying as a `Parameter`. The problem is the current implementation in `value` is assuming a fixed discount rate and it also does not deal with parameter dependency properly in the `__init__` method (see: https://pywr.github.io/pywr-docs/master/extending_pywr/extending_pywr_parameters.html#dependency-on-other-parameters)

By using 

```json
@classmethod
    def load(cls, model, data):
        return cls(model, **data)
```

It will pass the values in the JSON straight to `__init__` with no conversion. This is the correct way to do it because your code is treating `self.discount_rate` as a `float` and not a `Parameter` instance.

Obviously this means implementation is limited to a fixed discount rate at the moment, but let's get that right first.

tests/test_parameters.py
jetuk: Yes, this traceback is confusing me also. I think you may need to re run `python setup.py develop` on Pywr. I think `_parameters.pyx` has changed since you last compiled it and the line numbers are wrong. The test passed on Travis.

KevisPachos: Thanks James, it has now passed the test on my machine as well.

KevisPachos:
> This is working now. Does it do what you want? I.e. have you used it in conjunction with other stuff yet? If you want to you could now Cythonise it by moving to to `_parameters.pyx` and adding some `cdef` statements. This will make it much faster if you intend to use it a lot.

Yes, I have used it in conjunction with the total cost recorder to calculate a weighted total where the weights are based on discount rate of each year.
I tried to cythonise the discount parameter by adding it to `_parameters.pyx` using the same name for the class. Do I need to somehow update the test or write a new one for this? I assume I also need to remove the parameter from `parameters.py` now?

jetuk:
> Do I need to somehow update the test or write a new one for this? 

Nope. It should be functionally identical. In fact this is the point of a good test. You can now more safely make changes to the implementation with the knowledge that if you break it the test(s) will fail.

> I assume I also need to remove the parameter from parameters.py now?

Yes, remove it from `parameters.py` and place it in `_parameters.pyx`. A simple cut and paste should get you started. I.e. it will work with no further effort (Python is valid Cython) besides recompiling. However, to make it efficient you need to start typing things. 

The most difficult change is the switch to `cdef class ...` instead of `class ...`. This will require you to also add it and any attributes to `_parameters.pxd`.

pywr/parameters/_parameters.pyx
```diff
@@ -2006,5 +2006,32 @@ def load_parameter_values(model, data, values_key='values', url_key='url',
                          "Please provide either a '{}', '{}' or '{}' entry.".format(values_key, url_key, table_key, name=name, ptype=ptype))
     return values
 
+cdef class DiscountFactorParameter(Parameter):
```
jetuk, line 2009: Can you move this above line 1909 (i.e. above the functions and below the last parameter class)?

KevisPachos:
> This will require you to also add it and any attributes to `_parameters.pxd`.

I have defined `discount_rate` as a `double` instead of `float` as I saw that all other parameters in `_parameters.pxd where defined as `double`, and `base_year` as  `int`, and made them both public:

```
    cdef public double[:] discount_rate
    cdef public int[:] base_year
```
Is that how you would define those?

jetuk:
The `[:]` part means these are 1D arrays (which need allocating, etc). I think you just want a single value here for both of these:

```
    cdef public double discount_rate
    cdef public int base_year
```

KevisPachos:
> Yes, remove it from `parameters.py` and place it in `_parameters.pyx`. A simple cut and paste should get you started. I.e. it will work with no further effort (Python is valid Cython) besides recompiling. However, to make it efficient you need to start typing things

Thanks James. I am trying to re-compile pywr with the new cython code but I get an error. So I did it step-by-step to understand where the error is. As you suggested, I did a simple cut-paste of the new parameter from `parameters.py` to `_parameters.pyx`. Am I missing another step? I tried importing `DiscountFactorParameter` in `parameters.py` by adding the parameter to the `._parameters import` section, is that required? This did not solve the error though:

```
[1/1] Cythonizing pywr/parameters/_parameters.pyx

Error compiling Cython file:
------------------------------------------------------------
...
        parameter = load_parameter(model, data.pop('parameter'))
        return cls(model, parameter, **data)
PiecewiseIntegralParameter.register()


class DiscountFactorParameter(Parameter):
^
------------------------------------------------------------

pywr/parameters/_parameters.pyx:1909:0: Assignment to non-lvalue 'DiscountFactorParameter'
  File "setup.py", line 195, in <module>
    setup_package()
  File "setup.py", line 124, in setup_package
    setup(**metadata)
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/command/develop.py", line 38, in run
    self.install_for_development()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/command/develop.py", line 136, in install_for_development
    self.run_command('egg_info')
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/command/egg_info.py", line 296, in run
    self.find_sources()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/command/egg_info.py", line 303, in find_sources
    mm.run()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/command/egg_info.py", line 534, in run
    self.add_defaults()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/setuptools/command/egg_info.py", line 570, in add_defaults
    sdist.add_defaults(self)
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/command/sdist.py", line 228, in add_defaults
    self._add_defaults_ext()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/command/sdist.py", line 311, in _add_defaults_ext
    build_ext = self.get_finalized_command('build_ext')
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/cmd.py", line 299, in get_finalized_command
    cmd_obj.ensure_finalized()
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/distutils/cmd.py", line 107, in ensure_finalized
    self.finalize_options()
  File "setup.py", line 30, in finalize_options
    annotate=annotate,
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/Cython/Build/Dependencies.py", line 1096, in cythonize
    cythonize_one(*args)
  File "/Users/macuser/miniconda3/envs/py37/lib/python3.7/site-packages/Cython/Build/Dependencies.py", line 1219, in cythonize_one
    raise CompileError(None, pyx_file)
Cython.Compiler.Errors.CompileError: pywr/parameters/_parameters.pyx
```

jetuk:
If you've added it to the `.pxd` file as `cdef class DiscountFactorParameter(Parameter)`  then this definition in the `.pyx` file must match. Notice the `cdef` part.

KevisPachos:
> If you've added it to the `.pxd` file as `cdef class DiscountFactorParameter(Parameter)` then this definition in the `.pyx` file must match. Notice the `cdef` part.

Thanks James. I made the change you suggested regarding the `cdef` part but now I get the following KeyError:

```
>   def __init__(self, model, discount_rate, base_year, **kwargs):
E   KeyError: 'discountfactorparameter'

pywr/parameters/_parameters.pyx:1920: KeyError

During handling of the above exception, another exception occurred:

self = <test_parameters.TestDiscountFactorParameter object at 0x10c9290f0>

    def test_discount_json(self):
        """ Test loading a DiscountFactorParameter from JSON. """
>       model = load_model("discount.json")

tests/test_parameters.py:1794: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/helpers.py:17: in load_model
    model = Model.loads(data, path=path, solver=solver)
pywr/_model.pyx:205: in pywr._model.Model.loads
    return cls.load(data, model, path, solver, **kwargs)
pywr/_model.pyx:402: in pywr._model.Model.load
    for parameter in load_components(model._parameters_to_load, load_parameter):
pywr/_model.pyx:943: in pywr._model.listify.wrapper
    return list(f(*args, **kwargs))
pywr/_model.pyx:398: in load_components
    component = load_component(model, component_data, name)
pywr/parameters/_parameters.pyx:1956: in pywr.parameters._parameters.load_parameter
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   self.discount_rate = discount_rate
E   TypeError: Unknown parameter type: "DiscountFactorParameter"

pywr/parameters/_parameters.pyx:1922: TypeError
```

Do I need to define the value function in `_parameters.pxd`? It now reads as:
```
cdef class DiscountFactorParameter(Parameter):
    cdef public double discount_rate
    cdef public int base_year
    cpdef double value(self, discount_rate, base_year) except? -1
```

The file `_parameters.pyx` reads as:

```
cdef class DiscountFactorParameter(Parameter):
    """Parameter that returns the current discount factor based on discount rate and a base year.

    Parameters
    ----------
    discount_rate : float
        Discount rate (expressed as 0 - 1) used calculate discount factor for each year.
    base_year : int
        Discounting base year (i.e. the year with a discount factor equal to 1.0).
    """

    def __init__(self, model, discount_rate, base_year, **kwargs):
        super(DiscountFactorParameter, self).__init__(model, **kwargs)
        self.discount_rate = discount_rate
        self.base_year = base_year

    cpdef double value(self, discount_rate, base_year) except? -1:
        cdef double discount_rate
        cdef int base_year 
        cdef int ct = self.model.timestepper.current
        cdef int cy = ct.year
        cdef int cyi = cy - self.base_year
        return 1/pow((1+self.discount_rate), cyi) 

    @classmethod
    def load(cls, model, data):
        discount_rate = load_parameter(model, data.pop("value"))
        base_year = load_parameter(model, data.pop("base_year"))
        return cls(model, discount_rate, base_year, **data)

DiscountFactorParameter.register()
```
Any thoughts why this is happening?

jetuk:
> Do I need to define the value function in _parameters.pxd? It now reads as:

The `value` method is defined on the parent class. No need to redefine it here.

Are you sure the updated `_parameters.pyx` has actually compiled? The only reason I can see for it not finding the parameter is if it hasn't compiled properly (i.e. you need to run something like `python setup.py develop` every time you change a `pyx` or `pxd` file).

You load method should look like this:
```
    @classmethod
    def load(cls, model, data):
        return cls(model, **data)
```

KevisPachos:
> You load method should look like this:

Thanks for your help James, I was indeed not compiling every time I made a change to the Cython code.. 
In my last commit I have made it pass the test with the cythonised discount parameter. Also I changed the `load` method to look as you suggested. Do you want to test it on Travis too?

I did get a `OrphanedParameterWarning` though, here is the full message:

```
========================================================= warnings summary =========================================================
test_parameters.py::TestDiscountFactorParameter::test_discount_json
  /Users/macuser/Github/pywr/tests/helpers.py:18: OrphanedParameterWarning: Model has 1 orphaned parameters
    model.check()
```
Do I need to make any changes to remove this?

jetuk:
@KevisPachos I have resolved the conflict. Please pull your branch before making any further commits.

> Do I need to make any changes to remove this?

It's nice to not to have warnings in the tests. This one is because you are not using the output of the parameter in the model anywhere. Perhaps change the `max_flow` of the output node to the discount factor:

```json
        {
            "name": "demand",
            "type": "Output",
            "max_flow": "discount_factor",
            "cost": -100
        }

```
It's a bit silly, but it'll get rid of the warning.

KevisPachos:
Thanks James, I managed to remove the warning with the change in the `json` file you suggested. I re-run the tests and all looks okay.
What is the next step on the merge request now?

jetuk:
I think we're good. I'll do one last check and then merge it.

Thanks for the contribution!

pywr/parameters/_parameters.pyx
```diff
@@ -1882,6 +1882,34 @@ cdef class PiecewiseIntegralParameter(Parameter):
...
+    base_year : int
+        Discounting base year (i.e. the year with a discount factor equal to 1.0).
+    """
+
+    def __init__(self, model, rate, base_year, **kwargs):
+        super(DiscountFactorParameter, self).__init__(model, **kwargs)
+        self.rate = rate
+        self.base_year = base_year
+
+    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
```
jetuk, line 1901: This can be simplified to:
```
    cpdef double value(self, Timestep ts, ScenarioIndex scenario_index) except? -1:
        return 1 / pow(1.0 + self.rate, ts.year - self.base_year)
```

By using `ts` directly it'll be more efficient than looking up `self.model.timestepper.current`.

tests/test_parameters.py
```diff
@@ -1786,3 +1786,17 @@ def test_variable_api(self, simple_linear_model):
...
+        """ Test loading a DiscountFactorParameter from JSON. """
+        model = load_model("discount.json")
+        # run model for period 2015-2020, with base year 2015 and discount rate of 0.035 (3.5%)
+        p = model.parameters['discount_factor']
+        @assert_rec(model, p)
+        def expected_func(timestep, scenario_index):
+            year = timestep.year
+            return 1/pow(1.035, year - 2015)
+
+        model.run()
```
jetuk, line 1802: Add an empty newline.

pywr/parameters/_parameters.pxd
```diff
@@ -200,3 +200,7 @@ cdef class PiecewiseIntegralParameter(Parameter):
     cdef public double[:] x
     cdef public double[:] y
     cdef public Parameter parameter
+
+cdef class DiscountFactorParameter(Parameter):
+    cdef public double rate
+    cdef public int base_year
```
jetuk, line 206: Add an empty newline.

tests/models/discount.json
```diff
@@ -0,0 +1,35 @@
...
+        ["supply", "demand"]
+    ],
+    "parameters": {
+        "discount_factor": {
+            "type": "DiscountFactorParameter",
+            "base_year": 2015,
+            "rate": 0.035
+        }
+    }
+}
```
jetuk, line 35: Add an empty newline.

jetuk:
I just noticed a simplification, and a couple of newlines at end of files. I'll do those tomorrow unless you get to them first.

KevisPachos:
> I just noticed a simplification,

Thank you James, I have made the changes you suggested. Let me know if you pick up anything else.

jetuk:
Thanks @KevisPachos !

PR #912, opened 2020-07-16
Selective bounds and objective updates for the glpk-edge solver

jetuk:
This adds a couple of optional flags to the `glpk-edge` solver:

- `set_fixed_flows_once` - If a non-storage node has a float for `min_flow` and `max_flow` (i.e. not a parameter). Then it identifies those nodes as being "fixed". These nodes have their constraint bounds defined once during setup. Other non-fixed nodes have their constraints updated each time-step as usual. 
- `set_fixed_costs_once` - If a non has the property `has_fixed_cost` and it is True then these fixed costs are pre-cached, and not evaluated every time-step. This is a bit more fiddly because the codes are mapped to the edges. 

Both changes reduce the number of nodes which are looped through to only those with non-fixed flows and/or costs. This has a nice speed improvement in larger models where a high proportion of nodes have fixed attributes. I've seen a roughly 5% speed improvement in one large model for each of these settings.

The same approach could be applied to the regular `glpk` solver.

=====================================================

Actions to complete:
- [x] Implement for regular `glpk` solver.
- [x] Provide a mechanism for solver settings to be reported (serialised) as part of `ModelResult`.
- [x] Add test suite runs with new settings to CI.


jetuk:
A quick update on this. These options have been tested on a large model with and without the new options enabled. Results were identical.

pywr/_core.pyx
```diff
@@ -527,6 +527,11 @@ cdef class Node(AbstractNode):
             return self._cost
         return self._cost_param.get_value(scenario_index)
 
+    property has_fixed_cost:
+        """Returns true if both cost is not a Parameter."""
```
Batch21, line 531: delete 'both'

pywr/_core.pyx
```diff
@@ -787,6 +797,11 @@ cdef class StorageInput(BaseInput):
         # Return negative of parent cost
         return -self.parent.get_cost(scenario_index)
 
+    property has_fixed_cost:
+        """Returns true if both cost is not a Parameter."""
```
Batch21, line 801: delete 'both'

pywr/_core.pyx
```diff
@@ -796,6 +811,10 @@ cdef class StorageOutput(BaseOutput):
         # Return parent cost
         return self.parent.get_cost(scenario_index)
 
+    property has_fixed_cost:
+        """Returns true if both cost is not a Parameter."""
```
Batch21, line 815: delete 'both'

pywr/_core.pyx
```diff
@@ -867,6 +886,11 @@ cdef class Storage(AbstractStorage):
             return self._cost
         return self._cost_param.get_value(scenario_index)
 
+    property has_fixed_cost:
+        """Returns true if both cost is not a Parameter."""
```
Batch21, line 890: delete 'both'

pywr/solvers/cython_glpk.pyx
```diff
@@ -961,41 +973,81 @@ cdef class CythonGLPKEdgeSolver(GLPKSolver):
...
+                fixed_cost = False
+
+            if self.set_fixed_costs_once and fixed_cost:
+                # With a fixed cost this should work with no scenario index
+                cost = some_node.get_cost(None)
+                data = some_node.__data
+
+                # Link nodes have edges connected upstream & downstream. We apply
+                # half the cost assigned to the node to all the connected edges.
+                # The edge costs are then the mean of the node costs at either end.
```
Batch21, line 991: Only when its two link nodes?

jetuk:
I think we should also add this mode(s) to the tests.

pywr/solvers/cython_glpk.pyx
jetuk: Correct. I'll update the comment.

Batch21:
I got a few errors when using the new environment variables to run the tests:

image

It looks like these are the ones you mentioned that use `.step()` and then manually change flow/cost attributes on the mode. One suggestion I had was to make the default of the two new solver option true, as I can't see any good reasons why you wouldn't use them given the potential speed benefits.

jetuk:
One reason would be that it breaks for anyone relying on this functionality. E.g. if Pywr is integrated into some other system.

The test suite can probably be modified to to skip if the environment variable is detected/.

Batch21:
> One reason would be that it breaks for anyone relying on this functionality. E.g. if Pywr is integrated into some other system.

Ok that makes sense

Would it be worth adding the value of the variables to the `ModelResult` object so that its clearer to a user when they are being used?

jetuk:
> Would it be worth adding the value of the variables to the ModelResult object so that its clearer to a user when they are being used?

Yes, and probably a number of the other settings/options that are already in the solvers.

I'll update the top comment with some tasks.

jetuk:
@Batch21 this is ready now. I've implemented it for the other GLPK solver, enabled the settings in the CI (only on Linux) and added a new "solver_settings" section the `ModelResults` object.

The CI nows takes a LONG time to run through all solver settings and Python versions.

Batch21 (changes requested):
Couple of typos but otherwise looks good. Would it be possible to also check fixed flows on the aggregated nodes or would that get complicated? 

I did a quick test on a complex model and theses changes increase speed by 10-15%!

pywr/solvers/cython_glpk.pyx
```diff
@@ -1003,41 +1051,81 @@ cdef class CythonGLPKEdgeSolver(GLPKSolver):
         if len(cross_domain_cols) > 0:
             self.idx_row_cross_domain = glp_add_rows(self.prob, len(cross_domain_cols))
 
+        self.edge_fixed_cost_arr[...] = 0.0
+        # Calculate fixed costs
+        for some_node in self.all_nodes:
```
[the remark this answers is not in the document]
jetuk: This is because `node` is type `Node` which is need for the flow constraints, but `some_node` is `AbstractNode` which is any node including storage, aggregated, etc.

jetuk:
> I did a quick test on a complex model and theses changes increase speed by 10-15%!

But did it give the same results?!

PR #915, opened 2020-08-06
Add BisectionSearchModel

jetuk:
This is a new subclass of Model that performs a bisection search
by varying a single parameter's value to find its largest value
that satisfies all model constraints.

There are updates to allow loading of different Model subc-classes
from JSON and also in the optimisation wrapper.

This addresses the "English & Welsh" method of #7 .


pywr/utils/bisect.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if self.bisect_epsilon <= 0.0:
+            raise ValueError("Bisection epsilon value must be greater than zero.")
+
+        best_feasible = -np.inf
+        while (max_value - min_value) > self.bisect_epsilon:
+            # Compute the current value to try as the middle of the bounds
+            current_value = (max_value + min_value) / 2
+
+            print(min_value, current_value, max_value)
```
Batch21, line 59: Could use a logger instead of print and maybe include some runtime information

pywr/utils/bisect.py
```diff
@@ -0,0 +1,78 @@
...
+            if self.is_feasible():
+                # No constraint failures; tighten lower bounds
+                min_value = current_value
+                best_feasible = max(best_feasible, current_value)
+            else:
+                # Failed; tighten upper bounds
+                max_value = current_value
+
+        # Finally, rerun at the best found feasible value
+        # This is slightly inefficient, but ensures the model's recorders are at the best_feasible
```
Batch21, line 73: The only alternative I can think of would be to store previous model runs which would likely be more inefficient than this

pywr/utils/bisect.py
```diff
@@ -0,0 +1,78 @@
...
+        # value's results at the end of the simulation.
+        param.set_double_variables(np.array([best_feasible]))
+        return super().run()
+
+
```
Batch21, line 78: delete extra lines

tests/test_bisect.py
```diff
@@ -0,0 +1,51 @@
...
+    assert p.problem.nvars == 1
+    assert p.problem.nobjs == 1
+    assert p.problem.nconstrs == 1
+    # Check the correct model class is created by the wrapper.
+    assert isinstance(p.model, BisectionSearchModel)
+
+
+def test_platypus_nsgaii_step(simple_bisection_problem):
+    """ Undertake a single step of the NSGAII algorithm with a small population. """
+    algorithm = NSGAII(simple_bisection_problem.problem, population_size=10)
```
Batch21, line 50: If I understand this correctly, each population member is going to run the bisect search, so, for example, you could use this setup to optimise curves for maximum DO?

jetuk: Correct. Which is why it is important to have the model end with the final feasible values so that the optimisation wrappers pick up the correct objective values.

pywr/utils/bisect.py
jetuk: It's a bit of mess to do that. Requires creating a whole new model for every run and keeping it. That doesn't really work as a subclass of Model.

jetuk:
@Batch21 I've addressed your comments. I also added a new infeasible test to make sure this errors if it can't satisfy the constraints.

PR #917, opened 2020-08-07
Fix timeseries1_weekly.h5 to work with pandas v1.1.0

### LINKED ISSUE #916: Test failures with pandas v1.1.0
We're getting some test failures with the new version of pandas.

```
    @property
    def freqstr(self):
        """
        Return the frequency object as a string if its set, otherwise None.
        """
        if self.freq is None:
            return None
>       return self.freq.freqstr
E       AttributeError: 'numpy.bytes_' object has no attribute 'freqstr'

/opt/_internal/cpython-3.8.1/lib/python3.8/site-packages/pandas/core/arrays/datetimelike.py:1104: AttributeError
```

Here for full log: https://api.travis-ci.org/v3/job/715545101/log.txt

jetuk:
Recent versions of pandas were unable to load this file. The
file has been recreated from the equivalent CSV with a 7D
frequency.

Also tidied up a few unneeded attributes in the associated test
models.

Fixes #916.


KevisPachos:
Hi @jetuk, 
How can I update my `h5` file to be compatible with Pandas v.1.1.0?
For instance, I have a file with 180 flow scenarios for 5 catchments called `catchmod_flows_by_catchment_sub200.h5` that you had shared with me in the past. 
Can I make changes to the existing file? Or do I need to recreate it by changing something?
And if so, how do I do that?
Thanks!

jetuk:
@KevisPachos for this one I recreated it from the CSV data. The issue is here in pandas: https://github.com/pandas-dev/pandas/issues/33186

The only work around is to open it in an old version of pandas, save to another format (e.g. CSV), read in new version of pandas and then save to HDF again.

KevisPachos:
@jetuk, oh I see. It is a matter of importing and exporting using different pandas version. This is where virtual environments will be handy :) Thanks James

PR #919, opened 2020-08-16
Dynamic factors for Aggregated Node

Batch21:
@jetuk I think I've got this working for a simple test case for the original glpk solver. As you suggested, I've cached the row and col information for the agg constraints in a simple class, however storing all the column information as a simple list is probably not the most efficient approach. Had a play around with some other approaches using different data types without much success. Any suggestions?

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -92,6 +92,7 @@ cdef class AggregatedNode(AbstractNode):
     cpdef double get_min_flow(self, ScenarioIndex scenario_index) except? -1
     cpdef double get_max_flow(self, ScenarioIndex scenario_index) except? -1
     cpdef double[:] get_factors(self, ScenarioIndex scenario_index)
+    cpdef double[:] get_factors_norm(self, ScenarioIndex scenario_index)
 
 cdef class BaseInput(Node):
     cdef object _licenses
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -81,13 +81,14 @@ cdef class Node(AbstractNode):
 
 cdef class AggregatedNode(AbstractNode):
     cdef list _nodes
-    cdef double[:] _factors
-    cdef list _factor_parameters
+    cdef list _factors
     cdef double[:] _flow_weights
     cdef double _max_flow
     cdef double _min_flow
     cdef Parameter _min_flow_param
     cdef Parameter _max_flow_param
+    cdef public bint has_fixed_factors
+    cdef public object __agg_factor_data 
 
     cpdef double get_min_flow(self, ScenarioIndex scenario_index) except? -1
     cpdef double get_max_flow(self, ScenarioIndex scenario_index) except? -1
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -87,8 +87,7 @@ cdef class AggregatedNode(AbstractNode):
     cdef double _min_flow
     cdef Parameter _min_flow_param
     cdef Parameter _max_flow_param
-    cdef public bint has_fixed_factors
-    cdef public object __agg_factor_data 
+    cdef public object __agg_factor_data
 
     cpdef double get_min_flow(self, ScenarioIndex scenario_index) except? -1
     cpdef double get_max_flow(self, ScenarioIndex scenario_index) except? -1
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -82,6 +82,7 @@ cdef class Node(AbstractNode):
 cdef class AggregatedNode(AbstractNode):
     cdef list _nodes
     cdef double[:] _factors
+    cdef list _factor_parameters
     cdef double[:] _flow_weights
     cdef double _max_flow
     cdef double _min_flow
@@ -90,6 +91,7 @@ cdef class AggregatedNode(AbstractNode):
 
     cpdef double get_min_flow(self, ScenarioIndex scenario_index) except? -1
     cpdef double get_max_flow(self, ScenarioIndex scenario_index) except? -1
+    cpdef double[:] get_factors(self, ScenarioIndex scenario_index)
 
 cdef class BaseInput(Node):
     cdef object _licenses
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -672,10 +672,12 @@ cdef class AggregatedNode(AbstractNode):
 
     property factors:
         def __get__(self):
-            if self._factors is None:
-                return None
-            else:
+            if self._factors is not None:
                 return np.asarray(self._factors, np.float64)
+            elif self.factor_parameters is not None:
+                return self.factor_parameters
+            else:
+                return None
 
         def __set__(self, values):
             values = np.array(values, np.float64)
@@ -683,6 +685,7 @@ cdef class AggregatedNode(AbstractNode):
                 warnings.warn("Very small factors in AggregateNode result in ill-conditioned matrix")
             self._factors = values
             self.model.dirty = True
+
 
     property factor_parameters:
         def __get__(self):
@@ -789,16 +792,22 @@ cdef class AggregatedNode(AbstractNode):
 
     @classmethod
     def load(cls, data, model):
-        from pywr.parameters import load_parameter
+        from pywr.parameters import load_parameter, ConstantParameter
         name = data["name"]
         nodes = [model._get_node_from_ref(model, node_name) for node_name in data["nodes"]]
         agg = cls(model, name, nodes)
         try:
-            agg.factors = data["factors"]
-        except KeyError:
-            pass
-        try:
-            agg.factor_parameters = [load_parameter(p) for p in data["factor parameters"]]
+            factors = data["factors"]
+            if all([isinstance(x, (float, int)) for x in factors]):
+                agg.factors = factors
+            else:
+                parameters = []
+                for f in factors:
+                    if isinstance(f, (float, int)):
+                        parameters.append(ConstantParameter(model, f))
+                    else:
+                        parameters.append(load_parameter(model, f))
+                agg.factor_parameters = parameters
         except KeyError:
             pass
         try:
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -680,12 +680,15 @@ cdef class AggregatedNode(AbstractNode):
                 return None
 
         def __set__(self, values):
-            values = np.array(values, np.float64)
-            if np.any(values < 1e-6):
-                warnings.warn("Very small factors in AggregateNode result in ill-conditioned matrix")
-            self._factors = values
-            self.model.dirty = True
-
+            from pywr.parameters import Parameter
+            if any([isinstance(v, Parameter) for v in values]):
+                self.factor_parameters = values
+            else:
+                values = np.array(values, np.float64)
+                if np.any(values < 1e-6):
+                    warnings.warn("Very small factors in AggregateNode result in ill-conditioned matrix")
+                self._factors = values
+                self.model.dirty = True
 
     property factor_parameters:
         def __get__(self):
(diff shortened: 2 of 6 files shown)


jetuk:
> Had a play around with some other approaches using different data types without much success. Any suggestions?

I think the most efficient approach would be to store a point to pre-allocated (during setup) arrays of the columns and resptive values. Only the values need to change during each time-step by overwriting the already allocated data. This would avoid the repeat malloc/free cycle in each solve.  

However, I don't really know the performance implications of this versus what you've done here.

tests/test_agg_constraints.py
```diff
@@ -279,3 +280,38 @@ def test_multipiecewise_constraint(model, flow):
...
+    B = Input(model, "B", max_flow=10.0)
+    C = Input(model, "C", max_flow=10.0)
+    Z = Output(model, "Z", max_flow=10.0, cost=-10)
+
+    agg = AggregatedNode(model, "agg", [A, B, C])
+    factor1 = DailyProfileParameter(model, np.append(np.array([0.8, 0.3]), np.ones(364)))
+    factor2 = DailyProfileParameter(model, np.append(np.array([0.1, 0.3]), np.ones(364)))
+    factor3 = DailyProfileParameter(model, np.append(np.array([0.1, 0.4]), np.ones(364)))
+
+    agg.factor_parameters = [factor1, factor2, factor3]
```
jetuk, line 299: Can we make this `agg.factors = [factor1, factor2, factor3]`. I think external API is the same regardless of whether you are using floats or parameters.

Batch21:
I tried storing a pointer reference as an attribute on the the `AggNodeFactorData` helper class but I don't think this is allowed as it causes a `Cannot convert 'int *' to Python object` error

Does the data being passed to `set_mat_row` have to pointers? One alternative might be to cache the column data as cython arrays and use these directly or maybe there is an effiecient way to convert them to pointers?

tests/test_agg_constraints.py
Batch21: I'll change this

jetuk:
This following is valid, but obviously you can't access it from Python.

```
cdef class AbstractAggNodeData:
    cdef public int* ind
    cdef public double* val
```

And, yes, we could use cython arrays here. You can get the raw pointer from `my_array.data.as_voidptr`, and give that to the GLPK routines. See here: https://cython.readthedocs.io/en/latest/src/tutorial/array.html

jetuk:
Do you know why the test fails?

Batch21:
> Do you know why the test fails?

I think its just for the `glpk-edge` solver as I haven't made any changes for that yet.

jetuk:
OK let me know if you want me to review properly. I just had a quick look through this morning.

pywr/solvers/cython_glpk.pyx
```diff
@@ -677,6 +723,29 @@ cdef class CythonGLPKSolver(GLPKSolver):
...
+
+                ind = agg_node.__agg_factor_data.ind[indptr_array[n]:indptr_array[n+1]]
+                val = agg_node.__agg_factor_data.val[indptr_array[n]:indptr_array[n+1]]
+                length = len(ind) - 1
+
+                # only update factors for second node of each row
+                for i in range(agg_node.__agg_factor_data.node_ind, val.shape[0]):
+                    val[i] = -factors_norm[n+1]
+
+                set_mat_row(self.prob, row+n, length, &ind[0], &val[0])
```
jetuk, line 746: This is definitely more how I envisaged this. No re-allocation during the timesteps.

pywr/solvers/cython_glpk.pyx
```diff
@@ -632,15 +678,15 @@ cdef class CythonGLPKSolver(GLPKSolver):
...
-        cdef double* val
         cdef double lb
         cdef double ub
         cdef Timestep timestep
         cdef int status, simplex_ret
         cdef cross_domain_col
-        cdef list route
+        cdef list route, indptr_array
         cdef int node_id, indptr, nroutes
+        cdef int[::1] ind
```
jetuk, line 688: What does the `[::1]` syntax mean? Can't this just be `[:]`?

pywr/solvers/cython_glpk.pyx
```diff
@@ -677,6 +723,29 @@ cdef class CythonGLPKSolver(GLPKSolver):
         self.stats['objective_update'] += time.perf_counter() - t0
         t0 = time.perf_counter()
 
+        for agg_node in self.aggregated_with_factor_params:
+            factors = agg_node.get_factors(scenario_index)
+
+            # normalise factors
+            f0 = factors[0]
```
jetuk, line 730: Some of these temporary variables can be `cdef`'d. Check the annotated output by running `setup.py --annotate` and looking at the generated HTML file for this module.

pywr/solvers/cython_glpk.pyx
Batch21: It ensures the array is contiguous in memory. I think cython arrays are by default but I've seen `[::1]`used in a few examples where syntax like `&ind[0]` is used to get a pointer reference.

What I'm not sure about is whether setting a slice of the `agg_node.__agg_factor_data.ind` array to the `ind` variable in `_solve_scenario` involves the reallocation of some memory

Batch21:
@jetuk I think this is probably ready for a review

pywr/solvers/cython_glpk.pyx
```diff
@@ -343,7 +357,10 @@ cdef class CythonGLPKSolver(GLPKSolver):
             elif isinstance(some_node, Storage):
                 storages.append(some_node)
             elif isinstance(some_node, AggregatedNode):
-                if some_node.factors is not None:
+                if some_node.factor_parameters is not None:
+                    aggregated_with_factor_params.append(some_node)
+                    some_node.__agg_factor_data = AggNodeFactorData()
```
jetuk, line 362: Please add `__agg_factor_data` to the definition of the `AggregatedNode` in a similar way to the `__data` attribute.

pywr/_core.pyx
```diff
@@ -754,14 +779,42 @@ cdef class AggregatedNode(AbstractNode):
...
 
+    cpdef double[:] get_factors(self, ScenarioIndex scenario_index):
+        """Get node factor for a give timestep
+        """
+        if  self._factor_parameters is None:
+            return self.factors
+        else:
+            return np.array([p.get_value(scenario_index) for p in self.factor_parameters], np.float64)
+
+    cpdef double[:] get_factors_norm(self, ScenarioIndex scenario_index):
```
jetuk, line 790: This could be made more efficient if an "out" array was an optional argument. Instead of creating a new array the function would write to the provided array instead.

jetuk (changes requested):
Changes look good to me. My only real suggestion is to simply the factors and convert everything to parameters. I think this will make it easier to maintain and understand going forward.

pywr/_core.pxd
```diff
@@ -90,6 +91,8 @@ cdef class AggregatedNode(AbstractNode):
 
     cpdef double get_min_flow(self, ScenarioIndex scenario_index) except? -1
     cpdef double get_max_flow(self, ScenarioIndex scenario_index) except? -1
+    cpdef double[:] get_factors(self, ScenarioIndex scenario_index)
```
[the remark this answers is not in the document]
Batch21: I don't think `except? -1` works when return type is `double[:]`?

pywr/solvers/cython_glpk.pyx
```diff
@@ -527,6 +544,36 @@ cdef class CythonGLPKSolver(GLPKSolver):
                 # glp_set_row_name(self.prob, row+n,
                 #                  'ag.f{}.{}'.format(n, agg_node.fully_qualified_name).encode('utf-8'))
 
+        
+        if len(aggregated_with_factor_params):
+            self.idx_row_aggregated_with_factors = self.idx_row_virtual_storages + len(aggregated_with_factors)
```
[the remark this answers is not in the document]
Batch21: Isn't it done this way because at this point the number or rows each agg nodes needs has node be calculated?

pywr/solvers/cython_glpk.pyx
```diff
@@ -527,6 +544,36 @@ cdef class CythonGLPKSolver(GLPKSolver):
...
+                agg_node.__agg_factor_data.ind_ptr[i] = v
+
+            agg_node.__agg_factor_data.inds = cvarray(shape=(len(cols),), itemsize=sizeof(int), format="i")
+            agg_node.__agg_factor_data.vals = cvarray(shape=(len(cols),), itemsize=sizeof(double), format="d")
+            for i, v in enumerate(cols):
+                agg_node.__agg_factor_data.inds[i] = v
+                agg_node.__agg_factor_data.vals[i] = 1.0
+
+            for n in range(len(nodes)-1):
+                set_row_bnds(self.prob, row+n, GLP_FX, 0.0, 0.0)
```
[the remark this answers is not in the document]
Batch21: The bounds don't get updated in solve just the row values

pywr/solvers/cython_glpk.pyx
```diff
@@ -1131,6 +1213,42 @@ cdef class CythonGLPKEdgeSolver(GLPKSolver):
...
+            cols = []
+            ind_ptr = [0,]
+            if isinstance(nodes[0], BaseOutput):
+                first_node_cols = [0] + [c+1 for c in nodes[0].__data.in_edges]
+            else:
+                first_node_cols = [0] + [c+1 for c in nodes[0].__data.out_edges]
+
+            agg_node.__agg_factor_data.node_ind = len(first_node_cols)
+            for i, some_node in enumerate(nodes[1:]):
+                if isinstance(nodes[0], BaseOutput):
```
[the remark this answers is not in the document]
Batch21: yeah, looks like it. Do you think that a test that needs adding?

pywr/_core.pyx
Batch21: Is it possible to have an optional arg for a cpdef function. From a quick test, I was unable to get this to work.

jetuk: Yes, see `_reset_storage_only` in this file. You need to declare it in the pxd.

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#optional-arguments

pywr/_core.pxd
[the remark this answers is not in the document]
jetuk: Oh maybe it's not needed for a memoryview because that's a Python object?

Batch21:
@jetuk thanks for the review. I've made some updates and responded to some of the comments

What I've done with the factors property still feels a little messy. I've tried to make sure it will deal with the cases where factors are set on the node object after it has been created. This is done in many of the tests and also in the init of `MultiSplitLink`

jetuk:
@Batch21 I've made the changes we discussed today. This will now always update the factors in the constraint matrix instead of only if they are parameters. Making it so efficient by only updating the dynamic factors makes the internal implementation more complicated than necessary. I suggest we do it like this and if someone discovers that only updating dynamic factors is a massive bottleneck in their performance do we revisit this.

However, probably worth checking with a realish example that it's not horrendously slowly for some silly reason.

jetuk:
OK, so this is broken for lpsolve. With the changes its either implement it correctly or removed aggregated node factors entirely. I think we have to do the former.

Batch21:
@jetuk your changes look good. I tested them on a model that contains 7 agg nodes with fixed factors and 5 with dynamic factors. In comparison to the last commit I made on this branch, it doesn't look like your changes had any negative impact on runtimes:

|   | total time | factor update time| factor update % of total time|
|-|-|-|-|
|Update dynamic factors only|37.42747 | 8.635756 | 0.230802
|Update all factors| 34.87279845 | 7.553858556 | 0.216891663

Did you want me to have a look at the lpsolve updates?

jetuk:
> Did you want me to have a look at the lpsolve updates?

I'm looking at it now. It's not easy to update individual constraints in lpsolve (you have to delete rows and re-add). I'm just going to patch it to work with only fixed factors as we originally intended.

jetuk:
@Batch21 I think I am happy with this now. Can you check last lpsolve changes and confirm it looks OK? We can then merge.

jetuk (approved):

Batch21:
I'm happy with lpsolve changes

PR #922, opened 2020-08-22
Fix AggregatedStorage initial volume when defined as proportions.

### LINKED ISSUE #918: AggregatedStorage initial volume is broken when using proportional values
I'm pretty sure if the reservoirs referenced by the `AggregatedStorage` node have their volume defined by as a proportion then this will not create the correct initial volume in the aggregated node. This needs a regression test.

Relevant lines:

https://github.com/pywr/pywr/blob/db1c8baa71a48b581466456b77fefdf6944eb37e/pywr/_core.pyx#L1081

https://github.com/pywr/pywr/blob/db1c8baa71a48b581466456b77fefdf6944eb37e/pywr/_core.pyx#L1093

jetuk:
This refactors the reset storage code to provide two helper
functions that return the initial volume in absolute and
proprtional terms. These calculations depend on how the initial
and maximum volume is defined.

The AggregatedStorage node uses one these new functions to
correctly calculate its initial volume.

Fixes #918.

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -110,6 +110,8 @@ cdef class Storage(AbstractStorage):
     cdef Parameter _max_volume_param
     cdef Parameter _level_param
     cdef Parameter _area_param
+    cpdef double get_initial_volume(self) except? -1
+    cpdef double get_initial_pc(self) except? -1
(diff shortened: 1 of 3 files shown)


Batch21 (approved):
Looks good to me

PR #923, opened 2020-08-28
Add SeasonalVirtualStorage Node

Batch21:


tests/test_run.py
```diff
@@ -441,6 +441,57 @@ def test_annual_virtual_storage_with_dynamic_cost():
     assert_allclose(rec.data[3], 5)
 
 
+def test_seasonal_virtual_storage():
```
jetuk, line 444: Can you put these tests in a class for organisational purposes.

I think we need to also test the following:

- Case with a licence ending before the reset date (i.e. a licence crossing two years).
- The value of a non-depleted licence outside of the active period. I suspect that the licence's volume is still lowered even though it is not active. It should remain unchanged outside the active period?

jetuk (changes requested):
This is a good change. It finally fixes #2! However, I think there are a couple of edge cases that need testing for.

pywr/nodes.py
```diff
@@ -623,6 +623,45 @@ def before(self, ts):
                 # Reset to maximum volume (i.e. full capacity. )
                 self._reset_storage_only(use_initial_volume=self.reset_to_initial_volume)
                 self._last_reset_year = ts.year
+                self.active = True
+
+
+class SeasonalVirtualStorage(AnnualVirtualStorage):
```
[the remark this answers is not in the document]
Batch21: PeriodVirtualStorage, SubAnnualVirtualStorage???

I can't think of a good one

tests/test_run.py
Batch21: Done. The volume of the node is unchanged outside of the active period

pywr/solvers/cython_glpk.pyx
```diff
@@ -1396,6 +1398,8 @@ cdef class CythonGLPKEdgeSolver(GLPKSolver):
 
             if max_volume == min_volume:
                 set_row_bnds(self.prob, self.idx_row_virtual_storages+row, GLP_FX, 0.0, 0.0)
+            elif not storage.active:
+                set_row_bnds(self.prob, self.idx_row_virtual_storages+row, GLP_FR, -np.inf, np.inf)
```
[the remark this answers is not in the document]
Batch21: changed

pywr/solvers/cython_glpk.pyx
```diff
@@ -765,6 +765,8 @@ cdef class CythonGLPKSolver(GLPKSolver):
...
```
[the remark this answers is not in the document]
Batch21: changed

pywr/nodes.py
```diff
@@ -623,6 +623,45 @@ def before(self, ts):
...
+        self.end_month = kwargs.pop('end_month', 12)
+        self._last_active_year = None
+
+        super().__init__(*args, **kwargs)
+
+    def before(self, ts):
+        super().before(ts)
+
+        if ts.year != self._last_active_year:
+            if ts.month > self.end_month or \
```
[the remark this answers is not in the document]
Batch21: Added test to show that node works when end date is earlier in year than reset date

Batch21:
@jetuk fixed a couple of edge cases that meant the node was not working correctly for the initial year. The nested conditionals for this are a little ugly but it works.

I still can't think of a better name..

jetuk:
OK. I'll read the new tests. I think the docstring could be better still.

pywr/_core.pyx
```diff
@@ -1210,6 +1210,7 @@ cdef class VirtualStorage(Storage):
     def __cinit__(self, ):
         self._allow_isolated = True
         self.virtual = True
+        self.active = True
```
jetuk, line 1213: I think we should set this to `True` in a reset method. Then it ensures it resets to the init value.

Batch21: done

Batch21:
@jetuk I've extending the docstring a little. Is there any more information you think should be captured in it?

pywr/nodes.py
```diff
@@ -623,6 +623,68 @@ def before(self, ts):
                 # Reset to maximum volume (i.e. full capacity. )
                 self._reset_storage_only(use_initial_volume=self.reset_to_initial_volume)
                 self._last_reset_year = ts.year
+                self.active = True
+
+
+class SeasonalVirtualStorage(AnnualVirtualStorage):
+    """A virtual storage node that operates for a specified period within a year. This is useful for representing
```
jetuk, line 630: I think this should just be one sentence.

pywr/nodes.py
```diff
@@ -623,6 +623,68 @@ def before(self, ts):
...
+    when it stops operating. For the period when the node is not operating, the volume of the node remains unchanged
+    and the node does not apply any constraints to the model.
+
+    The end_day and end_month can represent a date earlier in the year that the reset_day and and reset_month. This
+    situation represents a licence that operates across a year boundary. For example, one that is active between
+    October and March and not active between April and September.
+
+    Parameters
+    ----------
+    reset_day: int
```
jetuk, line 643: There should be a space either side of the colon in these.

pywr/nodes.py
```diff
@@ -623,6 +623,68 @@ def before(self, ts):
...
+                    # First timestep is earlier in year than reset date
+                    if self.end_month > self.reset_month or \
+                            (self.end_month == self.reset_month and self.end_day >= self.reset_day):
+                        # end date is later in year than reset date so node needs to be deactivated
+                        self.active = False
+            elif ts.month > self.end_month or \
+                    (ts.month == self.end_month and ts.day >= self.end_day):
+                self._last_active_year = ts.year
+                self.active = False
+
```
jetuk, line 686: Remove additional spaces.

jetuk (approved):

jetuk:
Thanks @Batch21! We can close #2!

PR #925, opened 2020-09-10
Add a threshold when checking GLPK constraints bounds equality.

jetuk:
Use a small threshold in the difference between upper bounds and lower bounds for a constraint to qualify as GLP_FX. This is a practical consideration where exact floating point equality might not be true in all cases.

May help with #728.

My only comment on this is whether this threshold should be a configurable setting in the GLPK solvers.


Batch21 (approved):
I happy with it not being configurable. Can always be made so in the future, if required.

Issue #926, opened 2020-09-14
Read csv files for InterpolatedVolumeParameter


VJander:
In a past issue it is made clear that `InterpolatedVolumeParameter` can't read volume and values from external data, and are limited to direct array coding. I need to use a level-volume curve that has over 3000 values and doesn't adjust quite well to many functions, so i need to add the feature of reading external data to this parameter, which files to i have to modify in order to make this pull request? And also, would you like me to upload a `InterpolatedVolumeParameter` test in case of doing it?

jetuk:
Hi @VJander 

I think this would be a good addition. We already have a generic function for load external data: `pywr.parameters._parameters.load_parameter_values`. Therefore, I think you can probably modify the `load` method of `InterpolatedVolumeParameter` https://github.com/pywr/pywr/blob/master/pywr/parameters/parameters.py#L142

Something like this should work:

```python
    @classmethod
    def load(cls, model, data):
        node = model._get_node_from_ref(model, data.pop("node"))

        volumes = data.pop("volumes")
        if isinstance(volumes, list):
            volumes = np.array("volumes")
        elif isinstance(volumes, dict):
            volumes = load_parameter_values(model, volumes)
        else:
            raise TypeError("Unexpected type for \"volumes\" in {}".format(cls.__name__))
        # TODO Repeat above for "values"
        values = np.array(data.pop("values"))
        kind = data.pop("kind", "linear")
        return cls(model, node, volumes, values, interp_kwargs={'kind': kind})
```

You would then be able to specify the JSON like so:

```json
my_interp_volume: {
  "type": "InterpolatedVolume",
  "volumes": {
    "url": "mydata.csv",
    "column": "volumes"
  },
  "values": {
    "url": "mydata.csv",
    "column": "values"
  }
}     
```

A similar approach is used for the `ArrayIndexedScenarioMonthlyFactorsParameter`: https://github.com/pywr/pywr/blob/master/pywr/parameters/_parameters.pyx#L532

VJander:
Thanks James!
Just did the pull request, the different tests didn't check any errors, would you check them out to confirm this?

jetuk:
Closed in #930

PR #930, opened 2020-09-16
Add reading external data option to InterpolateVolumeParameter. (#926)

VJander:
The InterpolateVolumeParameter can now receive volumes and values from external data, and not only as a direct array input.


pywr/parameters/parameters.py
```diff
@@ -139,8 +139,20 @@ def _value_to_interpolate(self, ts, scenario_index):
     @classmethod
     def load(cls, model, data):
         node = model._get_node_from_ref(model, data.pop("node"))
-        volumes = np.array(data.pop("volumes"))
-        values = np.array(data.pop("values"))
+        volumes = data.pop("volumes")
+        if isinstance(volumes, list):
+            volumes = np.array("volumes")
```
jetuk, line 144: This should be:

```
volumes = np.array(volumes)
```

pywr/parameters/parameters.py
```diff
@@ -139,8 +139,20 @@ def _value_to_interpolate(self, ts, scenario_index):
...
+        elif isinstance(volumes, dict):
+            volumes = load_parameter_values(model, volumes)
+        else:
+            raise TypeError("Unexpected type for \"volumes\" in {}".format(cls.__name__))
+        values = data.pop("values")
+        if isinstance(values, list):
+            values = np.array("values")
```
jetuk, line 151: This should be:

```
values = np.array(values)
```

tests/models/reservoir_evaporation_areafromfile.json
```diff
@@ -0,0 +1,66 @@
+{
```
jetuk, line 1: Add a function to `tests/test_run.py` to load, run and check the output of this example.

tests/models/reservoirarea.csv
```diff
@@ -0,0 +1,3 @@
+volumes,values
```
jetuk, line 1: I think this is formatted incorrectly. Shouldn't it be?

```csv
volumes,values
0,0
200,100
400,200
600,300
800,400
1000,500
```

jetuk (changes requested):
Changes look good, but has broken the old array loading. I think this was my mistake in the comment in #926 . 

We should add a test function to run and verify the output of the example model. You can run the tests locally by using `pytest tests` from the root of your pywr folder.

tests/models/reservoirarea.csv
VJander: totally right!

tests/models/reservoir_evaporation_areafromfile.json
VJander: done!

pywr/parameters/parameters.py
VJander: fixed!

pywr/parameters/parameters.py
VJander: fixed !

VJander: however, the pytest tests still isn't able to read the .csv input as an array, as the problem in the test appears to be linked to the interpolation itself (it doesn't recieve an array, or perhaps not one it can read). 
Could you confirm this?

VJander:
however, the pytest tests still isn't able to read the .csv input as an array, as the problem in the test appears to be linked to the interpolation itself (it doesn't recieve an array, or perhaps not one it can read).
Could you confirm this?

jetuk:
@VJander I'm not sure I see a problem with this. Your test works fine? Do you still think there is an issue with the current code?

VJander:
@jetuk perhaps im running the test file in a wrong way, if it works fine for you i'm sure its ok then, because i didn't notice any errors on the code itself.

jetuk:
Can you paste the output of your error? `pytest tests` should be the command. It may be the development version hasn't installed / compiled correctly.

I am happy to merge this though.

jetuk (approved):

VJander:
James:
Thank you very much, you were right the version wasn't installed correctly!
Also thanks for the fast replies!

PR #932, opened 2020-09-18
Added NumpyArrayNodeCostRecorder

jetuk:


tests/test_recorders.py
```diff
@@ -162,11 +164,15 @@ def test_numpy_recorder(simple_linear_model):
...
     assert rec.data.shape == (365, 1)
+    assert cost_rec.data.shape == (365, 1)
     assert np.all((rec.data - 10.0) < 1e-12)
+    np.testing.assert_allclose(cost_rec.data, np.ones_like(cost_rec.data) * -2.0)
 
     df = rec.to_dataframe()
+    df2 = cost_rec.to_dataframe()
     assert df.shape == (365, 1)
     assert np.all((df.values - 10.0) < 1e-12)
+    np.testing.assert_allclose(df2.values, np.ones_like(cost_rec.data) * -2.0)
```
Batch21, line 175: Is it worth having a test where the node cost varies?

Batch21 (changes requested):
Looks good - surprised this doesn't exist already. Just a query on whether an additional test is necessary.

tests/test_recorders.py
jetuk: Done.

Batch21 (approved):
Tests look good, thanks!

PR #939, opened 2020-10-19
Add to_dataframe methods to two annual count recorders.

jetuk:
`AnnualTotalFlowRecorder` and `AnnualCountIndexThresholdRecorder` store their data internally in an annual timeseries. This commit exposes that data as a pandas DataFrame.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -1873,6 +1873,19 @@ cdef class AnnualCountIndexThresholdRecorder(Recorder):
         """
         return self._temporal_aggregator.aggregate_2d(self._data, axis=0, ignore_nan=self.ignore_nan)
 
+    def to_dataframe(self):
+        """ Return a `pandas.DataFrame` of the recorder data
+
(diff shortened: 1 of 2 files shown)


Batch21 (commented):
Changes look good. However, when a model also has a `NumpyArray*Recorder` and the `Model.to_dataframe()` method is called an error will result due to the different dataframe indexes. Might be worth adding a check in the method that returns a warning for this situation

jetuk:
> However, when a model also has a NumpyArray*Recorder and the Model.to_dataframe() method is called an error will result due to the different dataframe indexes

Hmmm. This is already possible with FDC recorders ...

Batch21:
Didn't think about the FDC recorders... tbh I only realised the model object had a `to_dataframe` method a couple of months ago. I haven't seen it used anywhere

jetuk:
Sounds like we should open an issue for that bug, and decided how to fix it. This won't make anything worse in the sense that it's possible to have that method fail already.

Batch21 (approved):

jetuk:
Tag #940

Issue #889, opened 2020-05-21
InterpolatedVolumeParameter documentation is confusing


Batch21:
The API documentation shows this:

> Generic interpolation parameter calculated from current volume
>
> __init__(self, model, node, x, y, interp_kwargs=None, \*\*kwargs)
> Initialize self. See help(type(self)) for accurate signature.

However if you defined the parameter in json you have to use `volumes` and `values` instead of `x` and `y` as the argument keys.

`InterpolatedFlowParameter` has the same issue

Easiest solution would be to add an `__init__` function with correct arg names that just supers the parent class

jetuk:
https://github.com/pywr/pywr/blob/e7b4e35c5eef08442e1f390c4a5b8270cb0ae926/pywr/parameters/parameters.py#L118

Just rename `x` and `y` to `volumes` and `values` respectively?

Same issue here: 

https://github.com/pywr/pywr/blob/e7b4e35c5eef08442e1f390c4a5b8270cb0ae926/pywr/parameters/parameters.py#L140

Batch21:
Not sure what I was looking at when I thought the classes didn't have __init__ funcs. I've update the arg names and docstrings in #890 

It looks like neither parameter is tested directly though they are used in other tests? 

Also it is not possible to set any of the `interp_kwargs` in the json other than the `kind`. Maybe the `loads` function should assume that anything left in data should be added to that input arg?

jetuk:
> Also it is not possible to set any of the interp_kwargs in the json other than the kind. Maybe the loads function should assume that anything left in data should be added to that input arg?

I think it would be better to have `interp_kwargs` as a key with a sub-dictionary in the JSON. Let the `__init__` method handle default arguments. This is how the schema branch does the pandas kwargs and it means top level of the JSON does not have open ended keywords.

Batch21:
That is a better way of doing it. Might have to account for the `kind` arg not being in the `interp_kwargs` sub-dictionary for backwards compatibility.

jetuk:
#890 solved the main issue here. Shall we keep this open for the `interp_kwargs` issue as well?

PR #943, opened 2020-11-03
update load methods of interpolation parameters so that they accept interp1d kwargs

Batch21:
This could be made a bit cleaner by changing the api and requiring the `kind` kwarg to be included with `interp_kwargs` rather than allowing it to still be defined separately. What do you think @jetuk 

See also #889

### DIFF
--- a/pywr/parameters/parameters.py
+++ b/pywr/parameters/parameters.py
@@ -73,6 +73,17 @@ class AbstractInterpolatedParameter(Parameter):
     def _value_to_interpolate(self, ts, scenario_index):
         raise NotImplementedError()
 
+    @property
+    def interp_kwargs(self):
+        return self._interp_kwargs
+
+    @interp_kwargs.setter
+    def interp_kwargs(self, data):
+        if "fill_value" in data and isinstance(data["fill_value"], list):
+            # SciPy's interp1d expects a tuple when defining fill values for the upper and lower bounds
+            data["fill_value"] = tuple(data["fill_value"])
+        self._interp_kwargs = data
+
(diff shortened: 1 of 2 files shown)


jetuk:
Here you've not removed `kind`? I agree we should probably break this (by removing kind) to make it neater with the other parameters. It means your code would then be cleaner as it could just pass through `interp_kwargs` directly to `__init__`.

pywr/parameters/parameters.py
```diff
@@ -109,7 +109,17 @@ def load(cls, model, data):
         x = np.array(data.pop("x"))
         y = np.array(data.pop("y"))
         kind = data.pop("kind", "linear")
-        return cls(model, parameter, x, y, interp_kwargs={'kind': kind})
+        interp_kwargs = data.pop("interp_kwargs", None)
+        if interp_kwargs:
+            interp_kwargs["kind"] = kind
```
jetuk, line 114: Could this be a staticmethod on the `InterpolatedParameter` base class?

pywr/parameters/parameters.py
```diff
@@ -109,7 +109,17 @@ def load(cls, model, data):
...
+            if "fill_value" in interp_kwargs and isinstance(interp_kwargs["fill_value"], list):
```
jetuk, line 115: Could this be a static method on the InterpolatedParameter base class?

Batch21: I've added a property for inter_kwargs on the base class

jetuk (approved):

Issue #944, opened 2020-11-18
Migrate to Github Actions


jetuk:
Travis' free tier has been removed:

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

We'll run out of our free credits soon if we haven't already. I suggest we move to Github Actions for both Linux and Windows - possibly re-add OSX. We can also start doing the pypi builds and deployment directly from the actions.

snorfalorpagus:
I'm :+1: on doing builds and deployments via the CI.

jetuk:
Maybe use this: https://github.com/joerick/cibuildwheel

jetuk:
The majority of this was done in #948.

Remaining todos:

- [ ] OSX builds
- [ ] Coverage build
- [ ] Pypi upload

jetuk:
Pypi upload is now done.

jetuk:
Closing this as complete.

PR #948, opened 2020-11-22
Initial commit of Github Action using cibuildhweel.

jetuk:
Work in progress to address #944 

Todo:

- [x] Linux builds
- [x] Windows builds
- [ ] OSX builds
- [x] Documentation build
- [ ] Coverage build
- [x] sdist
- [ ] Pypi upload
- [x] Test all solvers on all platforms
- [x] Disable Travis
- [x] Disable Appveyor


jetuk:
This now works on Linux (but not 32-bit), but for some reason it's not getting the package version correctly.

Windows builds will require porting over the glpk installation script (glpk is not in Chocolatey) from Appveyor.

Not tried to fix OSX yet, but `brew install glpk` should be used.

jetuk:
@Batch21 @snorfalorpagus what do you think to this? I've got Windows and Linux working. However, they are currently only testing the main GLPK solver.

jetuk:
This now does tests with all the solvers.

jetuk:
@Batch21 this is now deploying the docs to: http://pywr.github.io/pywr/

This would make the pywr-docs repo redundant (not a bad thing). The CSS is currently not working for me, but that might just be delay in the loading / cache thing. At the moment this would mean we just have master docs (not the version docs in sub-folders as before), but I think that's OK at the moment.

If you're happy with the rest I propose we merge this now and use the wheels it builds going forward (i.e. for 1.10.0).

The last it is to uncomment the line to make the doc deploy only happen on master.

jetuk:
Actually there will be some housekeeping things to do in the README - updating doc links and badges. I'll do those later.

jetuk:
> Actually there will be some housekeeping things to do in the README - updating doc links and badges. I'll do those later.

I've done this now.

Batch21 (approved):
Had a quick look through but didn't go into great depth on how Github Actions work. It looks good though. Only comment is that it felt like you have to dig a little into the outputs to find where the tests have been run but that might be unfamiliarity with it.

PR #950, opened 2020-12-02
Add optional exclude_months to AnnualCountIndexThreshRecorder

jetuk:
This new option means the counts are ignored for the excluded months if they are defined. This allows counting only a sub-set of months during a year.


Batch21 (approved):
Looks good

PR #958, opened 2021-01-04
Add `get_all_xxx` methods to core nodes.

jetuk:
Add methods to core nodes to get values for all scenarios. This is more efficient than call the `get_xxx` methods many times.

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -477,6 +477,14 @@ cdef class AbstractNode:
     cpdef double get_cost(self, ScenarioIndex scenario_index) except? -1:
         return 0.0
 
+    cpdef double[:] get_all_cost(self, double[:] out=None):
+        if out is None:
+            out = np.zeros(len(self.model.scenarios.combinations))
+        else:
+            out[:] = 0.0
+        return out
+
+
 cdef class Node(AbstractNode):
     """ Node class from which all others inherit
     """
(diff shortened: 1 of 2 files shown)


pywr/_core.pyx
```diff
@@ -477,6 +477,14 @@ cdef class AbstractNode:
     cpdef double get_cost(self, ScenarioIndex scenario_index) except? -1:
         return 0.0
 
+    cpdef double[:] get_all_cost(self, double[:] out=None):
```
Batch21, line 480: Is there a specific reason for having the `out` arg? and does there need to be any check on its shape?

pywr/_core.pyx
```diff
@@ -800,19 +858,54 @@ cdef class StorageInput(BaseInput):
         BaseInput.commit(self, scenario_index, volume)
         self._parent.commit(scenario_index, -volume)
 
+    cpdef commit_all(self, double[:] value):
```
Batch21, line 861: I'm guessing these method are just to allow net flow for a storage node to be easily recorded?

Batch21 (commented):
Changes look good. Are the pipeline failures all related to xlrd?

pywr/_core.pyx
jetuk: It allows the solver to set net flows for all scenarios simultaneously. A `commit_all` method already exists on `AbstractNode`, but these storage nodes need different behaviour to update the parent storage node.

pywr/_core.pyx
jetuk: It allows for a pre-allocated array to be given in which the values are set (i.e. avoids the need to reallocate space on every call). A solver would allocate space once per run to accept all data, and then pass those arrays as the `out` argument. This avoids an additional allocation and possibly a copy of the data.

jetuk:
> Changes look good. Are the pipeline failures all related to xlrd?

I think so.

jetuk:
This is now rebased on master so the pipeline should work.

Batch21 (approved):

PR #959, opened 2021-01-04
Switch to openpyxl instead of xlrd and drop Python 3.6 support.

jetuk:
Pandas 1.2 has been released and no longer supports reading xlsx files with `xlrd`. Openpyxl is the preferred default.


Batch21:
How come some of the pipeline jobs are still trying to install xlrd?

jetuk:
The issue here is that pandas has dropped 3.6 support with the 1.2 release. The 1.2 release is needed for it to work with openpyxl and not xlrd. I suggest we also drop 3.6 support. Pandas decision ultimately goes back to numpy which dropped support in June 2020:

https://numpy.org/neps/nep-0029-deprecation_policy.html

jetuk:
Last commit drops 3.6 from CI, which should make it work.

Batch21 (approved):

PR #960, opened 2021-01-06
Add LossLink node.

jetuk:


pywr/nodes.py
```diff
@@ -1074,4 +1074,96 @@ def after(self, timestep):
...
+    A fixed proportional loss of all flow through this node is sent to an internal output node. Max and min flows
+    applied to this node are enforced on the net output after losses. The node itself records the net output
+    in its flow attribute (which would be used by any attached recorders).
+
+    Parameters
+    ----------
+    model : `pywr.model.Model`
+    name : string
+        Name of the node.
+    loss_factor : float
```
jetuk, line 1089: This probably should have a way of specifying the loss_factor based on gross inflow or net outflow, and constraints on gross inflow and/or net outflow.

tests/test_run.py
```diff
@@ -672,6 +672,24 @@ def test_breaklink_node():
     assert_allclose(transfer.storage.volume, 0)
 
 
+def test_loss_link_node():
```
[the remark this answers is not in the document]
Batch21: and 1.0

pywr/nodes.py
Batch21: any plans to still do this?

jetuk: I'm not sure it's too tricky? What do you think?

Batch21: probably worth doing if its not too tricky. Does it just involve changing which component node commits flow to the parent node?

jetuk: I think you have to change which nodes are in the internal `AggregatedNode`. Or perhaps just the calculation of the factor.

jetuk: Let's leave it be for now.

Batch21 (approved):

PR #968, opened 2021-03-02
Add an optional tags dictionary to Component

jetuk:
Parameters and Recorders can now optionally take a dict of tags. The component just stores these in a public attribute. It will
be useful for various tools to provide and save metadata about components. E.g. for filtering and grouping.

This PR doesn't use this metadata anywhere. It should be possible to add tags to the `TablesRecorder` parameter arrays as metadata.


Batch21 (approved):
Looks good

PR #969, opened 2021-03-08
MultiThresholdIndexParameter

laurenpetch:
MultiThresholdIndexParameter to work similarly to a ControlCurveIndexParameter except with Nodes and flows

### DIFF
--- a/pywr/parameters/_thresholds.pyx
+++ b/pywr/parameters/_thresholds.pyx
@@ -232,6 +232,53 @@ cdef class MultipleThresholdIndexParameter(IndexParameter):
 MultipleThresholdIndexParameter.register()
 
 
+cdef class MultipleThresholdParameterIndexParameter(IndexParameter):
+    """ Multiple threshold holder returning an index based on the value in the parameter
+    against the given thresholds.
+
+    Parameters
+    ----------
+    parameter : Parameter
+    thresholds : iterable of `Parameter` instances or floats
+    """
+    def __init__(self, model, parameter, thresholds, use_max_flow=False, **kwargs):
+        super(MultipleThresholdParameterIndexParameter, self).__init__(model, **kwargs)
+        self.parameter = parameter
+        self.thresholds = thresholds
+
+    cpdef int index(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
+        """Returns the index of the first threshold the node flow is above
+
+        The index is zero-based. For example, if only one threshold is
+        supplied then the index is either 0 (above) or 1 (below). For two
+        thresholds the index is either 0 (above both), 1 (in between), or 2 (below
+        both), and so on.
+        """
+        cdef double flow
+        cdef int index, j
+
+        flow = self.parameter.value(timestep, scenario_index)
+        index = len(self.thresholds)
+        for j, threshold in enumerate(self.thresholds):
+            if isinstance(threshold, (float, int)):
+                target_threshold = threshold
+            else:
+                target_threshold = threshold.get_value(scenario_index)
+            if flow >= target_threshold:
+                index = j
+                break
+        return index
+
+    @classmethod
+    def load(cls, model, data):
+        node = load_parameter(model, data.pop("parameter"))
+        thresholds = [load_parameter(model, d) for d in data.pop("thresholds")]
+        return cls(model, node, thresholds, **data)
+MultipleThresholdParameterIndexParameter.register()
+
+
+
+
 cdef class ParameterThresholdParameter(AbstractThresholdParameter):
     """ Returns one of two values depending on the value of a Parameter
 
--- a/pywr/parameters/_thresholds.pyx
+++ b/pywr/parameters/_thresholds.pyx
@@ -187,6 +187,51 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
 NodeThresholdParameter.register()
 
 
+cdef class MultipleThresholdIndexParameter(IndexParameter):
+    """ Multiple threshold holder returning an index based on the previous flow in a node
+    against the given thresholds.
+
+    Parameters
+    ----------
+    node : `pywr.core.AbstractNode`
+    thresholds : iterable of `Parameter` instances or floats
+    """
+    def __init__(self, model, node, thresholds, use_max_flow=False, **kwargs):
+        super(MultipleThresholdIndexParameter, self).__init__(model, **kwargs)
+        self.node = node
+        self.thresholds = thresholds
+
+    cpdef int index(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
+        """Returns the index of the first threshold the node flow is above
+
+        The index is zero-based. For example, if only one threshold is
+        supplied then the index is either 0 (above) or 1 (below). For two
+        thresholds the index is either 0 (above both), 1 (in between), or 2 (below
+        both), and so on.
+        """
+        cdef double flow
+        cdef int index, j
+
+        flow = self.node._prev_flow[scenario_index.global_id]
+        index = len(self.thresholds)
+        for j, threshold in enumerate(self.thresholds):
+            if isinstance(threshold, (float, int)):
+                target_threshold = threshold
+            else:
+                target_threshold = threshold.get_value(scenario_index)
+            if flow >= target_threshold:
+                index = j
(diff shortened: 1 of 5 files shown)


pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+
+    Parameters
+    ----------
+    node : `pywr.core.AbstractNode`
+    thresholds : iterable of `Parameter` instances or floats
+    """
+    def __init__(self, model, node, thresholds, use_max_flow=False, **kwargs):
+        super(MultipleThresholdIndexParameter, self).__init__(model, **kwargs)
+        self.node = node
+        self.thresholds = thresholds
```
jetuk, line 202: If any of these are `Parameter` objects they will need adding as children to this object. Otherwise their calculation order is not guaranteed.

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+        thresholds the index is either 0 (above both), 1 (in between), or 2 (below
+        both), and so on.
+        """
+        cdef double flow
+        cdef int index, j
+
+        flow = self.node._prev_flow[scenario_index.global_id]
+        index = len(self.thresholds)
+        for j, threshold in enumerate(self.thresholds):
+            if isinstance(threshold, (float, int)):
```
jetuk, line 218: I think the control curve equivalent of this turns floats and ints into a `ConstantParameter` on load/creation. At the moment this loop won't Cythonise too well because `threshold` can not be cdef'd. If you knew they were all `Parameter` instances then it could be `cdef Parameter` and probably be quicker.

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+    cpdef int index(self, Timestep timestep, ScenarioIndex scenario_index) except? -1:
+        """Returns the index of the first threshold the node flow is above
+
+        The index is zero-based. For example, if only one threshold is
+        supplied then the index is either 0 (above) or 1 (below). For two
+        thresholds the index is either 0 (above both), 1 (in between), or 2 (below
+        both), and so on.
+        """
+        cdef double flow
+        cdef int index, j
```
jetuk, line 213: Add `cdef double target_threshold`

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+
+        flow = self.parameter.value(timestep, scenario_index)
+        index = len(self.thresholds)
+        for j, threshold in enumerate(self.thresholds):
+            if isinstance(threshold, (float, int)):
```
jetuk, line 263: As above.

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+
+        The index is zero-based. For example, if only one threshold is
+        supplied then the index is either 0 (above) or 1 (below). For two
+        thresholds the index is either 0 (above both), 1 (in between), or 2 (below
+        both), and so on.
+        """
+        cdef double flow
+        cdef int index, j
+
+        flow = self.parameter.value(timestep, scenario_index)
```
jetuk, line 260: This is a generic value now, not a "flow".

It should also be `.get_value(scenario_index)`

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+            else:
+                target_threshold = threshold.get_value(scenario_index)
+            if flow >= target_threshold:
+                index = j
+                break
+        return index
+
+    @classmethod
+    def load(cls, model, data):
+        node = load_parameter(model, data.pop("parameter"))
```
jetuk, line 274: Not a "node"

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
 NodeThresholdParameter.register()
 
 
+cdef class MultipleThresholdIndexParameter(IndexParameter):
+    """ Multiple threshold holder returning an index based on the previous flow in a node
```
jetuk, line 191: "threshold holder" - typo?

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+    @classmethod
+    def load(cls, model, data):
+        node = model.nodes[data.pop("node")]
+        thresholds = [load_parameter(model, d) for d in data.pop("thresholds")]
+        return cls(model, node, thresholds, **data)
+MultipleThresholdIndexParameter.register()
+
+
+cdef class MultipleThresholdParameterIndexParameter(IndexParameter):
+    """ Multiple threshold holder returning an index based on the value in the parameter
```
jetuk, line 236: "threshold holder" - typo?

jetuk (changes requested):
A couple of comments here. 

Is it possible to make a shared based class? A bit like the one for the regular threshold parameters?

pywr/parameters/_thresholds.pyx
laurenpetch: when i put this to `get_value(scenario_index)` it returns the value from the previous day i think whereas this gives the current day which is what i was after

pywr/parameters/_thresholds.pyx
```diff
@@ -187,6 +187,98 @@ cdef class NodeThresholdParameter(AbstractThresholdParameter):
...
+    against the given thresholds.
+
+    Parameters
+    ----------
+    parameter : Parameter
+    thresholds : iterable of `Parameter` instances or floats
+    """
+    def __init__(self, model, parameter, thresholds, use_max_flow=False, **kwargs):
+        super(MultipleThresholdParameterIndexParameter, self).__init__(model, **kwargs)
+        self.parameter = parameter
```
jetuk, line 246: `self.children.add(parameter)`

pywr/parameters/_thresholds.pyx
jetuk: This is most likely due to `self.parameter` not being registered as a child of self. See comment I just added. It means that `self.parameter` is not evaluated before `self`, and hence giving you the old value.

pywr/parameters/_thresholds.pyx
laurenpetch: I've changed this to load a `ConstantParameter` if not already a parameter and then add children. I took this from the `IndexedArrayParameter` so might not be exactly how the control curve equivalent does it?

pywr/parameters/_thresholds.pyx
laurenpetch: changed

pywr/parameters/_thresholds.pyx
laurenpetch: added

pywr/parameters/_thresholds.pyx
laurenpetch: see comment above

pywr/parameters/_thresholds.pyx
laurenpetch: changed

pywr/parameters/_thresholds.pyx
laurenpetch: changed

jetuk:
My only remaining comment is to add these two to the API documentation.

laurenpetch:
@jetuk thanks - think i've added them to the API documentation, let me know if there's anything else to do for this

laurenpetch:
@jetuk is this OK to merge now?

jetuk:
> @jetuk is this OK to merge now?

Yes. I just made a couple of minor edits to the docstrings. Will merge when tests pass.

jetuk (approved):

PR #971, opened 2021-03-16
Add NormalisedGaussianKDEStorageRecorder

jetuk:
Works in a similar way to `GaussianKDEStorageRecorder`, but derives from a new array recorder. `NormalisedGaussianKDEStorageRecorder` keeps an array of storage that is normalised between -1 and 1 where the 0 value is at a control curve. The new KDE recorder estimates a probability of being at or below zero in the normalised series.

### DIFF
--- a/pywr/recorders/_recorders.pyx
+++ b/pywr/recorders/_recorders.pyx
@@ -1126,6 +1126,55 @@ cdef class NumpyArrayStorageRecorder(NumpyArrayAbstractStorageRecorder):
 NumpyArrayStorageRecorder.register()
 
 
+cdef class NumpyArrayNormalisedStorageRecorder(NumpyArrayAbstractStorageRecorder):
+    """Recorder for timeseries information normalised relative to a control curve for a `Storage` node.
+
+    This class stores normalised storage from a specific node for each time-step of a simulation. The normalisation
+    is relative to a `Parameter` which defines a control curve. The data is normalised such that values of 1, 0 and
+    -1 align with full, at control curve and empty volumes respectively. The data is saved internally using a
+     memory view. The data can be accessed through the `data` attribute or `to_dataframe()` method.
+
+    Parameters
+    ----------
+    model : `pywr.core.Model`
+    node : `pywr.core.Node`
+        Node instance to record.
+    parameter : `Parameter`
+        The control curve for normalisation.
+    temporal_agg_func : str or callable (default="mean")
+        Aggregation function used over time when computing a value per scenario. This can be used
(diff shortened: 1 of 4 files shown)


pywr/recorders/_recorders.pyx
```diff
@@ -1126,6 +1126,55 @@ cdef class NumpyArrayStorageRecorder(NumpyArrayAbstractStorageRecorder):
...
+    Parameters
+    ----------
+    model : `pywr.core.Model`
+    node : `pywr.core.Node`
+        Node instance to record.
+    parameter : `Parameter`
+        The control curve for normalisation.
+    temporal_agg_func : str or callable (default="mean")
+        Aggregation function used over time when computing a value per scenario. This can be used
+        to return, for example, the median flow over a simulation. For aggregation over scenarios
```
matrosoe, line 1146: This can be used to return, for example, the minimum normalised storage reached over a simulation. For aggregation over scenarios  see the `agg_func` keyword argument.

matrosoe (approved):

Batch21 (commented):
Implementation looks good. Only comment is that there looks like there is quite a bit of repeated code in the two kde recorders that could possibly be refactored into an abstract class.

jetuk:
Yes, there is, but I am happy with that at the moment. I think it's a bit more readable as is.

pywr/recorders/_recorders.pyx
```diff
@@ -1126,6 +1126,55 @@ cdef class NumpyArrayStorageRecorder(NumpyArrayAbstractStorageRecorder):
...
+    parameter : `Parameter`
+        The control curve for normalisation.
+    temporal_agg_func : str or callable (default="mean")
+        Aggregation function used over time when computing a value per scenario. This can be used to return, for
+        example, the minimum normalised storage reached over a simulation. For aggregation over scenarios see the
+        `agg_func` keyword argument.
+    """
+    def __init__(self, *args, **kwargs):
+        # Optional different method for aggregating across time.
+        self.parameter = kwargs.pop('parameter')
```
matrosoe, line 1151: @jetuk Should there be a load method in the class to load the parameter with the load_parameter method? At the moment passing a parameter in the json results in 'parameter' being type str.

jetuk: Yes. I have fixed here: https://github.com/pywr/pywr/pull/974

I would open an issue in future. This comment on a closed MR might have got missed.

matrosoe: Yes, thanks. Will open an issue next time

PR #976, opened 2021-04-12
register event recorders and add load methods for them

Batch21:


tests/test_recorders.py
```diff
@@ -1882,6 +1882,16 @@ class TestEventRecorder:
...
 
+    def test_load(self, cyclical_storage_model):
+        """Test load method""" 
+        m = cyclical_storage_model
+        strg = m.nodes['Storage']
+        param = StorageThresholdRecorder(m, strg, 4.0, predicate='<=', name="trigger")
+        EventRecorder.load(m, {"name": "event_rec", "threshold": "trigger", "tracked_parameter":"inpt_flow"})
+        EventDurationRecorder.load(m, {"event_recorder": "event_rec"})
+        EventStatisticRecorder.load(m, {"event_recorder": "event_rec"})
+        m.step()
```
jetuk, line 1893: Change this to `m.run()`. I have a feeling something bad happens if you don't specify `recorder_agg_func` or `agg_func`, but it won't happen until the `finish` method.

pywr/recorders/events.py
```diff
@@ -178,6 +178,22 @@ def to_dataframe(self):
 
         return pandas.DataFrame(df_dict)
 
+    @classmethod
+    def load(cls, model, data):
+        from pywr.parameters import load_parameter
+        threshold = data.pop("threshold")
+        try:
+            threshold = load_parameter(model, threshold)
+        except KeyError:
```
jetuk, line 187: I guess we don't test this.

jetuk (changes requested):
Just check that finish method works by changing the test to a full model run.

tests/test_recorders.py
```diff
@@ -1882,6 +1882,17 @@ class TestEventRecorder:
...
     funcs = {"min": np.min, "max": np.max, "mean": np.mean, "median": np.median, "sum": np.sum}
 
+    @pytest.mark.parametrize("threshold_component", [StorageThresholdRecorder, StorageThresholdParameter])
+    def test_load(self, cyclical_storage_model, threshold_component):
+        """Test load method""" 
+        m = cyclical_storage_model
+        strg = m.nodes['Storage']
+        param = threshold_component(m, strg, 4.0, predicate='<=', name="trigger")
+        EventRecorder.load(m, {"name": "event_rec", "threshold": "trigger", "tracked_parameter": "inpt_flow"})
+        EventDurationRecorder.load(m, {"event_recorder": "event_rec", "recorder_agg_func": "max"})
```
jetuk, line 1892: So, what happens  if you don't provide `recorder_agg_func` or `agg_func`? Does it error? Should we have a default here?

Batch21: yeah, it errors. We probably should have a default. Any preferences on what this should be?

jetuk: I think "mean" is the default for "agg_func" in general, so can likely be the same here. Make sure the docstring says the default value too.

Batch21: Done

jetuk (approved):

PR #977, opened 2021-04-12
load fdc deviation targets from file

Batch21:


pywr/recorders/_recorders.pyx
```diff
@@ -1042,6 +1054,19 @@ cdef class FlowDurationCurveDeviationRecorder(FlowDurationCurveRecorder):
         else:
             return df
 
+    @classmethod
+    def load(cls, model, data):
+        node = model.nodes[data.pop("node")]
+        percentiles = data.pop("percentiles")
+        from pywr.parameters import load_parameter_values
+        upper_target_fdc = data.pop("upper_target_fdc", None)
```
jetuk, line 1062: Should these use the `percentiles` to get the values from the dataframe? E.g. if you specified the percentiles in a different order, or had more values defined in the dataframe?

jetuk: Something like,

```python
upper_target_fdc = upper_target_fdc.loc[percentiles, :]
```

tests/test_recorders.py
```diff
@@ -382,6 +382,60 @@ def test_fdc_dev_recorder(self, agg_func, aggregate):
         df = rec.to_dataframe()
         assert df.shape == (len(percentiles), len(model.scenarios.combinations))
 
+    def test_deviation_single_target_lower(self):
```
jetuk, line 385: Docstring

tests/test_recorders.py
```diff
@@ -382,6 +382,60 @@ def test_fdc_dev_recorder(self, agg_func, aggregate):
...
+
+        model.run()
+
+        actual_fdc = np.maximum(natural_fdc - 23, 0.0)
+
+        lower_deviation = (lower_input_fdc - actual_fdc) / lower_input_fdc
+        deviation = np.maximum(lower_deviation, np.zeros_like(lower_deviation))
+        assert_allclose(rec.fdc_deviations[:, 0], deviation[:, 0])
+
+    def test_deviation_single_target_upper(self):
```
jetuk, line 412: Docstring

jetuk (approved):
We should check make sure there's no way the wrong percentiles can be used from an external file.

pywr/recorders/_recorders.pyx
Batch21: Done but as `load_parameter_values` returns a numpy array I had to make a small update to the dataframe functions

jetuk: OK, are you happy with this now basically requiring the input data to be indexed?

Batch21: yes, I think its a good check and not to onerous for anyone to do. I added comments to the docstring stating that this is necessary.

jetuk:
Could we just add the docstrings for the tests? Then I will merge.

jetuk (approved):

PR #981, opened 2021-04-30
Make DF parameter only load required data.

jetuk:
TODO:

- [x] Tests


Batch21 (approved):
look goods. Always get a bit confused by the scenario indices code but it works!

PR #985, opened 2021-05-15
Update constant aggregated node factors only during rest.

jetuk:
This is a performance improvement that follows on from #983
by only updating aggregated node factors in the GLPK solvers
during reset instead of every timestep. It should have most
benefit for models which use many of aggregated nodes with factors
as the code both for calculating the normalised factors is
not very efficient.

A new solver option and corresponding environment variable is
added to turn this feature on. It is off by default.

Modified the CI test suite to run only with and without the
extra options. As opposed to running each option individually.

### DIFF
--- a/.github/workflows/run-tests.sh
+++ b/.github/workflows/run-tests.sh
@@ -3,16 +3,12 @@
 # GLPK solver
 echo "Running GLPK (route-based) tests in: $1";
 PYWR_SOLVER="glpk" pytest "$1"/tests  || exit 1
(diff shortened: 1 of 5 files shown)


Batch21 (changes requested):
Looks good. Gave the same results when testing on a complex model and seemed to give a good speed bump (~10% faster) compared to the latest commit on master.

couple of things:
- It looks the the path solver is still looping through all the agg nodes in `_solve_scenario`. Line 803 needs updating to use `self.aggregated_with_dynamic_factors`
- it would be good to add the new solver setting to the model results table.

jetuk:
Addressed both the comments.

Batch21 (approved):

PR #986, opened 2021-05-19
updated MonthlyProfileParameter and RbfProfileParameter to take list …

tom-gribbin:
…of values for upper and lower bounds, instead of only a constant across the year

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -626,14 +626,12 @@ cdef class MonthlyProfileParameter(Parameter):
     values : iterable, array
         The 12 values that represent the monthly profile.
     lower_bounds : float or array_like (default=0.0)
-        Defines the lower bounds when using optimisation.
-        If a float given, same bound applied for every month.
-        Otherwise an array like object of length 12 should be given for as separate value each month.
+        Defines the lower bounds when using optimisation. If a float given, same bound applied for every
+        month. Otherwise an array like object of length 12 should be given for as separate value each month.
     upper_bounds : float or array_like (default=np.inf)
-        Defines the upper bounds when using optimisation.
-        If a float given, same bound applied for every month.
-        Otherwise an array like object of length 12 should be given for as separate value each month.
-    inter_day : str or None (default=None)
+        Defines the upper bounds when using optimisation. If a float given, same bound applied for every
+        month. Otherwise an array like object of length 12 should be given for as separate value each month.
+    interp_day : str or None (default=None)
         If `interp_day` is None then no interpolation is undertaken, and the parameter
          returns values representing a piecewise monthly profile. Otherwise `interp_day`
          must be a string of either "first" or "last" representing which day of the month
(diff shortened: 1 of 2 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -625,10 +625,14 @@ cdef class MonthlyProfileParameter(Parameter):
     ----------
     values : iterable, array
         The 12 values that represent the monthly profile.
-    lower_bounds : float (default=0.0)
+    lower_bounds : float (default=0.0) or list
```
jetuk, line 628: I did a bit more reading about this. These should accept general `array_like` rather than just lists.

```
lower_bounds, upper_bounds : float or array_like (default=0.0)
    Defines the lower and upper bounds when using optimisation. If a float given, same bound applied for every
    month. Otherwise an array like object of length 12 should be given for as separate value each month.
```

pywr/parameters/_parameters.pyx
```diff
@@ -650,8 +654,21 @@ cdef class MonthlyProfileParameter(Parameter):
             raise ValueError("12 values must be given for a monthly profile.")
         self._values = np.array(values)
         self.interp_day = interp_day
-        self._lower_bounds = np.ones(self.double_size)*lower_bounds
-        self._upper_bounds = np.ones(self.double_size)*upper_bounds
+
+        if isinstance(lower_bounds, list):
```
jetuk, line 658: We can then change this to:

```
if np.isscalar(lower_bounds):
    lb = np.ones(self.double_size) * lower_bounds
else:
    lb = np.array(lower_bounds)
    if len(lb) != self.double_size:
        raise ValueError("Lower bounds must be a scalar or array like of 12 values.")
self._lower_bounds = lb

File: pywr/parameters/_parameters.pyx [hunk 3]
```diff
@@ -650,8 +654,21 @@ cdef class MonthlyProfileParameter(Parameter):
...
+            if len(lower_bounds) != self.double_size:
+                raise ValueError("if lower bound is list, 12 values must be given for a monthly profile.")
+            self._lower_bounds = np.array(lower_bounds)
+        else:
+            self._lower_bounds = np.ones(self.double_size)*lower_bounds
+
+        if isinstance(upper_bounds, list):
```
jetuk, line 665: See above comment.

pywr/parameters/_parameters.pyx
```diff
@@ -916,10 +933,14 @@ cdef class RbfProfileParameter(Parameter):
         value should be one.
     values : iterable, float
         Values to use for interpolation corresponding to the `days_of_year`.
-    lower_bounds : float (default=0.0)
+    lower_bounds : float (default=0.0) or list
```
jetuk, line 936: Follow suggestions from monthly profile.

jetuk (changes requested):
Thanks for these updates. Would be good to write a simple test to make sure these work as expected. I can probably do that if you can make the changes.

jetuk (approved):

jetuk:
I made a few minor tweaks and added some tests. Thanks for the contribution.

PR #993, opened 2021-07-20
Fix divide by zero error in two recorders.

### LINKED ISSUE #991: NumpyArrayNodeSuppliedRatioRecorder gives a divide by zero error
I think this relates to the line below (line 727 in _recorders.pyx:
self._data[ts.index,scenario_index.global_id] = node._flow[scenario_index.global_id] / max_flow

There is no check on whether max_flow is zero before the operation.

jetuk:
`NumpyArrayNodeSuppliedRatioRecorder` and `NumpyArrayNodeCurtailmentRatioRecorder`
could both error with a divide by zero if the max flow of the parameter
they monitor returned zero.

Fixes #991


matrosoe (commented):
Looks good!

matrosoe (approved):
Looks good!

Issue #992, opened 2021-07-20
 AnnualTotalFlowRecorder after method does not multiply each time step's flow by the number of days in the timestep and therefore undercountes the total accumulated annual flow when time steps of more than a single day in length are used


matrosoe:
Proposed fix is in this branch

https://github.com/pywr/pywr/tree/workingphase2

jetuk:
Agree this is a bug. We need a test for it, and would be a good idea to check the other recorders.

jetuk:
Fixed in #994

PR #994, opened 2021-08-02
Add methods to Timestep to return days in current and next year

Batch21:
These methods can be used by annual recorder to make sure that the correct proportion of flow is allocated to correct year when the timestep delta is greater than 1 day

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -36,6 +36,7 @@ cdef class Timestep:
     cdef readonly int day
     cdef readonly int month
     cdef readonly int year
+    cdef readonly int end_year
     cpdef public double days_in_current_year(self)
     cpdef public double days_in_next_year(self)
 
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -37,8 +37,8 @@ cdef class Timestep:
     cdef readonly int month
     cdef readonly int year
     cdef readonly int end_year
-    cpdef public int days_in_current_year(self)
-    cpdef public int days_in_next_year(self)
+    cpdef public double days_in_current_year(self)
+    cpdef public double days_in_next_year(self)
 
 cdef class Domain:
     cdef object name
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -37,8 +37,8 @@ cdef class Timestep:
     cdef readonly int month
     cdef readonly int year
     cdef readonly int end_year
-    cpdef public double days_in_current_year(self)
-    cpdef public double days_in_next_year(self)
+    cpdef public int days_in_current_year(self)
+    cpdef public int days_in_next_year(self)
 
 cdef class Domain:
     cdef object name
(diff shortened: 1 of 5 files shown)


jetuk:
Is this part of the plan to resolve #992 ?

tests/test_recorders.py
```diff
@@ -1677,6 +1677,26 @@ def test_annual_total_flow_recorder_factored(self, simple_linear_model):
         df = rec_fact.to_dataframe()
         assert_allclose([[3650.0*3]], df.values)
 
+    @pytest.mark.parametrize("end_date, expected", [("2013-01-04", [30.0, 40.0]), ("2012-12-31", [30.0]),])
+    def test_annual_total_flow_recorder_year_end(self, simple_linear_model, end_date, expected):
+        """
+        Test AnnualTotalFlowRecorder when timestep crosses year end
+        """
```
jetuk, line 1684: Can we add some longer explanation of the parametrized data here? I think I understand it, but would be best if it were written down that these two tests are each running 1 time-step but the first one overlaps into 2013.

jetuk (changes requested):
A couple of things to tidy up, but otherwise LGTM.

pywr/_core.pxd
```diff
@@ -36,6 +36,9 @@ cdef class Timestep:
     cdef readonly int day
     cdef readonly int month
     cdef readonly int year
+    cdef readonly int end_year
+    cpdef public double days_in_current_year(self)
```
jetuk, line 40: These should return `int`.

tests/test_core.py
```diff
@@ -564,3 +564,27 @@ def test_timestep_greater_than_zero_days():
...
+    simple_linear_model.step()
+
+    ts = simple_linear_model.timestepper.current
+
+    assert days_in_current_year == ts.days_in_current_year()
+    assert days_in_next_year == ts.days_in_next_year()
+
+    simple_linear_model.step()
+
+    #assert ts.days() == ts.days_in_current_year()
```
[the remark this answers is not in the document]
Batch21: done

pywr/_core.pxd
Batch21: done

tests/test_recorders.py
Batch21: done

jetuk:
What did we agree on this in the end? The current methods return an `int`, which is different to the `.days` attribute (a `double`), but the calculations are only accurate to the nearest day because they are based on days of the year.

Batch21:
> What did we agree on this in the end? The current methods return an `int`, which is different to the `.days` attribute (a `double`), but the calculations are only accurate to the nearest day because they are based on days of the year.

I've reverted the method return types back to `double` and made the calculations more accurate. What do you think?

jetuk:
Looks good to me. I guess the only issue might be the performance of the string parsing & creating the Timestamp object, but it's probably good for now.

jetuk (approved):

PR #997, opened 2021-09-03
Fixed remove methods in aggregated parameters

s-simoncelli:
Fixed attribute name in remove method of AggregatedParameter and AggregatedIndexParameter classes


jetuk:
Thanks for the fix.

jetuk (approved):

s-simoncelli:
No problem at all. Thanks for the quick merge!

Issue #1001, opened 2021-10-01
Aggregated and Piecewiselink nodes not constrained by licences


e-lancaster:
Storage recorders on licences show negative values.

Not working for AnnualVirtualStorage or RollingAnnualVirtualStorage (custom).[test2 (1).zip](https://github.com/pywr/pywr/files/7267471/test2.1.zip)

e-lancaster:
@jetuk @Batch21

jetuk:
@e-lancaster if you are able you can try the changes in #1002. This will not fix the reported issue, but should result in an error instead.

jetuk:
Version 1.15.3 should now error on this type of model configuration.

jetuk:
I've been thinking a long-term solution to this problem. I think the fundamental problems is that the nodes are added to the networkx graph that aren't really part of the network. Rather those nodes contain internal nodes which *are* part of the network. `PiecewiseLink`, for example, inherits `Node` which automatically adds it to the networkx graph. I propose that these container or isolated nodes, including `AggregatedNode` and `VirtualStorage` should not inherit from `Node` and therefore not be added to the networkx graph. 

The consequence of this will be that the solver code will no longer find these nodes. When a licence (or other node) references one of these container nodes it will need to informed which of the internal nodes to use for the constraint. Perhaps this looks like a `VirtualNode` abstract class that provides a method to iterate through the internal `Node` instances. Ideally it would handle the nested case where a licence is applied to an aggregated node which itself aggregates over two piecewise links.

jetuk:
@Batch21 thoughts on the above?

PR #1002, opened 2021-10-03
Assert entries when creating rows in the GLPK matrix.

jetuk:
In response to #1001 this commit implements a check to the `glp_set_mat_row` call to ensure that there is at least one entry in the row. Zero entry rows are likely caused by invalid or unsupported model / network configuration (e.g. those reported in #1001). For the time being this change will ensure such models fail to run with an `AssertionError` instead of failing to apply expected constraints silently.

I suggest we merge this rapidly and release a bug fix ASAP.

A proper fix to enable support for a `PiecewiseLink` or other complex nodes with `VirtualStorage` is still required.

### DIFF
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -256,6 +256,9 @@ cdef set_col_bnds(glp_prob *P, int i, int type, double lb, double ub):
 
 
 cdef set_mat_row(glp_prob *P, int i, int len, int* ind, double* val):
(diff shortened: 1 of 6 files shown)


Batch21 (approved):
changes look good. Tested it on a couple of models and it raises an assertion when it should

jetuk:
@Batch21 I've added another test for `AggregatedNode` and `AnnualVirtualStorage`.

Batch21:
> @Batch21 I've added another test for `AggregatedNode` and `AnnualVirtualStorage`.

looks good. Are there any other composite nodes that need a test? I think most inherit from the piecewise link..

PR #1007, opened 2021-10-04
Build & test Python 3.10 wheels.

jetuk:
This might not work just yet due to our dependencies, but we can re-run the pipeline until it does.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,8 +10,7 @@ env:
   # Only build on Python 3.7+
   CIBW_BUILD: cp37-* cp38-* cp39-* cp310-*
   # Skip 32-bit builds
-  # Skip 3.9 & 3.10 on Windows because of missing PyTables wheels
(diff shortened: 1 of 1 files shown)


jetuk:
This wasn't using the latest version of `cibuildwheel`. It is now and is failing to install `scipy`. They will issue wheels (1.7.2) soon. After this I will re-run the pipeline.

jetuk:
This is now blocked by PyTables

jetuk:
I've re-based this and enabled support for 3.9 and 3.10 builds on Windows and Linux. PyTables has released 3.7.0 which has wheels for all these platforms.

jetuk:
This is all working now. Wheels for 3.7 to 3.10 for Windows and Linux.

Batch21 (approved):
Looks good. Is skipping the MUSL linux build a major issue?

jetuk:
I believe its main use would be on Alpine Linux containers because they use
MUSL not Glibc. I think good enough for now.

PR #1010, opened 2021-11-15
Switch to black code formatter.

jetuk:
This moves all the Python code to black format and adds the formatter check to the CI.

Black doesn't yet support Cython, so none of the Cython code is reformatted.


Batch21 (approved):
good idea! I can't see any issues with the changes.

PR #1013, opened 2021-11-15
Uniform drawdown residual licence

Batch21:
Updates the `UniformDrawdownProfileParameter` so that a `residual_days` argument can be set to stop the profile reaching zero at the end of the year. This can be useful  as a way to avoid failure at the end of the year when the parameter is being used to drawdown licences evenly.


tests/test_parameters.py
```diff
@@ -1864,6 +1864,28 @@ def expected_func(timestep, scenario_index):
 
         m.run()
 
+    @pytest.mark.parametrize(['year'], [("2015",), ("2016",)])
+    def test_residual_days(self, year):
+
```
jetuk, line 1869: This might benefit from a quick explanation / docstring

jetuk (changes requested):

pywr/parameters/_parameters.pyx
```diff
@@ -898,10 +898,11 @@ cdef class UniformDrawdownProfileParameter(Parameter):
     --------
     AnnualVirtualStorage
     """
-    def __init__(self, model, reset_day=1, reset_month=1, **kwargs):
+    def __init__(self, model, reset_day=1, reset_month=1, residual_days=0, **kwargs):
```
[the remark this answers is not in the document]
Batch21: done

tests/test_parameters.py
Batch21: done

jetuk (approved):

PR #1015, opened 2021-11-22
Support inital_volume_pc in VirtualStorage nodes.

jetuk:
Add support for initial_volume_pc keyword argument when initialising `VirtualStorage` nodes and its subclasses. This allows
these nodes to use a parameter for maximum volume, which requires specifying both `initial_volume` and `initial_volume_pc`. Previously an error message indicated the user should provide `initial_volume_pc`, but when they tried this was not an valid keyword argument.

The tests handle the case where an `AnnualVirtualStorage` node resets of the first time-step (i.e. before the maximum volume parameter is evaluated).


Batch21 (approved):
Looks good

jetuk:
Rebased against master to fix conflict with test model names.

PR #1020, opened 2021-12-04
Improve handling of NaN current pc in storage nodes.

jetuk:
The current approach requires Parameter writers to handle
NaN values for current_pc. NaNs can occur when storage nodes
have zero maximum volume. This PR adds a convenience method
to get a finite and bounded current_pc for all storage nodes.
Pywr's internal parameters are updated to use this in preference
to direct access of the ._current_pc memory view.

Parameter writers should also use this new cpdef function instead
of using .current_pc directly.

There are two recorders which still use the raw ._current_pc
memory view for their data outputs. Those could be updated too.

We should also add some explicit tests for cases where using this
new internal function makes a difference.

### DIFF
--- a/pywr/_core.pxd
+++ b/pywr/_core.pxd
@@ -131,7 +131,7 @@ cdef class Storage(AbstractStorage):
     cdef Parameter _level_param
     cdef Parameter _area_param
     cpdef double get_initial_volume(self) except? -1
-    cpdef double get_initial_pc(self) except? -1    
+    cpdef double get_initial_pc(self) except? -1
     cpdef _reset_storage_only(self, bint use_initial_volume=*)
     cpdef double get_min_volume(self, ScenarioIndex scenario_index) except? -1
     cpdef double[:] get_all_min_volume(self, double[:] out=*)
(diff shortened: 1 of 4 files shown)


pywr/_core.pyx
```diff
@@ -1013,10 +1015,28 @@ cdef class AbstractStorage(AbstractNode):
             return np.asarray(self._volume)
 
     property current_pc:
-        """ Current percentage full """
+        """ Current proportion full.
+         
+        Note that this property is the raw internal value of the current_pc and may contain `NaN` values. Prefer
+        use of the `get_current_pc` method to return a guaranteed a finite value between 0.0 and 1.0.
```
Batch21, line 1021: Remove 2nd 'a'

pywr/_core.pxd
```diff
@@ -130,7 +131,7 @@ cdef class Storage(AbstractStorage):
     cdef Parameter _level_param
     cdef Parameter _area_param
     cpdef double get_initial_volume(self) except? -1
-    cpdef double get_initial_pc(self) except? -1
+    cpdef double get_initial_pc(self) except? -1
```
Batch21, line 134: remove space at end of line

Batch21 (commented):
The new method looks good. As you say, it would be good to have a few test to check the method deals correctly with NaNs and out of range values.

jetuk:
I wonder if we should have an equivalent for `get_current_volume`? That would avoid accessing `._volume` directly?

jetuk:
@Batch21 shall we merge this?

Batch21:
> 

👍

Batch21 (approved):

PR #1021, opened 2021-12-06
Improve GLPK error handling.

jetuk:
This picks up some work from #759 and the work done in #762 to
improve the error handling in GLPK. Best to read the new page
in the documentation that's been added about this. The TLDR
is that this adds NaN checks and GLPK error handling to Pywr
by default at the cost of some performance. The older
behaviour can be created with `--glpk-unsafe` compile time flag.

There are two new tests that cause seg faults with the unsafe
interface (i.e. current behaviour). Raising exceptions is better
from a handling point of view for users wrapping models in
multi-processing, etc.

Queries:
 - Should the unsafe interface be a runtime option instead of
a compile time option?
- Is changing the default error handling behaviour a good idea?

Useful reference: https://en.wikibooks.org/wiki/GLPK/Error_handling

### DIFF
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -99,6 +99,12 @@ else:
                 "PYWR_SOLVER_GLPK_FIXED_FACTORS_ONCE",
                 bool,
             )
+            kwargs = _parse_env_kwarg(
+                kwargs,
+                "use_safe_api",
+                "PYWR_SOLVER_GLPK_UNSAFE_API",
+                bool,
+            )
             self._cy_solver = cy_CythonGLPKSolver(**kwargs)
 
         def setup(self, model):
@@ -215,6 +221,12 @@ else:
                 kwargs,
                 "set_fixed_factors_once",
                 "PYWR_SOLVER_GLPK_FIXED_FACTORS_ONCE",
+                bool,
+            )
+            kwargs = _parse_env_kwarg(
+                kwargs,
+                "use_safe_api",
+                "PYWR_SOLVER_GLPK_UNSAFE_API",
                 bool,
             )
             self._cy_solver = cy_CythonGLPKEdgeSolver(**kwargs)
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -101,7 +101,7 @@ else:
             )
             kwargs = _parse_env_kwarg(
                 kwargs,
-                "use_safe_api",
+                "use_unsafe_api",
                 "PYWR_SOLVER_GLPK_UNSAFE_API",
                 bool,
             )
@@ -229,7 +229,7 @@ else:
             )
             kwargs = _parse_env_kwarg(
                 kwargs,
-                "use_safe_api",
+                "use_unsafe_api",
                 "PYWR_SOLVER_GLPK_UNSAFE_API",
                 bool,
             )
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -164,6 +164,10 @@ else:
             return self._cy_solver.use_presolve
 
         @property
+        def use_unsafe_api(self):
+            return self._cy_solver.use_unsafe_api
+
+        @property
         def set_fixed_flows_once(self):
             return self._cy_solver.set_fixed_flows_once
 
@@ -267,6 +271,10 @@ else:
         @property
         def use_presolve(self):
             return self._cy_solver.use_presolve
+
+        @property
+        def use_unsafe_api(self):
+            return self._cy_solver.use_unsafe_api
 
         @property
         def set_fixed_flows_once(self):
--- a/pywr/solvers/__init__.py
+++ b/pywr/solvers/__init__.py
@@ -62,7 +62,11 @@ class Solver(object):
 
 # Attempt to import solvers. These will only be successful if they are built correctly.
 try:
-    from .cython_glpk import CythonGLPKSolver as cy_CythonGLPKSolver
+    from .cython_glpk import (
+        CythonGLPKSolver as cy_CythonGLPKSolver,
+        GLPKError,
+        GLPKInternalError,
+    )
 except ImportError:
     pass
 else:
--- a/pywr/solvers/cython_glpk.pyx
+++ b/pywr/solvers/cython_glpk.pyx
@@ -137,6 +137,7 @@ cdef class GLPKSolver:
 
     cdef public bint use_presolve
     cdef bint has_presolved
+    cdef public bint use_unsafe_api
     cdef public bint set_fixed_flows_once
     cdef public bint set_fixed_costs_once
     cdef public bint set_fixed_factors_once
@@ -150,8 +151,9 @@ cdef class GLPKSolver:
             glp_delete_prob(self.prob)
 
     def __init__(self, use_presolve=False, time_limit=None, iteration_limit=None, message_level="error",
-                 set_fixed_flows_once=False, set_fixed_costs_once=False, set_fixed_factors_once=False):
+                 set_fixed_flows_once=False, set_fixed_costs_once=False, set_fixed_factors_once=False, use_safe_api=False):
         self.use_presolve = use_presolve
+        self.use_unsafe_api = use_safe_api
         self.set_solver_options(time_limit, iteration_limit, message_level)
         glp_term_hook(term_hook, NULL)
         self.has_presolved = False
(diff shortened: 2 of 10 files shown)


pywr/solvers/cython_glpk.pyx
```diff
@@ -217,55 +241,94 @@ cdef int term_hook(void *info, const char *s):
...
+
+    cdef int simplex(glp_prob *P, glp_smcp parm) except? -1:
+        """Wrapped call to `glp_simplex`"""
+        if not setjmp(error_ctx):
+            return glp_simplex(P, &parm)
+        else:
+            raise GLPKInternalError("An error occurred in `glp_simplex`." + glpk_error_msg)
+
+
+    cdef int set_obj_coef(glp_prob *P, int j, double coef) except? -1:
```
jetuk, line 282: `except` not `except?` as it doesn't return anything.

pywr/solvers/cython_glpk.pyx
```diff
@@ -217,55 +241,94 @@ cdef int term_hook(void *info, const char *s):
...
+        if isnan(coef):
+            raise GLPKError(f"NaN detected in objective coefficient of column {j}")
+
+        if not setjmp(error_ctx):
+            glp_set_obj_coef(P, j, coef)
+        else:
+            raise GLPKInternalError("An error occurred in `glp_set_obj_coef`." + glpk_error_msg)
+
+
+    cdef int set_row_bnds(glp_prob *P, int i, int type, double lb, double ub) except? -1:
```
jetuk, line 293: `except` not `except?` as it doesn't return anything.

jetuk:
Some benchmarks with random models. Note there's no heavy calculations in parameters in these models. There's no huge penalty. 

master:
```
-------------------------------------------------------------------------------------- benchmark: 6 tests -------------------------------------------------------------------------------------
Name (time in ms)                Min                 Max                Mean            StdDev              Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_random_model[5-5]       49.6719 (1.0)       51.3513 (1.0)       50.0567 (1.0)      0.4568 (2.42)      49.8965 (1.0)      0.1950 (1.0)           3;3  19.9774 (1.0)          19           1
test_random_model[10-5]      53.9678 (1.09)      56.4145 (1.10)      54.9284 (1.10)     0.9185 (4.87)      54.6484 (1.10)     1.2386 (6.35)          6;0  18.2055 (0.91)         18           1
test_random_model[20-5]      68.1346 (1.37)      68.7667 (1.34)      68.3732 (1.37)     0.1888 (1.0)       68.3751 (1.37)     0.2492 (1.28)          6;0  14.6256 (0.73)         14           1
test_random_model[30-5]      85.6237 (1.72)      88.9146 (1.73)      86.9593 (1.74)     1.0341 (5.48)      86.9716 (1.74)     1.6047 (8.23)          5;0  11.4996 (0.58)         12           1
test_random_model[40-5]     118.7882 (2.39)     119.9543 (2.34)     119.4524 (2.39)     0.4344 (2.30)     119.5829 (2.40)     0.7444 (3.82)          3;0   8.3715 (0.42)          8           1
test_random_model[50-5]     142.7505 (2.87)     143.8946 (2.80)     143.2559 (2.86)     0.4047 (2.14)     143.3535 (2.87)     0.5975 (3.06)          2;0   6.9805 (0.35)          7           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

```

issue759-glpk-error-handling (0557243):
```
-------------------------------------------------------------------------------------- benchmark: 6 tests -------------------------------------------------------------------------------------
Name (time in ms)                Min                 Max                Mean            StdDev              Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_random_model[5-5]       49.3335 (1.0)       50.3394 (1.0)       49.6144 (1.0)      0.2865 (1.0)       49.5389 (1.0)      0.2066 (1.0)           3;3  20.1554 (1.0)          19           1
test_random_model[10-5]      53.4245 (1.08)      55.9519 (1.11)      55.3322 (1.12)     0.5580 (1.95)      55.4927 (1.12)     0.3418 (1.65)          3;2  18.0727 (0.90)         18           1
test_random_model[20-5]      67.2573 (1.36)      68.5313 (1.36)      67.5973 (1.36)     0.3427 (1.20)      67.4687 (1.36)     0.2260 (1.09)          2;2  14.7935 (0.73)         15           1
test_random_model[30-5]      84.3992 (1.71)      85.6483 (1.70)      84.7664 (1.71)     0.4258 (1.49)      84.5793 (1.71)     0.5487 (2.66)          2;0  11.7971 (0.59)         12           1
test_random_model[40-5]     114.8334 (2.33)     115.9896 (2.30)     115.2580 (2.32)     0.4235 (1.48)     115.2147 (2.33)     0.5038 (2.44)          3;0   8.6762 (0.43)          9           1
test_random_model[50-5]     140.0376 (2.84)     141.8902 (2.82)     140.5383 (2.83)     0.6916 (2.41)     140.1434 (2.83)     0.7734 (3.74)          1;0   7.1155 (0.35)          7           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

```

Batch21:
I've tested the branch on a large model containing lots of complex custom components and there is not a notable impact on runtimes. The changes all look sensible but I'd need to spend a bit more time on it to understand how the error hook stuff works. 

Given the performance impacts do not appear that great, its probably better to keep the interface as a compile time option. I think any potential downsides of changing the default error handling (are there any major ones?) are outweighed by having better error messages. From experience, errors in the linear programme can often take a long time to debug due to lack of information on what the cause might be. So these changes should be a great help.

Batch21 (commented):
Gone through this again in more detail and though I still don't fully understand the error stuff I can see if uses the same approach as the GLPK example.

Only additional query is whether there should be a pipeline job that tests the unsafe option?

pywr/solvers/cython_glpk.pyx
```diff
@@ -217,55 +241,94 @@ cdef int term_hook(void *info, const char *s):
...
+
+    cdef int simplex(glp_prob *P, glp_smcp parm) except? -1:
+        """Wrapped call to `glp_simplex`"""
+        if not setjmp(error_ctx):
+            return glp_simplex(P, &parm)
+        else:
+            raise GLPKInternalError("An error occurred in `glp_simplex`." + glpk_error_msg)
+
+
+    cdef int set_obj_coef(glp_prob *P, int j, double coef) except -1:
```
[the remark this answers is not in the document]
jetuk: We can add `try...except` blocks around these function calls. In `setup` the performance cost is not an issue, but doing the same in the solve functions might be more costly. These could then add some more context to the error message. For example,

```cython
            try:
                set_mat_row(self.prob, self.idx_row_non_storages+row, len(cols), ind, val)
            except GLPKError:
                raise ModelStructureError(f'An error occurred creating the constrains for node "{node}".')
```

jetuk:
> Only additional query is whether there should be a pipeline job that tests the unsafe option?

Yes, there probably should be.

jetuk:
I've also noticed that if we don't catch these errors we'll be leaking memory by not freeing the data given to GLPK.

jetuk:
> I've also noticed that if we don't catch these errors we'll be leaking memory by not freeing the data given to GLPK.

This is now fixed.

jetuk:
CI error is solved in #1029

Batch21 (approved):
Error handling updates look good. As discussed, it would be good to do some testing around the performance impact of adding similar error handling to the problem update calls  in the `_solve_scenario` method, as I think this would be really useful. Maybe it could be another compile time option?

A final change that could be made in the PR would be to add a pipeline job to run the tests for the unsafe solver.

jetuk:
I'm working on making this a runtime option.

jetuk:
See latest commit for a runtime option to use the unsafe API.

jetuk:
This should all be working now. Some tests cause a seg fault with the unsafe API and must be skipped.

jetuk:
This is only failing on the docs on this branch. It works in master fine.

pywr/solvers/__init__.py
```diff
@@ -95,6 +99,12 @@ def __init__(self, *args, **kwargs):
                 "PYWR_SOLVER_GLPK_FIXED_FACTORS_ONCE",
                 bool,
             )
+            kwargs = _parse_env_kwarg(
+                kwargs,
+                "use_safe_api",
+                "PYWR_SOLVER_GLPK_UNSAFE_API",
```
Batch21, line 105: This is a little confusing. The env variable is for the unsafe api which is setting an arg for the safe api. Does one of them need updating?

pywr/solvers/cython_glpk.pyx
```diff
@@ -123,17 +146,22 @@ cdef class GLPKSolver:
...
 
     def __dealloc__(self):
-        glp_delete_prob(self.prob)
+        # If there's been an error the GLPK environment is destroyed and the pointer is invalid.
+        if not has_glpk_errored:
+            glp_delete_prob(self.prob)
 
     def __init__(self, use_presolve=False, time_limit=None, iteration_limit=None, message_level="error",
-                 set_fixed_flows_once=False, set_fixed_costs_once=False, set_fixed_factors_once=False):
+                 set_fixed_flows_once=False, set_fixed_costs_once=False, set_fixed_factors_once=False, use_safe_api=False):
```
Batch21, line 154: should the kwarg be `use_unsafe_api` ?

pywr/solvers/__init__.py
jetuk: This is actually incorrect. Well spotted.

pywr/solvers/cython_glpk.pyx
jetuk: Yes. I've fixed it.

PR #1025, opened 2021-12-07
Improve solver error messages for invalid networks.

jetuk:
If the user doesn't run `Model.check()` then the error from the solver is quite cryptic if their are missing connections between nodes. This change raises a `ModelStructureError` with a message containing the relevant node name which should help debugging some common mistakes.


jetuk:
I'll rebase this soon.

jetuk:
I've rebased this to work with #1021 and updated the top comment.

Batch21 (approved):
Looks good

PR #1028, opened 2021-12-09
Add section on checksums to the external data documentation.

jetuk:


Batch21 (changes requested):
Looks good - just one potential typo that I could spot

Batch21 (approved):

PR #1036, opened 2022-01-06
Update RollingVirtualStorage to calculate initial utilisation using the initial volume

Batch21:
Fixes #1036

### DIFF
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -1461,6 +1461,8 @@ cdef class RollingVirtualStorage(VirtualStorage):
         cdef double initial_pc = self.get_initial_pc()
         
         if initial_pc == 0.0:
+            if isinstance(self.max_volume, Parameter):
+                raise ValueError("`max_volume` cannot be a parameter if `initial_volume` or `initial_volume_pc` is 0.0")
             self._initial_utilisation = self._max_volume / (self.timesteps - 1)
         else:
             self._initial_utilisation = ((initial_vol / initial_pc) - initial_vol) / (self.timesteps - 1)
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -1459,7 +1459,11 @@ cdef class RollingVirtualStorage(VirtualStorage):
             raise ValueError('The number of time-steps for a RollingVirtualStorage node must be greater than one.')
         cdef double initial_vol = self.get_initial_volume()
         cdef double initial_pc = self.get_initial_pc()
-        self._initial_utilisation = ((initial_vol / initial_pc) - initial_vol) / (self.timesteps - 1)
+        
+        if initial_pc == 0.0:
+            self._initial_utilisation = self._max_volume / (self.timesteps - 1)
+        else:
+            self._initial_utilisation = ((initial_vol / initial_pc) - initial_vol) / (self.timesteps - 1)
         self._memory = np.full((self.timesteps-1, ncomb), self._initial_utilisation, dtype=np.float64)
         self._memory_pointer = 0
 
--- a/pywr/_core.pyx
+++ b/pywr/_core.pyx
@@ -1457,12 +1457,15 @@ cdef class RollingVirtualStorage(VirtualStorage):
         cdef int ncomb = len(model.scenarios.combinations)
         if self.timesteps < 2:
             raise ValueError('The number of time-steps for a RollingVirtualStorage node must be greater than one.')
-        self._memory = np.zeros((self.timesteps-1, ncomb), dtype=np.float64)
+        cdef double initial_vol = self.get_initial_volume()
+        cdef double initial_pc = self.get_initial_pc()
+        self._initial_utilisation = ((initial_vol / initial_pc) - initial_vol) / (self.timesteps - 1)
+        self._memory = np.full((self.timesteps-1, ncomb), self._initial_utilisation, dtype=np.float64)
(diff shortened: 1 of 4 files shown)


pywr/nodes.py
```diff
@@ -495,6 +495,9 @@ class RollingVirtualStorage(
 ):
     """A rolling virtual storage node useful for implementing rolling licences.
 
+    If the initial volume of the storage is less than the maximum volume then the parameter will calculate
+    an initial utilisation value. This is set equal to: (max volume - initial volume) / timesteps - 1.
```
jetuk, line 499: I think this needs to explain a bit more clearly that the initial utilisation is assumed to have happened equally over a number of periods prior to the model start. I.e. the licence will replenish at this rate until a full period is complete.

tests/test_run.py
```diff
@@ -441,6 +441,83 @@ def test_run_weekly(self):
...
+
+        from pywr.parameters import ConstantParameter
+
+        mx_vol = ConstantParameter(model, 100)
+        vs = pywr.core.RollingVirtualStorage(
+            model,
+            "Licence",
+            [lnk],
+            timesteps=3,
+            initial_volume=70.0,
```
jetuk, line 497: Can we test the case with zero initial volume here? Also what happens if initial volume > max volume?

pywr/nodes.py
jetuk: Should we also note the change in behaviour?

jetuk (changes requested):
Looks good, but I think the docstring could be clearer.

tests/test_run.py
Batch21: Added test for zero initial volume. When initial volume > max volume the `after` method on storage forces the volume back in bounds. However that extra volume is available for the first timestep. This test passes:

```python
model = pywr.core.Model(timestep="1D")

inpt = Input(model, "Input", max_flow=200.0)
lnk = Link(model, "Link")
inpt.connect(lnk)
otpt = Output(model, "Output", max_flow=200.0, cost=-10.0)
lnk.connect(otpt)

vs = pywr.core.RollingVirtualStorage(
    model,
    "Licence",
    [lnk],
    timesteps=3,
    initial_volume=200.0,
    max_volume=100.0,
)

model.setup()

assert_allclose(vs.volume, [200.0], atol=1e-7)

model.step()

assert_allclose(otpt.flow, [200.0], atol=1e-7)
assert_allclose(vs.volume, [0.0], atol=1e-7)
```
Where do you think this error should captured given that we can't rely on knowing the max volume during setup? maybe in the storage `before` method or in the solver `_solve_scenario` method when it checks for available volume..

pywr/nodes.py
Batch21: I've expanded the docstring a bit

tests/test_run.py
jetuk: Is this just an issue for RollingVirtualStorage or does some weird happen with other Storage nodes if initial_volume > max_volume? I think if it is a general issue then we can merge this as is and open an issue to resolve the wider problem.

Batch21: Looks like weird stuff happens with the other storage nodes. This passes:

```python

model = Model(timestep="1D")

storage = Storage(model, "storage", inputs=1, outputs=0, max_volume=20.0, initial_volume=30.0)
otpt = Output(model, "output", max_flow=0, cost=-1)
storage.connect(otpt)

model.setup()

assert_allclose(storage.volume, [30.0], atol=1e-7)

model.step()

assert_allclose(otpt.flow, [0.0], atol=1e-7)
assert_allclose(storage.volume, [30.0], atol=1e-7)

model.step()

assert_allclose(otpt.flow, [0.0], atol=1e-7)
assert_allclose(storage.volume, [30.0], atol=1e-7)
```
Unlike the virtual storages, the additional capacity of storage nodes is not removed after first timestep as the adjustment block in the `after` method does not run.

I think this line in the solver avoids any weird stuff happening in the LP in this situation as the bound just gets set to 0.0:

```python
 ub = max(max_volume - storage._volume[scenario_index.global_id], 0.0) / timestep.days
```
If it could go negative, I assume it would try and force water out of the storage and fail to find a solution if it could not.

I agree that it would be better to deal with this as a separate issue.

Batch21: Ok. Are you happy with the changes I've made?

pywr/nodes.py
```diff
@@ -558,6 +567,11 @@ def __init__(self, model, name, nodes, **kwargs):
         self.days = days
         self.timesteps = timesteps
 
+        if self.get_initial_volume() == 0.0 and isinstance(self.max_volume, Parameter):
```
jetuk, line 570: Should this not be raised in `.setup`?

pywr/_core.pyx
```diff
@@ -1459,7 +1459,11 @@ cdef class RollingVirtualStorage(VirtualStorage):
             raise ValueError('The number of time-steps for a RollingVirtualStorage node must be greater than one.')
         cdef double initial_vol = self.get_initial_volume()
         cdef double initial_pc = self.get_initial_pc()
-        self._initial_utilisation = ((initial_vol / initial_pc) - initial_vol) / (self.timesteps - 1)
+        
+        if initial_pc == 0.0:
```
jetuk, line 1463: I think we should do the error handling here to be on the safe side.

jetuk (approved):

PR #1040, opened 2022-01-20
Move project dependencies to pyproject.toml

jetuk:
Add a new optional dependency group for notebook sub-package. This
removes ipython, matplotlob and jinja2 from core dependencies.


jetuk:
This was a very long winded way to remove `ipython`, `jinja2` and `matplotlib` as core dependency (i.e. `install_requires`).

Batch21 (approved):

PR #1041, opened 2022-01-26
Model inception

jetuk:
To be updated once complete.


pywr/model.py
```diff
@@ -1 +1,154 @@
...
+            sub_model_name = sub_model_definition["name"]
+            sub_model_path = sub_model_definition["path"]
+            sub_model_solver = sub_model_definition.get("solver", solver)
+            if path is not None:
+                sub_model_path = os.path.join(os.path.dirname(path), sub_model_path)
+
+            sub_model = Model.load(sub_model_path, path=path, solver=sub_model_solver)
+            sub_model.timestepper.start = timestepper_data["start"]
+            sub_model.timestepper.end = timestepper_data["end"]
+            sub_model.timestepper.timestep = timestepper_data["timestep"]
```
Batch21, line 83: Does the model also need to update the scenario attributes of the child models?

pywr/model.py
```diff
@@ -1 +1,154 @@
...
+    def setup(self):
+        for model in self.models.values():
+            model.setup()
+
+    def reset(self):
+        for model in self.models.values():
+            model.reset()
+
+    # def _step(self):
+    #     for model in self.models.values():
```
Batch21, line 104: You thinking of keeping this implementation? Might be useful to be able to specify this one or the one below in the model definition. Though I guess it would change how the new parameters work.

pywr/model.py
```diff
@@ -1 +1,154 @@
...
+        try:
+            if self.dirty:
+                self.setup()
+            else:
+                self.reset()
+            t1 = time.time()
+
+            # Begin time-stepping
+            for timesteps in zip(*timesteppers):
+                # TODO assert these are all the same date
```
Batch21, line 143: If the timesteppers are being set equal in the load method is an assert needed here?

pywr/parameters/multi_model_parameters.py
```diff
@@ -0,0 +1,85 @@
...
+        super().__init__(model, **kwargs)
+        self.other_model = other_model
+        self.node = node
+
+        self._other_model = None
+        self._other_model_node = None
+
+    def setup(self):
+        super().setup()
+        # Find the references to the other model and one of its parameters.
```
Batch21, line 73: Update comment to say 'its storage nodes'

tests/models/three-dependent-storage-sub-models/integrated-model.json
```diff
@@ -0,0 +1,29 @@
+{
+  "metadata": {
+    "title": "Simple 1",
+    "description": "A very simple example.",
+    "minimum_version": "1.16.0"
```
Batch21, line 5: minimum version will need updating

pywr/model.py
```diff
@@ -1 +1,154 @@
...
+class MultiModel:
+    """A class for running multiple Pywr model simultaneously.
+
+    The `MultiModel` class allows running multiple `pywr.Model` simultaneously. The combined simulation runs
+    the sub-models in an interleaved fashion. Each sub-model is progressed one time-step in a pre-determined order
+    before moving on the next time-step.
+    """
+
+    def __init__(self):
+        self.models = {}  # TODO this does depend on ordering here
```
Batch21, line 22: using a dict will only an issue if someone uses python 3.5 or earlier?

tests/test_multi_models.py
```diff
@@ -0,0 +1,168 @@
...
+
+    np.testing.assert_allclose(
+        multi_model.models["model-0"].nodes["output-0"].flow[0], 5.0
+    )
+    np.testing.assert_allclose(
+        multi_model.models["model-1"].nodes["output-1"].flow[0], 6.0
+    )
+
+
+def test_run_two_independent_models_from_json():
```
Batch21, line 54: Maybe needs a simple test with multiple scenarios

Batch21 (commented):
The implementation of the MultiModel looks good. Main gap I think is dealing with scenarios. Currently having one sub-model with more scenarios than another can result in out of bounds index errors.

pywr/model.py
```diff
@@ -1 +1,154 @@
...
+
+    def add_model(self, name: str, model: Model):
```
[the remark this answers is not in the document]
jetuk: No, this is helpful to build a sub-model using Python and not from JSON.

pywr/model.py
jetuk: No, removed.

pywr/model.py
jetuk: They can be modified between load and run, or also created without the load method. So, this check is a good one. It's now implemented.

pywr/model.py
jetuk: True, still makes me nervous!

pywr/recorders/recorders.py
```diff
@@ -660,3 +660,19 @@ def generate_dataframes(h5file, time="/time", scenarios="/scenarios"):
 
 
 TablesRecorder.register()
+
+
+class H5PyRecorder(Recorder):
```
jetuk, line 665: This should be here.

pywr/profiler.py
```diff
@@ -0,0 +1,59 @@
...
+@dataclass
+class ProfilerEntry:
+    phase: str
+    class_name: str
+    name: str
+    perf_counter: float
+    memory_usage: float
+
+
+class Profiler:
```
jetuk, line 25: @Batch21 do we want to keep this bundled with Pywr? Or should I remove it?

Batch21: I think it could be useful to keep but maybe with an expanded docstring that shows how to use it.

jetuk: OK, I'll have to make some changes because the `resources` module does not work on Windows.

Batch21 (approved):
Had another look through this, focusing mainly on the profiler, and it looks good to merge.

Sorry for taking so long to do this.

PR #1057, opened 2022-07-05
Add StorageParameter

### LINKED ISSUE #1056: Parameter equal to Node values
Hi, I was wondering if there is a parameter which is equal to the value of a node. For example, I want to have a parameter that is the storage of my reservoir and it's equivalent to the storage values of this node. I mean, If I recoder the node and the parameter the values will be the same. 

Thanks in advance.

jetuk:
Fixes #1056.

### DIFF
--- a/docs/source/api/pywr.parameters.rst
+++ b/docs/source/api/pywr.parameters.rst
@@ -161,6 +161,7 @@ Other parameters
    PiecewiseIntegralParameter
    FlowParameter
    FlowDelayParameter
+   StorageParameter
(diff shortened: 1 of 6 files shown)


jetuk:
> This looks good. Only query is whether the test should also check the `use_proportional_volume` attribute

I've added this test. Can you approve if happy?

Batch21 (approved):

aiduran:
Thank you for all, how I can upgrade the new version of pywr?

PR #1072, opened 2023-01-11
feat: update TablesRecorder to save scenario slice data

Batch21:


pywr/recorders/recorders.py
```diff
@@ -473,6 +473,33 @@ def setup(self):
                         entry[s.name] = i
                     entry.append()
                 tbl.flush()
+            elif any([s.slice for s in self.model.scenarios.scenarios]):
+                # Slices are only applied in a model run if there are no user combinations
+                description = {
+                    "name": tables.StringCol(1024),
+                    "slice_start": tables.Int64Col(),
```
jetuk, line 480: I think the "slice_" prefix is redundant here as we're in the context of "scenario slices" for the table.

pywr/recorders/recorders.py
```diff
@@ -473,6 +473,33 @@ def setup(self):
...
+                    "slice_stop": tables.Int64Col(),
+                    "slice_step": tables.Int64Col(),
+                }
+                tbl = self.h5store.file.create_table(
+                    group_name,
+                    "scenario_slices",
+                    description=description,
+                    createparents=True,
```
jetuk, line 488: Should `createparents=True` also be applied to the "scenario_combinations" table creation on line 468?

tests/test_recorders.py
```diff
@@ -1232,6 +1232,43 @@ def test_user_scenarios(self, simple_linear_model, tmpdir):
...
+            model.run()
+
+            for node_name in model.nodes.keys():
+                ca = h5f.get_node("/", node_name)
+                assert ca.shape == (365, 3, 2)
+
+            # check slices table exists
+            slices = h5f.get_node("/scenario_slices")
+            for s in slices.iterrows():
+                print(s)
```
jetuk, line 1268: This should assert some actual content and remove the print statements.

jetuk (changes requested):
This is a good addition to the `TablesRecorder`. I think we should add a longer description of these meta tables to the docstring. If you don't have time don't worry, but would be good to get done.

jetuk:
Updates look good. Do you want to complete the docstring changes?

jetuk (approved):

PR #1073, opened 2023-01-17
feat: Add timestep offset to DataFrameParameter and TablesArrayParameter

jetuk:
This adds a new feature to allow offsetting the timeseries provided by DataFrameParameter and TablesArrayParameter. This would allow the user to look-up future or past values if required.

Some minor black formatting updates too.


Batch21 (commented):
This looks good. Only question I have is whether the docstring should state what happens for negative offsets for the initial timesteps and positive offsets for the final timesteps.

PR #1074, opened 2023-01-17
feat: Add RollingMeanFlowNodeParameter

jetuk:
Add a new parameter that computes the rolling mean flow of the previous N timesteps. This is the parameter equivalent of `RollingMeanFlowNodeRecorder`. This recorder has been marked as deprecated. Users should prefer this new parameter to the older recorder.

### DIFF
--- a/pywr/parameters/_parameters.pyx
+++ b/pywr/parameters/_parameters.pyx
@@ -2161,6 +2161,88 @@ cdef class DiscountFactorParameter(Parameter):
 DiscountFactorParameter.register()
 
 
+cdef class RollingMeanFlowNodeParameter(Parameter):
+    """Returns the mean flow of a Node for the previous N timesteps.
+
+    Parameters
+    ----------
+    model : `pywr.core.Model`
+    node : `pywr.core.Node`
+        The node to record
+    timesteps : int
+        The number of timesteps to calculate the mean flow for
+    name : str (optional)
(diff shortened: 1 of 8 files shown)


pywr/parameters/_parameters.pyx
```diff
@@ -2161,6 +2161,88 @@ cdef class DiscountFactorParameter(Parameter):
...
+    initial_flow : float
+        The initial value to use in the first timestep before any flows have been recorded.
+    """
+    def __init__(self, model, node, timesteps=None, days=None, initial_flow=0.0, **kwargs):
+        super(RollingMeanFlowNodeParameter, self).__init__(model, **kwargs)
+        self.node = node
+        self.initial_flow = initial_flow
+
+        if not timesteps and not days:
+            raise ValueError("Either `timesteps` or `days` must be specified.")
```
Batch21, line 2185: Should it also error if both are given. Looks like it would default to using `days` in that case but this might not be obvious to a user.

Batch21 (changes requested):
A few minor comments but otherwise looks good.

pywr/parameters/_parameters.pyx
jetuk: I've added a comment that `timesteps` is ignored if `days` is given.

Batch21 (approved):

Issue #1070, opened 2023-01-05
A new graphical user interface for pywr


s-simoncelli:
Hello,

I have just released an open-source graphical user interface for pywr, called pywr-editor. 

The application, written in Python with Qt, is available at https://github.com/pywr-editor/editor and lets you build and fully customise every component of a pywr model, from parameters, nodes, recorders, metadata, tables, imports to slots. It also offers an interactive schematic with most CAD-like features such as panning, zooming and drag and drop. 

Although the application has been fully tested, I would appreciate any feedback and help with reporting potential bugs.

Cheers,
Stefano

wdvichete84:
@s-simoncelli that´s greate tool for large and complex network.
I appreciate your work.

jetuk:
Hi Stefano,

This is really cool! It looks very well written. I'll take a look more closely and give it a try soon.

I'll also mention that we have released a second beta of our Pywr result viewer here: https://github.com/jtaltd/pywr-viewer-desktop-releases

Perhaps we could add a related projects / tools section to Pywr's README to sign post users to these tools.

jetuk:
This all sounds good to me. Pywr's README could do with some work anyway. I'll try to find some time to do a bit of an update and include a section for related projects & tools.

PR #1076, opened 2023-01-19
Update changelog for v1.20.0 release.

jetuk:
Couple of queries ahead of this release:

- Should we do a README update (see #1070) first?
- Should we also convert `RollingWindowParameterRecorder` to a parameter as per #1074 ?


Batch21 (approved):
Both definitely need doing but probably aren't essential for this release. The readme could do with a more general refresh to update stuff like the copyright dates.

PR #1083, opened 2023-03-13
feat: Add metadata to TablesRecorder arrays.

jetuk:
Add metadata to the CArray nodes created in the HDF5 file by TablesRecorder. Metadata saves the type of attribute and the class name of the source of the data.


Batch21 (commented):
Looks good. Is it with adding a simple test to check that these attributes are in the output file?

Batch21 (approved):

PR #1096, opened 2023-06-29
Pass load options to optimiser

s-simoncelli:
Hello,

there are some instances where passing the Model.load options to the optimisation wrapper can be useful (such as setting the model path to handle relative paths). I am not sure if this is something other users are interested in.

### DIFF
--- a/pywr/optimisation/__init__.py
+++ b/pywr/optimisation/__init__.py
@@ -60,7 +60,7 @@ class BaseOptimisationWrapper(object):
     def __init__(self, pywr_model_json, *args, **kwargs):
         uid = kwargs.pop("uid", None)
         self.pywr_model_klass = kwargs.pop("model_klass", Model)
-        self.pywr_model_options = kwargs.pop("model_options", {})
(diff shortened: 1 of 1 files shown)


jetuk (approved):
This looks good to me. 

Thanks for the contribution.

pywr/optimisation/__init__.py
```diff
@@ -60,6 +60,7 @@ class BaseOptimisationWrapper(object):
     def __init__(self, pywr_model_json, *args, **kwargs):
         uid = kwargs.pop("uid", None)
         self.pywr_model_klass = kwargs.pop("model_klass", Model)
+        self.pywr_model_options = kwargs.pop("model_options", {})
```
jetuk, line 63: On second thought, could we rename this to be `model_kwargs`?

jetuk (changes requested):

pywr/optimisation/__init__.py
s-simoncelli: Sorted

jetuk (approved):

Issue #1102, opened 2023-08-04
Pywr is incompatible with cython 3.0.0


knoxsp:
Trying to install Pywr using cython 3.0.0 (the latest as of this date) results in the error:
```
[ 1/10] Cythonizing pywr/solvers/cython_glpk.pyx

Error compiling Cython file:
------------------------------------------------------------
...
    def __init__(self, use_presolve=False, time_limit=None, iteration_limit=None, message_level="error",
                 set_fixed_flows_once=False, set_fixed_costs_once=False, set_fixed_factors_once=False, use_unsafe_api=False):
        self.use_presolve = use_presolve
        self.use_unsafe_api = use_unsafe_api
        self.set_solver_options(time_limit, iteration_limit, message_level)
        glp_term_hook(term_hook, NULL)
                      ^
------------------------------------------------------------

pywr/solvers/cython_glpk.pyx:158:22: Cannot assign type 'int (void *, const char *) except? -1' to 'int (*)(void *, const char *) noexcept'

Error compiling Cython file:
------------------------------------------------------------
...
        self.has_presolved = False
        self.set_fixed_flows_once = set_fixed_flows_once
        self.set_fixed_costs_once = set_fixed_costs_once
        self.set_fixed_factors_once = set_fixed_factors_once
        # Register the error hook
        glp_error_hook(error_hook, &error_ctx)
                       ^
------------------------------------------------------------
```

knoxsp:
Reverting cython to 0.29.36 resolves this issue

jetuk:
We pinned the version to < 3 in `pyproject.toml` in the last release. 

The issue is to do with handling exceptions in the Cython. I've not got around to digging in to it. There's a migration guide here: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html

knoxsp:
Thanks for the response. I mainly wanted this issue open in case others encouter the same problem and searched for a solution

jetuk:
Fixed in #1104

PR #1104, opened 2023-08-22
chore: Support Cython 3.x

jetuk:
Several updates required to handle building with Cython 3.x.

- Include `noexcept` for GLPK callback functions.
- Handle name mangling of attributes using __. Cython 3 aligns more closely with Python 3 in this respect.

There should be a review of the use of `except` given the new behaviour in Cython 3, but that is not done here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4", "cython<3", "numpy"]
+requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4", "cython", "numpy"]
 
 # enable version inference
 [tool.setuptools_scm]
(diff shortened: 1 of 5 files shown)


jetuk:
This fixes #1102

jetuk:
It might also be possible to remove the Numpy API warnings, but that could be a separate PR.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4", "cython<3", "numpy"]
```
Batch21, line 2: Changes to attribute names mean that Pywr can't now be built with cython<3 so this should probably be updated to "cython>=3"

jetuk: Done

Batch21 (approved):

PR #1118, opened 2024-02-27
feat: cache file hashes to avoid recomputation when the same file is referenced multiple times

Batch21:


Batch21:
@jetuk should I pull out the black formatting and text fix commits into separate PRs?

pywr/_model.pyx
```diff
@@ -396,6 +396,7 @@ class Model(object):
 
         model._parameters_to_load = collect_components(data, "parameters")
         model._recorders_to_load = collect_components(data, "recorders")
+        model._file_hashes = {}
```
jetuk, line 399: Maybe put this in `__init__` so it is present even when using the Python API.

tests/test_hashes.py
```diff
@@ -32,9 +33,12 @@ def test_hash_timeseries2_(filename, algorithm, hash, correct):
     """Test the hash value of files in the models directory"""
 
     fullname = os.path.join(TEST_FOLDER, "models", filename)
+    model = Model()
+    model._file_hashes = {}
```
jetuk, line 37: ... then this can be removed?

jetuk (changes requested):
Agreed it would probably be a good idea to pull out the test and formatting fixes to separate PRs.

pywr/hashes.py
```diff
@@ -20,17 +20,27 @@ class HashMismatchError(IOError):
...
-def check_hash(filename, hash, algorithm="md5", **kwargs):
+def check_hash(model, filename, hash, algorithm="md5", **kwargs):
     """Check the hash for filename using the named algorithm
 
     If the hashes do not match a HashMismatchError error is raised.
 
     This function is not case sensitive.
     """
 
+    if (cached_hash := model._file_hashes.get((filename, algorithm))) is not None:
```
jetuk, line 31: What's your opinion on this being a method on `Model`? It is bad practice to access private attributes of an object.

Batch21: yes, it probably is. I've moved it. Do you think its worth keeping this as a standalone module or could the rest of the code be moved as well?

pywr/hashes.py
```diff
@@ -18,19 +18,3 @@ def compute_hash(filename, algorithm="md5", chunk_size=io.DEFAULT_BUFFER_SIZE):
 
 class HashMismatchError(IOError):
     pass
-
-
-def check_hash(filename, hash, algorithm="md5", **kwargs):
```
jetuk, line 23: I wonder if some users might be relying on this a standalone  utility function? Removing it is a bit of an API change. You could keep this, but add an argument

```python
def check_hash(filename, hash, algorithm="md5", cache=None, **kwargs):
    ...
    if cache is not None:
        cache[(filename, algorithm)] = actual_hash
    
    # Then in the `Model.check_hash`
    def check_hash(self, filename, hash, algorithm="md5", **kwargs):
        ...
        check_hash(filename, hash, algorithm=algorithm, cache=self._file_hashes, **kwargs)
```

```

jetuk (approved):

PR #1124, opened 2024-04-24
feat: Add support for defining net or gross loss in LossLink.

jetuk:


Batch21 (approved):
Implementation looks good.

PR #1125, opened 2024-04-24
ci: Use delvewheel to repair Windows wheels.

### LINKED ISSUE #1123: Use delvewheel to repair Windows wheels
There's currently a warning about DLLs missing from the RECORD on Windows when installing the Pywr wheel from PyPI. We should update the build process to use https://github.com/adang1345/delvewheel

jetuk:
Fixes #1123.


Batch21 (approved):
LGTM

PR #1133, opened 2024-06-25
feat: Fallback to loading from values or tables.

jetuk:
Try to load parameter as values (i.e. from a table or values entry) if there is no "type" entry. This should allow more flexible definitions of models where tabular data can be used where a parameter is required. Previously only a literal constant or Parameter definition were allowed.


Batch21 (commented):
This is a useful change! Only query is whether there is a good place to document it to make users aware that tables can now be reference like this?

Batch21 (approved):

PR #1134, opened 2024-07-18
Replace interp1d in interpolation parameters

pmslavin:
This request replaces the use of `scipy.interpolate.interp1d` in interpolation parameters with the `scipy.interpolate.make_interp_spline` function. Profiling indicates this gives a ~60% reduction in time per call to `.value()` for interpolating functions of the same degree and tests continue to pass.

The same set of `interp_kwargs` arguments are accepted by the `AbstractInterpolatedParameter` as previously.  The behaviours of `interp1d` in terms of the `bounds_error` and `fill_values` arguments are maintained by having the `AbstractInterpolatedParameter` process these and delegate calls to `.value()` to a method which does the minimal amount of checking consistent with the kwargs; the specified fill values are returned for arguments below or above the interpolated range, or `ValueError` is raised as specified.

The str `kind` argument is mapped to an int degree which can be passed directly to `make_interp_spline`.

A use of the deprecated `pytest.warns(None)` is removed from `tests/test_parameters.py` and replaced with the alternative descibed in [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests).

### DIFF
--- a/pywr/parameters/parameters.py
+++ b/pywr/parameters/parameters.py
@@ -191,8 +191,7 @@ class AbstractInterpolatedParameter(Parameter):
         the default behaviour of the interpolating instance
         """
         v = self._value_to_interpolate(ts, scenario_index)
-        val = self.interp(v)
-        return val
+        return self.interp(v)
 
 
 class InterpolatedParameter(AbstractInterpolatedParameter):
--- a/pywr/parameters/parameters.py
+++ b/pywr/parameters/parameters.py
@@ -76,7 +76,7 @@ from .control_curves import (
 )
 from . import multi_model_parameters
 import numpy as np
-from scipy.interpolate import interp1d
+from scipy.interpolate import make_interp_spline
 from scipy.integrate import quad
 import pandas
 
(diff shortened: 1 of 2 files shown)


jetuk:
Thanks for this it looks good. I haven't profiled it myself. My only concern is that the new interface / function does not cover all the possibilities in `interp1d`. For example, I can not see a way you would implement "nearest", "next" or "previous" kinds that are currently supported. I worry someone is using that and this update will break their model.

Instead, perhaps we should have explicit spline interpolation parameters that use the `make_interp_spline` function to properly separate behaviour. It could also have a JSON schema that aligns more closely with the arguments to that function (instead of mapping from "kind"). Then we can follow the scipy approach and deprecated the ones using `interp1d`.

@Batch21 thoughts?

Batch21:
Separate spline interpolation parameters seems like a sensible approach to me. Worth making the effort to minimise the risk of breaking peoples models.

pmslavin:
Yes, I agree it makes more sense to move this to a new and distinct parameter rather than reimplement the whole functionality of the `interp1d` "kind" argument for any models that expect this to be passed through to scipy. I suppose it is also less confusing from a development perspective to have a new interface in the parameter for the new underlying `make_interp_spline` instead of presenting the interface of an underlying function that is no longer used!

I'll look at how this can be reimplemented to use other-than-interp1d-interpolation without affecting the existing family of `AbstractInterpolatedParameter`-derived parameters.

PR #1140, opened 2024-10-07
feat: allow a columns attribute to be sepecified for tables and dataframes

Batch21:
This allows datasets to be filtered so that only a subset of columns are used in a model.


jetuk:
I guess rebase this and see if it passes the CI?

jetuk (approved):

PR #1143, opened 2024-10-17
feat: Support Python 3.13

jetuk:


jetuk:
Looks like this is blocked on a PyTables release / wheel for 3.13. I guess we can park it for now.

Batch21 (approved):

PR #1145, opened 2024-10-18
chore: Drop support for Python 3.9

jetuk:


Batch21:
It doesn't look like 3.9 is EOL until next year. Is a dependency requiring this?

jetuk:
Numpy dropped support for Python 3.9 in v2.1.0 in August. We've followed Numpy in the past on this.

Batch21 (approved):

PR #1151, opened 2025-01-27
feat: add value method to the AggregatedParameter

laurenpetch:
Adding the value method to the AggregatedParameter so it can be used as the threshold value in a ParameterThresholdParameter


jetuk:
I think this particular issue was fixed in #1148 and release in v1.27.2 .

laurenpetch:
Woops wasn't running on the least version locally so didn't realise. I'll close

Issue #1110, opened 2023-10-19
Multiple scenario_combinations resulting in an index error.


knoxsp:
I have adapted one of the test models to have two scenarios, and to use combinations of the two. See below.

the key part is: `   "scenario_combinations": [[0, 0], [4, 1], [1, 0], [0, 2]],`

This results in an error:

```
  File "pywr/_model.pyx", line 628, in pywr._model.Model.run
  File "pywr/_model.pyx", line 604, in pywr._model.Model._step
  File "pywr/_model.pyx", line 825, in pywr._model.Model.before
  File "pywr/parameters/_parameters.pyx", line 76, in pywr.parameters._parameters.Parameter.calc_values
  File "pywr/parameters/_parameters.pyx", line 268, in pywr.parameters._parameters.DataFrameParameter.value
  IndexError: Out of bounds on buffer access (axis 1)
```

```json
{
    "metadata": {
        "title": "Timeseries example",
        "description": "A model with a timeseries",
        "minimum_version": "0.1"
    },
    "timestepper": {
        "start": "2025-01-01",
        "end": "2029-01-01",
        "timestep": "A"
    },
    "scenarios": [
        {
            "name": "scenario A",
            "size": 5
        },
        {
            "name": "scenario B",
            "size": 5
        }

    ],
    "scenario_combinations": [[0, 0], [4, 1], [1, 0], [0, 2]],
    "nodes": [
        {
            "name": "catchment1",
            "type": "catchment",
            "flow": "inflow"
        },
        {
            "name": "reservoir1",
            "type": "Storage",
            "initial_volume": 50.0,
            "max_volume": 9999
        },
        {
            "name": "abs1",
            "type": "link",
            "max_flow": 50
        },
        {
            "name": "demand1",
            "type": "Output",
            "max_flow": "demand",
            "cost": -10
        },
        {
            "name": "term1",
            "type": "Output",
            "cost": 5
        }
    ],
    "edges": [
        ["catchment1", "reservoir1"],
        ["reservoir1", "abs1"],
        ["abs1", "demand1"],
        ["reservoir1", "term1"]
    ],
    "parameters": {
        "inflow": {
            "type": "dataframe",
            "scenario": "scenario A",
            "data": {
                "0": {
                  "2025-01-01": 160,
                  "2026-01-01": 161,
                  "2027-01-01": 162,
                  "2028-01-01": 162,
                  "2029-01-01": 163
                },
                "1": {
                  "2025-01-01": 162,
[...]

PR #1153, opened 2025-05-07
Fixed index conversion to datetime index

s-simoncelli:
When the data is given as dictionary to build the `DataFrame` in `read_dataframe`, the conversion to datetime index using `pandas.DatetimeIndex` creates invalid dates as the index values are integer numbers. I added a new option called `index` to set the index first. For example:

```python
pywr_dict=  {
      "type": "DataFrameParameter",
      "parse_dates": True,
      "index": ["2000-1-1", "2000-1-2"],
      "data": {
          "values": [12,13, 14],
      },
}

df = read_dataframe(model, data=pywr_dict)
```

### DIFF
--- a/pywr/dataframe_tools.py
+++ b/pywr/dataframe_tools.py
@@ -273,7 +273,11 @@ def read_dataframe(model, data):
         df = pandas.read_hdf(url, key=key, **data)
     elif filetype == "dict":
         parse_dates = data.pop("parse_dates", False)
+        index = data.pop("index") if "index" in data else None
+
         df = pandas.DataFrame.from_dict(df_data, **data)
+        if index:
+            df.index = index
         if parse_dates:
             df.index = pandas.DatetimeIndex(df.index)
 


jetuk:
I think you can have this work with the current behaviour by using `orient='index' or `orient='tight'`. The following tests both pass for me. I also think the default `orient=columns` can work if you pass a dictionary of dictionaries where the inner one is keyed by row index. See also #1110.  

Do these options work for you, or do you need the special case?

```python
def test_read_dataframe_from_dict_orient_index():
    """Test reading a dataframe from a dictionary using index orientation"""
    model = load_model("simple1.json")

    pywr_dict = {
        "parse_dates": True,
        "orient": "index",
        "columns": ["values"],
        "data": {
            "2000-1-1": [12],
            "2000-1-2": [13],
            "2000-1-3": [14],
        },
    }

    df = read_dataframe(model, data=pywr_dict)
    assert df.shape == (3, 1)
    assert df.index[0] == pandas.to_datetime("2000-01-01")


def test_read_dataframe_from_dict_orient_tight():
    """Test reading a dataframe from a dictionary using tight orientation"""
    model = load_model("simple1.json")

    pywr_dict = {
        "parse_dates": True,
        "orient": "tight",
        "data": {
            "index": ["2000-1-1", "2000-1-2", "2000-1-3"],
            "columns": [0],
            "data": [[12], [13], [14]],
            "index_names": ["dates"],
            "column_names": ["scenario"],
        },
    }

    df = read_dataframe(model, data=pywr_dict)
    assert df.shape == (3, 1)
    assert df.index[0] == pandas.to_datetime("2000-01-01")

```

s-simoncelli:
Hi,

yes I think the first example should work for me. Thanks :)

PR #1154, opened 2025-05-21
fix: Fix warning about assigning to variables attribute.

jetuk:


jetuk:
@Batch21 if you could approve these without the linting. I will probably move to ruff soon for that anyway.

Batch21 (approved):

PR #1156, opened 2025-05-23
chore: pyproject.toml and build modernisation.

jetuk:
Modernisation of the build process and project metadata.

- Move more metadata to pyproject.toml (from setup.py).
- Specify the build-system and backend.
- Remove parsing command line arguments in build (this is not supported). Allow existing CLI arguments to be passed as environmental variables instead.
- Update CI accordingly.
- Update installation instructions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@ from setuptools.command.build_ext import build_ext as _build_ext
 
 
 def setup_package():
+    print("Building pywr with Cython extensions")
     compiler_directives = {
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -5,6 +5,7 @@
 
 
 def setup_package():
+    print("Building pywr with Cython extensions")
```
Batch21, line 8: Did this need deleting?

Batch21 (approved):

PR #1163, opened 2025-07-28
Pass timestep_offset in TablesArrayParameter.load()

s-simoncelli:
The timestep_offset  option cannot be set when the parameter is loaded from JSON


pywr/parameters/_parameters.pyx
```diff
@@ -508,7 +508,7 @@ cdef class TablesArrayParameter(IndexParameter):
         for algo, hash in checksums.items():
             model.check_hash(url, hash, algorithm=algo)
 
-        return cls(model, url, node, where=where, scenario=scenario)
+        return cls(model, url, node, where=where, scenario=scenario, timestep_offset=data.pop('timestep_offset', 0))
```
jetuk, line 511: In other parameters we pass the remaining `data` as keyword arguments: `**data`. That should also help catching an incorrect keys.

PR #1183, opened 2026-05-07
perf: Improve the performance of TablesRecorder2

jetuk:
Cythonise TablesRecorder2 and add buffering. The buffering seems to be the more important update for performance. It works by creating buffer array for each item to be saved. The size of this array is based from a user defined size of memory spread evenly over each item.

The buffers are then filled in chunks of timesteps, only passing their data to PyTables periodically.

For very small simulations (e.g. single scenarios) this will result in all of the data being buffered and only flushed to PyTables at the end of the simulation. This can speed-up those simulations a lot compared to the original implementation.

### DIFF
--- a/pywr/recorders/_tables_recorders.pyx
+++ b/pywr/recorders/_tables_recorders.pyx
@@ -258,7 +258,7 @@ class TablesRecorder2(Recorder):
 
         # Work out the number of timesteps to buffer
         if self.buffer_timesteps is not None:
-            self._buffer_num_timesteps = min(self._buffer_num_timesteps, len(self.model.timestepper))
+            self._buffer_num_timesteps = min(self.buffer_timesteps, len(self.model.timestepper))
         elif self.buffer_size is not None:
             # Divide by 8 as we are storing 64-bit floats
             # Also assume most data is nodes, and we store 3 values for each
--- a/pywr/recorders/_tables_recorders.pyx
+++ b/pywr/recorders/_tables_recorders.pyx
@@ -88,12 +88,16 @@ class TablesRecorder2(Recorder):
             The size in megabytes of buffer to keep in memory before appending to the PyTables data. This
             buffer is a total buffer for the whole recorder and is divided amongst the items to be recorded
             equally. Default is 100.
+        buffer_timesteps: int
+            The number of timesteps to buffer in memory before appending to the PyTables data. This is an alternative
+            to buffer_size, and if given will be used in preference to buffer_size.
         """
         self.filter_kwds = kwargs.pop("filter_kwds", {})
         self.mode = kwargs.pop("mode", "w")
         self.metadata = kwargs.pop("metadata", {})
         self.create_directories = kwargs.pop("create_directories", False)
         self.buffer_size = kwargs.pop("buffer_size", 100)
+        self.buffer_timesteps = kwargs.pop("buffer_timesteps", None)
         self._buffer_num_timesteps = 1
 
         title = kwargs.pop("title", None)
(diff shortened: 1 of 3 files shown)


pywr/recorders/_tables_recorders.pyx
```diff
@@ -0,0 +1,675 @@
...
+import warnings
+from pywr.parameters._parameters cimport IndexParameter, Parameter
+from pywr.recorders._recorders cimport Recorder
+from .._core cimport Timestep, Scenario, ScenarioIndex, AbstractNode, AbstractStorage, Storage
+
+class ParameterNameWarning(UserWarning):
+    pass
+
+
+class TablesRecorder2(Recorder):
```
Batch21, line 15: could this be a cdef class?

pywr/recorders/_tables_recorders.pyx
```diff
@@ -0,0 +1,675 @@
...
+            if buffer is None:
+                node.get_all_max_flow(out)
+                arrays.max_flow[idx, :] = np.reshape(out, scenario_shape)
+
+                node.get_all_min_flow(out)
+                arrays.min_flow[idx, :] = np.reshape(out, scenario_shape)
+
+                arrays.flow[idx, :] = np.reshape(node._flow, scenario_shape)
+            else:
+                buffer_idx = idx % self._buffer_num_timesteps
```
Batch21: should this be calculated outside of this loop? if there are no nodes then it will not be initialised when it is used in the param loops below

pywr/recorders/_tables_recorders.pyx
```diff
@@ -0,0 +1,675 @@
...
+
+        if self._routes_flow_array is not None:
+            routes_shape = [
+                len(self.model.solver.routes),
+            ] + scenario_shape
+            self._routes_flow_array[idx, ...] = np.reshape(
+                self.model.solver.routes_flows_array.T, routes_shape
+            )
+
+    def flush_buffer(self, int idx, int num_timesteps):
```
Batch21, line 593: could this and any of the other methods be cpdef methods?

pywr/recorders/_tables_recorders.pyx
```diff
@@ -0,0 +1,675 @@
...
+    def finish(self):
+
+        if self._buffer_num_timesteps > 1:
+            ts = self.model.timestepper.current
+            # The final time-step is one past the end of the data
+            idx = ts.index
+            residual_timesteps = idx % self._buffer_num_timesteps
+
+            # Flush the buffer if there are residual time-steps, or our buffer covered the entire simulation
+            if residual_timesteps > 0:
```
Batch21, line 625: if the final timestep fill the buffer it would not get flushed?

pywr/recorders/_tables_recorders.pyx
```diff
@@ -0,0 +1,684 @@
...
+
+                where = where.replace("//", "/")
+                items.append((where, node))
+
+        if self.parameters is not None:
+            items.extend(self.parameters)
+
+        # Work out the number of timesteps to buffer
+        if self.buffer_timesteps is not None:
+            self._buffer_num_timesteps = min(self._buffer_num_timesteps, len(self.model.timestepper))
```
Batch21: should this be `self._buffer_num_timesteps = min(self.buffer_timesteps, len(self.model.timestepper))` ?

Batch21 (changes requested):

pywr/recorders/_tables_recorders.pyx
jetuk: Fixed

Batch21 (approved):

PR #1185, opened 2026-05-30
fix: add get_all_min_volume and get_all_max_volume to methods to AggregatedStorage

Batch21:


pywr/_core.pyx
```diff
@@ -1426,6 +1426,34 @@ cdef class AggregatedStorage(AbstractStorage):
...
+        cdef int i
+        cdef Storage s
+        cdef double[:] mnvs
+        if out is None:
+            out = np.zeros(len(self.model.scenarios.combinations))
+        else:
+            out[:] = 0.0
+        for s in self._storage_nodes:
+            mnvs = s.get_all_min_volume()
+            for i in range(out.shape[0]):
```
jetuk: Should this use either the shape of `mnvs` or `len(self.model.scenarios.combinations)`? If the user passes the wrong `out` shape it could only populate the length of `out` and not error as intended?

pywr/_core.pyx
```diff
@@ -1426,6 +1426,34 @@ cdef class AggregatedStorage(AbstractStorage):
...
+        cdef int i
+        cdef Storage s
+        cdef double[:] mxvs
+        if out is None:
+            out = np.zeros(len(self.model.scenarios.combinations))
+        else:
+            out[:] = 0.0
+        for s in self._storage_nodes:
+            mxvs = s.get_all_max_volume()
+            for i in range(out.shape[0]):
```
jetuk: Same as above.

jetuk (changes requested):

pywr/_core.pyx
Batch21: Done

jetuk (approved):

PR #1186, opened 2026-06-02
fix: prevent tags from being passed to load_parameter_values function

Batch21:
This prevents the tags from being potentially passed to Pandas functions to read data from files resulting in an unexpected keyword error.


Batch21:
This is a quick fix. Maybe the better approach would be to ensure that all parameter pass the tags args correctly on to the base component class?

jetuk:
Can you provide a regression test for this?

jetuk (approved):

PR #1190, opened 2026-07-23
fix: store TablesRecorder2 ensemble names as per-scenario arrays

Batch21:


LaurenPetch1 (approved):
Looks good and I've been using for the tables recorder 1 implementation with no issues

